PR #58, opened 2021-03-23
Add configs for hades2

yichen88:

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -34,6 +34,7 @@
         <maven-shade-plugin.version>3.2.4</maven-shade-plugin.version>
         <slf4j.version>1.7.30</slf4j.version>
         <powysbl-core.version>4.0.1</powysbl-core.version>
+        <powsybl-hades2-integration.version>3.0.0</powsybl-hades2-integration.version>
         <powysbl-single-line-diagram.version>2.0.0</powysbl-single-line-diagram.version>
         <powsybl-open-load-flow.version>0.10.0-SNAPSHOT</powsybl-open-load-flow.version>
     </properties>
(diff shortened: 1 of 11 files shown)


java/pom.xml
```diff
@@ -205,6 +205,11 @@
             <artifactId>powsybl-open-loadflow</artifactId>
             <version>${powsybl-open-load-flow.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.rte-france.powsybl</groupId>
+            <artifactId>powsybl-hades2-integration</artifactId>
+            <version>3.0.0</version>
```
geofjamg: Add a property for the version like the other powsybl repo.

geofjamg (changes requested):
Gridpy should continue to work automatically without config.json  if we just want to use open loadflow and only requires a config.json if we want to switch to hades2. Could be done maybe by embedding a default config.yml in the compiled graalvm lib and stacking with the ~/.itools/config.yml one?

geofjamg:
@yichen88 the issue we have with Hades2Provider which read config.yml at construction will be fixed in next hades2 integration release

geofjamg:
@yichen88 in a previous PR I refactored all GraalVM metadata to be organized by dependency. Could you do the same for ADN/Hades2 ones?
Regarding Sensi integration, now there is 2 ways to run it
  - runV1 uses the PowSyBl Core sensi API
  - runV2 use a specific OLF sensi API => to have better performance
To still be able to run sensi from OLF or Hades2, we should just in the code runV1 if not OpenLoadFlow and runV2 if OpenLoadFlow.

@sylvlecl To make everybody happy, maybe we could (in this PR or not?) also integrate DynaFlow? 
So it would only consist on building powsybl-dynaflow (which contains a loadflow provider) from https://github.com/powsybl/powsybl-dynawo (release from Maven central exists). In the Maven central release there is the same issue that with hades2 provider 3.0.0, config.yml is read fro provider constructor but it is fixed in the master and so on in the next release (like hades2 provider).

PR #60, opened 2021-03-31
Add reduce

yichen88:

### DIFF
--- a/cpp/src/bindings.cpp
+++ b/cpp/src/bindings.cpp
@@ -66,6 +66,22 @@ PYBIND11_MODULE(_gridpy, m) {
 
     m.def("dump_network", &gridpy::dumpNetwork, "Dump network to a file in a given format");
 
+    py::class_<nominal_voltage_predicate>(m, "NominalVoltagePredicate")
+            .def(py::init([](double min, double max) {
+               auto p = new nominal_voltage_predicate();
+               p->min = min;
+               p->max = max;
+               return p;
+            }))
+            .def_property_readonly("min", [](const nominal_voltage_predicate& p) {
+                return p.min;
+            })
+            .def_property_readonly("max", [](const nominal_voltage_predicate& p) {
+                return p.max;
+            });
+
+    m.def("reduce_network", &gridpy::reduceNetwork, "Reduce network", py::arg("network"), py::arg("nominal_voltage_predicate"));
+
     py::enum_<gridpy::LoadFlowComponentStatus>(m, "LoadFlowComponentStatus")
             .value("CONVERGED", gridpy::LoadFlowComponentStatus::CONVERGED)
             .value("FAILED", gridpy::LoadFlowComponentStatus::FAILED)
--- a/cpp/src/bindings.cpp
+++ b/cpp/src/bindings.cpp
@@ -81,7 +81,7 @@ PYBIND11_MODULE(_gridpy, m) {
                 return p.max;
             });
 
-    m.def("reduce_network", &gridpy::reduceNetwork, "Reduce network", py::arg("network"), py::arg("nominal_voltage_predicate"), py::arg("ids"));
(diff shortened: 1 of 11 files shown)


geofjamg:
@yichen88 Great! Could you add a unit test in tests/test.py? For instance you can create eurostag tuto network, run a load flow and then reduce with only nominal voltage > 10 and < 30 and check there is only one bus remaining

cpp/src/bindings.cpp
```diff
@@ -66,6 +66,22 @@ PYBIND11_MODULE(_gridpy, m) {
 
     m.def("dump_network", &gridpy::dumpNetwork, "Dump network to a file in a given format");
 
+    py::class_<nominal_voltage_predicate>(m, "NominalVoltagePredicate")
+            .def(py::init([](double min, double max) {
```
geofjamg: Maybe we could define default values for min and max (0, MAX_DOUBLE) so we can just specify one bound?

geofjamg (approved):

tests/test.py
```diff
@@ -143,6 +143,26 @@ def test_exception(self):
...
+        gp.loadflow.run_ac(n)
+        self.assertEqual(4, len(n.buses))
+        n.reduce(ids=['P2'])
+        self.assertEqual(2, len(n.buses))
+
+    def test_reduce_by_subnetwork(self):
+        n = gp.network.create_eurostag_tutorial_example1_network()
+        gp.loadflow.run_ac(n)
+        self.assertEqual(4, len(n.buses))
+        n.reduce(vls=['VLGEN'], depths=[2])
```
geofjamg: I don't think it is useful to support a list of vl/depth. Only one is enough. Also maybe in this case we should have an object to group voltage level id and depth as it has to be defined at the same time.

geofjamg (changes requested):

tests/test.py
sylvlecl: I guess we can use a python tuple here for now : `vl_depth=('VL', 3)` ?

I think a list of VL + depth would be interesting but let's wait for a use case. Actually after thinking again I think the depth should be an option available for all predicates, for example a typical use case would be "my country + 4 substations around". In that case we could have a single option `depth=4`.

geofjamg: ok so with a default value of 1 so that we don't need to check if depth is defined in case if vl_id defined?

sylvlecl: I think for now it's better to stick with the predicates in powsybl --> only add `vl_depth` tuple argument, and for all other arguments there is no depth.

But yes if we add later a generic depth argument, it could default to 1 (or 0 ?)

geofjamg: Ok

gridpy/network.py
```diff
@@ -72,6 +74,12 @@ def disconnect(self, id: str):
     def dump(self, file: str, format: str = 'XIIDM'):
         _gridpy.dump_network(self.ptr, file, format)
 
+    def reduce(self, min: float = 0, max: float = sys.float_info.max, ids: List[str] = [],
```
sylvlecl: Maybe we can name `min` and `max` `vmin` and `vmax` instead, just to clarify a little their meaning.

(also a more general comment for the future: we should probably add some "docstrings" at some point if the library gets more used/maintained : it would help understand expected arguments)

geofjamg: Yes, docstrings is essential for wide usage of the library. It was on my TODO list but I did had time to start anything.
Also it would be great to have a readthedocs.io with a user doc and part made of generated doc from doncstrings. readthedocs.io seems to be a common way to document Python libraries (https://pandasguide.readthedocs.io/en/latest/ for instance)

cpp/src/bindings.cpp
```diff
@@ -66,6 +67,23 @@ PYBIND11_MODULE(_gridpy, m) {
 
     m.def("dump_network", &gridpy::dumpNetwork, "Dump network to a file in a given format");
 
+    py::class_<nominal_voltage_predicate>(m, "NominalVoltagePredicate")
```
geofjamg: Maybe if in the Python API, we only expose vmin vmax we can avoid this object and just pass 2 floats?

geofjamg (approved):

PR #83, opened 2021-04-20
Improve network load/save

geofjamg:
Signed-off-by: Geoffroy Jamgotchian <geoffroy.jamgotchian@rte-france.com>

**Other information**:


java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -150,23 +154,81 @@ public static ObjectHandle createEurostagTutorialExample1Network(IsolateThread t
...
+        return doCatch(exceptionHandlerPtr, () -> createCharPtrArray(new ArrayList<>(Importers.getFormats())));
+    }
+
+    @CEntryPoint(name = "getNetworkExportFormats")
+    public static ArrayPointer<CCharPointerPointer> getNetworkExportFormats(IsolateThread thread, ExceptionHandlerPointer exceptionHandlerPtr) {
+        return doCatch(exceptionHandlerPtr, () -> createCharPtrArray(new ArrayList<>(Exporters.getFormats())));
+    }
+
+    @CEntryPoint(name = "freeNetworkFormats")
+    public static void freeNetworkFormats(IsolateThread thread, ArrayPointer<CCharPointerPointer> formatsArrayPtr) {
```
sylvlecl: We should probably name it with a more generic "freeStringArray" or something like that

tests/test.py
```diff
@@ -31,6 +31,14 @@ def test_run_lf(self):
         results = pp.loadflow.run_dc(n, parameters)
         self.assertEqual(1, len(results))
 
+    def test_get_import_format(self):
+        formats = pp.network.get_import_formats()
+        self.assertEqual(['CGMES', 'MATPOWER', 'IEEE-CDF', 'PSS/E', 'UCTE', 'XIIDM'], formats)
+
+    def test_get_export_format(self):
```
sylvlecl, line 46: Can we add a test for parameters too ?

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
geofjamg: Done

tests/test.py
geofjamg: Added

sylvlecl (approved):

PR #86, opened 2021-04-23
Add sensi2

yichen88:

**Other information**:


pypowsybl/sensitivity_analysis.py
```diff
@@ -56,8 +56,8 @@ def set_factor_matrix(self, branches_ids: List[str], injections_or_transformers_
         self.branches_ids = branches_ids
         self.injections_or_transformers_ids = injections_or_transformers_ids
 
-    def run_dc(self, network: Network, parameters: Parameters = Parameters()) -> SensitivityAnalysisResult:
-        return SensitivityAnalysisResult(_pypowsybl.run_sensitivity_analysis(self.ptr, network.ptr, parameters),
+    def run_dc(self, network: Network, parameters: Parameters = Parameters(), provider = 'OpenLoadFlow') -> SensitivityAnalysisResult:
```
sylvlecl: Actually I don't think it will work !
The sensitivity analysis is named OpenSensitivityAnalysis

java/src/main/java/com/powsybl/python/SensitivityAnalysisContext.java
```diff
@@ -132,7 +132,15 @@ void setFactorMatrix(List<String> branchsIds, List<String> injectionsOrTransfosI
         return factors;
     }
 
-    SensitivityAnalysisResultContextV1 runV1(Network network, LoadFlowParameters loadFlowParameters, String provider) {
+    SensitivityAnalysisResultContext run(Network network, LoadFlowParameters loadFlowParameters, String provider) {
+        if ("OpenLoadFlow".equals(provider)) {
```
sylvlecl: We should be consistent with the actual provider name which is "OpenSensitivityAnalysis"

java/src/main/java/com/powsybl/python/SensitivityAnalysisResultContextV1.java
```diff
@@ -43,37 +45,62 @@
 
     @Override
     public PyPowsyblApiHeader.MatrixPointer createSensitivityMatrix(String contingencyId) {
```
sylvlecl, line 47: I encountered an "index out of bound" while testing with a large network and "open load flow" v1 : 
```
_pypowsybl.PyPowsyblError: Index 398412 out of bounds for length 370202
```
I suspect that in some cases, the collection of `SensitivityValue` is smaller than (functions*variables) : in that case, the computed index may be greater than `sensitivityValues.length`.

sylvlecl: I mean: it's not an issue for open load flow, since in reality we will use v2, but I think we need to handle the cases where some sensitivity values are missing.
Maybe we should fill the result with `NaN` in those cases ? (is it technically possible to pass `NaN` from java to python? )

PR #91, opened 2021-04-30
Network update more

### LINKED ISSUE #79: Add possibility to modify the powergrid python side: loads, generators, storage units and shunts
## **Do you want to request a *feature* or report a *bug*?**

This is a feature request.

## **What is the current behavior?**

Today, i did not see any way to change the grid from python side, beside the opening / closing of switches detailed in the example of the readme.

## **What is the motivation / use case for changing the behavior?**
This would allow for a great interoperability with grid2op (https://github.com/rte-france/Grid2Op) where the loads / generators are changed a lot: we represent the "time" and at each "time step" the productions / loads / shunts / storage units are modified.

## **What is the expected behavior?**

For example, i would like to be able to:

- change a load (active or reactive consumption )
- change a generator (active production or voltage setpoint)
- change a storage unit (active production / consumption)
- change a shunt (active / reactive)

This would give, for unitary changes, something like
```python
import pypowsybl as pp
net = pp.load(...)
net.change_load(load_name_or_id, new_p, new_q)
net.change_gen(gen_name_or_id, new_p, new_v)
net.change_storage(storage_name_or_id, new_p)
net.change_shunt(storage_name_or_id, new_p)
```

And for performance reasons, it would also be interesting to do it in a "vectorized" fashion:
```python
import pypowsybl as pp
net = pp.load(...)
net.change_loads(vector_of_load_name_or_id, vector_of_new_p, vector_of_new_q)
# same for `change_gens` and `change_storages`
```

## **Please tell us about your environment:**
  - PowSyBl Version: `all`
  - OS Version: `all`
Ideally, all version would be concerned.

yichen88:

**Other information**:


pypowsybl/network.py
```diff
@@ -161,6 +164,46 @@ def create_substations_data_frame(self) -> pd.DataFrame:
     def create_hvdc_lines_data_frame(self) -> pd.DataFrame:
         return self.create_elements_data_frame(_pypowsybl.ElementType.HVDC_LINE)
 
+    def create_switches_data_frame(self) -> pd.DataFrame:
+        return self.create_elements_data_frame(_pypowsybl.ElementType.SWITCH)
+
+    def update_elements_with_data_frame(self, element_type: _pypowsybl.ElementType, df: pd.DataFrame):
+        for seriesName in df.columns.values:
+            series = df[seriesName]
+            if seriesName.endswith('p0') or seriesName.endswith('q0') or seriesName.endswith('setpoint'):
```
geofjamg: We have to stay generic there is no way to guess real type for use input. We have to get column types from Java and then create the dataframe. So @yichen88 could you add to the other PR a 
_powsybl.get_series_types(element_type) to help create the dataframe with the correct type?

geofjamg (changes requested):

geofjamg (approved):

PR #100, opened 2021-05-16
Add voltage level id to all connectable data frames

geofjamg:
Signed-off-by: Geoffroy Jamgotchian <geoffroy.jamgotchian@rte-france.com>

**Other information**:

### DIFF
--- a/java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
+++ b/java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
@@ -719,6 +719,8 @@ public final class PyPowsyblApiLib {
                             .addDoubleSeries("q1", l -> l.getTerminal1().getQ())
                             .addDoubleSeries("p2", l -> l.getTerminal2().getP())
                             .addDoubleSeries("q2", l -> l.getTerminal2().getQ())
+                            .addStringSeries("voltage_level1_id", l -> l.getTerminal1().getVoltageLevel().getId())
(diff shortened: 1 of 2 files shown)


java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -916,8 +933,8 @@ private static String getBusId(Terminal t) {
                             .addDoubleSeries("max_p", HvdcLine::getMaxP)
                             .addDoubleSeries("nominal_v", HvdcLine::getNominalV)
                             .addDoubleSeries("r", HvdcLine::getR)
-                            .addStringSeries("converter_station1", l -> l.getConverterStation1().getId())
-                            .addStringSeries("converter_station2", l -> l.getConverterStation2().getId()))
+                            .addStringSeries("converter_station1_id", l -> l.getConverterStation1().getId())
+                            .addStringSeries("converter_station2_id", l -> l.getConverterStation2().getId()))
```
sylvlecl, line 937: I guess we should add voltage levels here too ?

geofjamg: Not needed because we can join with converter stations which have a voltage level id

sylvlecl: ok

sylvlecl (approved):

Issue #104, opened 2021-05-17
[Tests] Add unit tests on the java layer


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Technical improvement

* **What is the current behavior?**

There is almost no unit tests on the java layer, which does not allow to catch problems early in the dev cycle.

* **What is the expected behavior?**

We should have unit tests for all logic happening on the java side. In particular, we should have unit tests for the series based operations (read data as series, update network from series).

In order to achieve this, we will need to separate more between the logic (building the series in this case), and the pure conversion to C objects. 
Those 2 things should happen in 2 different classes, so that we can unit test the logic.
In order to not sacrifice to performances, we should not make additional copies for that : the "logic" component could use consumers as arguments to achieve this, instead of returning arrays which would then need to be copied.

* **What is the motivation / use case for changing the behavior?**

Speed up validation of the java layer, during dev.

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

sylvlecl:
Closing the ticket since it has mainly been tackled in #106 and #126

PR #106, opened 2021-05-18
Add series data helper class

yichen88:

#104 

**Other information**:

### DIFF
--- a/java/src/main/java/com/powsybl/python/SeriesArrayHelper.java
+++ b/java/src/main/java/com/powsybl/python/SeriesArrayHelper.java
@@ -25,122 +25,22 @@ final class SeriesArrayHelper {
     static SeriesPointerArrayBuilder prepareData(Network network, PyPowsyblApiHeader.ElementType elementType) {
         switch (elementType) {
             case BUS:
-                List<Bus> buses = network.getBusView().getBusStream().collect(Collectors.toList());
-                return addProperties(new SeriesPointerArrayBuilder<>(buses)
-                        .addStringSeries("id", true, Bus::getId)
-                        .addDoubleSeries("v_mag", Bus::getV)
-                        .addDoubleSeries("v_angle", Bus::getAngle));
+                return prepareBusData(network);
 
             case LINE:
-                List<Line> lines = network.getLineStream().collect(Collectors.toList());
-                return addProperties(new SeriesPointerArrayBuilder<>(lines)
-                        .addStringSeries("id", true, Line::getId)
-                        .addDoubleSeries("r", Line::getR)
-                        .addDoubleSeries("x", Line::getX)
-                        .addDoubleSeries("g1", Line::getG1)
-                        .addDoubleSeries("b1", Line::getB1)
-                        .addDoubleSeries("g2", Line::getG2)
-                        .addDoubleSeries("b2", Line::getB2)
-                        .addDoubleSeries("p1", l -> l.getTerminal1().getP())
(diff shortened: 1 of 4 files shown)


java/src/main/java/com/powsybl/python/SeriesArrayHelper.java
```diff
@@ -0,0 +1,568 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * @author Yichen TANG <yichen.tang at rte-france.com>
+ */
+final class SeriesArrayHelper {
+
+    static SeriesPointerArrayBuilder prepareData(Network network, PyPowsyblApiHeader.ElementType elementType) {
+        switch (elementType) {
+            case BUS:
```
sylvlecl, line 27: I propose to have 1 function per equipment type to continue improve readability.

java/src/main/java/com/powsybl/python/SeriesArrayHelper.java
```diff
@@ -0,0 +1,568 @@
...
+        }
+    }
+
+    static void updateNetworkElementsWithDoubleSeries(Network network, PyPowsyblApiHeader.ElementType elementType, int elementCount,
+                                                      String seriesName, IntFunction<String> idGetter, IntToDoubleFunction valueGetter) {
+        for (int i = 0; i < elementCount; i++) {
+            String id = idGetter.apply(i);
+            double value = valueGetter.applyAsDouble(i);
+            switch (elementType) {
+                case GENERATOR:
```
sylvlecl, line 95: Same for the updating : 1 function by type would improve readability.

java/src/test/java/com/powsybl/python/SeriesArrayHelperTest.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+class SeriesArrayHelperTest {
+
+    @Test
+    void testGetter() {
+        Network network = EurostagTutorialExample1Factory.create();
+
+        final SeriesPointerArrayBuilder builder = SeriesArrayHelper.prepareData(network, PyPowsyblApiHeader.ElementType.BUS);
+
+        assertEquals("VLGEN_0", ((SeriesPointerArrayBuilder.StringSeries) builder.seriesList.get(0)).stringGetter.apply(builder.getElements().get(0)));
```
sylvlecl: For tests it would be better to not need to apply all that logic, but to get arrays, like on python side, in which we can check the values.
Actually, getting tables from the network could be an interesting feature for java too, independently of C interface.

Maybe in a simple approach we can just add a method to the builder to get data as java arrays (+ names) ?

java/src/main/java/com/powsybl/python/SeriesPointerArrayBuilder.java
```diff
@@ -250,4 +289,15 @@ public PointerBase createDataPtr(List<T> elements) {
 
         return PyPowsyblApiHeader.allocArrayPointer(seriesPtr, seriesList.size());
     }
+
+    // used for unittest
+    List<List> buildJavaSeries() {
```
sylvlecl, line 294: Looks good : actually maybe we can just return the `List<Series>` ? this way we can check also the series names.

sylvlecl:
@yichen88 I think we could re-use your "series builder" interface from powsybl-core PR here ! 
Probably easier to merge this one first though, I will take a last look at it.

sylvlecl (approved):
I propose to merge it as a good step to improve readability and testability : 
I think we can push further, later, by re-using "series provider" interface from powsybl-core, and probably have smaller classes.

PR #125, opened 2021-06-09
Read/write for tap changers

sylvlecl:

Feature

We can list and update ratio tap changers, as well as phase tap changers.


sylvlecl:
@geofjamg I removed the redundant "tap_position" attributes from TWT table

geofjamg (commented):
Maybe we should do it also for 3 windings transformers?

sylvlecl:
> Maybe we should do it also for 3 windings transformers?

Indeed, I was wondering if we should do it in the same tables or not. I think we will not be able to, because for the 3 windings we will need a mixed index.

geofjamg:
> mixed

I would got for same table but the second index (with value 1?) won't be used when querying for a 2 windings transformer.

sylvlecl:
Closing since the 2 windings transformers change has been included in #126 .
Work on 3 windings transformers remains in the todo list though.

PR #127, opened 2021-06-11
[WIP] Create network elements by dataframe

yichen88:

**Other information**:


tests/test.py
```diff
@@ -32,6 +32,13 @@ def test_run_lf(self):
         results = pp.loadflow.run_dc(n, parameters)
         self.assertEqual(1, len(results))
 
+    def test_new_load(self):
+        n = pp.network.create_eurostag_tutorial_example1_network()
+        n.new_load('VLLOAD', 'LOAD2', double_map={"p0" : 1.1, "q0" : 2.1}, str_map={'connectable_bus':'NLOAD', 'bus': 'NLOAD'})
```
geofjamg: Why not doing it using a dataframe update? that would be more consistent with what has already been done.

geofjamg (changes requested):

cpp/src/pypowsybl.cpp
```diff
@@ -418,6 +418,19 @@ void updateNetworkElementsWithStringSeries(void* network, element_type elementTy
                 idPtr.get(), valuePtr.get(), elementCount);
 }
 
+void newElement(void* network, const std::string& id, element_type elementType,
```
geofjamg: @yichen88 have you tried to create Series from Python side instead of having all theses maps?

yichen88: When we add elements, we need a `ROW` of data to build an equipment together, we can not do with `COLUMN'` in Series.

sylvlecl:
**About scope of that evolution:** 
we can start with a first evolution with only generators and loads (which already have use cases).

**Then we will have a design question for elements with "nested" elements**, such as transformers and tap changer steps, or shunts and shunts models. They cannot be represented as a single dataframe. A possibility would be to be able to pass multiple dataframes : for example 1 for transformers data and 1 for the corresponding  steps data.

sylvlecl:
We will need to refactor a little how data is provided from python to java, based on the work on #234 .

As for the element updates, we should now transfer the whole dataframe at once.

sylvlecl:
Will be merged with / refactored in #259

PR #138, opened 2021-06-22
Interface for on demand results

EtienneLt:

**Other information**:

### DIFF
--- a/java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
+++ b/java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
@@ -10,6 +10,8 @@ import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
 import com.powsybl.commons.PowsyblException;
 import com.powsybl.computation.local.LocalComputationManager;
+import com.powsybl.contingency.ContingencyContext;
+import com.powsybl.contingency.ContingencyContextType;
 import com.powsybl.ieeecdf.converter.IeeeCdfNetworkFactory;
 import com.powsybl.iidm.export.Exporters;
 import com.powsybl.iidm.import_.ImportConfig;
@@ -30,6 +32,7 @@ import com.powsybl.openloadflow.sensi.WeightedSensitivityVariable;
 import com.powsybl.security.LimitViolation;
 import com.powsybl.security.LimitViolationsResult;
 import com.powsybl.security.SecurityAnalysisResult;
+import com.powsybl.security.monitor.StateMonitor;
 import com.powsybl.security.results.PostContingencyResult;
 import com.powsybl.tools.Version;
 import org.graalvm.nativeimage.IsolateThread;
@@ -53,6 +56,7 @@ import java.util.function.BooleanSupplier;
 import java.util.function.IntSupplier;
 import java.util.stream.Collectors;
 
+import static com.powsybl.python.CTypeUtil.toStringList;
 import static com.powsybl.python.PyPowsyblApiHeader.*;
 
 /**
@@ -222,8 +226,8 @@ public final class PyPowsyblApiLib {
 
     private static Properties createParameters(CCharPointerPointer parameterNamesPtrPtr, int parameterNamesCount,
                                                CCharPointerPointer parameterValuesPtrPtr, int parameterValuesCount) {
-        List<String> parameterNames = CTypeUtil.toStringList(parameterNamesPtrPtr, parameterNamesCount);
-        List<String> parameterValues = CTypeUtil.toStringList(parameterValuesPtrPtr, parameterValuesCount);
+        List<String> parameterNames = toStringList(parameterNamesPtrPtr, parameterNamesCount);
+        List<String> parameterValues = toStringList(parameterValuesPtrPtr, parameterValuesCount);
         Properties parameters = new Properties();
         for (int i = 0; i < parameterNames.size(); i++) {
             parameters.setProperty(parameterNames.get(i), parameterValues.get(i));
@@ -281,12 +285,12 @@ public final class PyPowsyblApiLib {
                 predicates.add(new NominalVoltageNetworkPredicate(vMin, vMax));
             }
             if (idsCount != 0) {
-                List<String> ids = CTypeUtil.toStringList(idsPtrPtr, idsCount);
+                List<String> ids = toStringList(idsPtrPtr, idsCount);
                 predicates.add(new IdentifierNetworkPredicate(ids));
             }
             if (depthsCount != 0) {
                 final List<Integer> depths = CTypeUtil.toIntegerList(depthsPtr, depthsCount);
-                final List<String> voltageLeveles = CTypeUtil.toStringList(vlsPtrPtr, vlsCount);
+                final List<String> voltageLeveles = toStringList(vlsPtrPtr, vlsCount);
                 for (int i = 0; i < depths.size(); i++) {
                     predicates.add(new SubNetworkPredicate(network.getVoltageLevel(voltageLeveles.get(i)), depths.get(i)));
                 }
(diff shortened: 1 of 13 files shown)


tests/test.py
```diff
@@ -450,7 +463,16 @@ def test_variant(self):
         self.assertEqual(1, len(n.get_variant_ids()))
 
 
-
+    def test_monitor_state(self):
+        n = pp.network.create_eurostag_tutorial_example1_network()
+        sa = pp.security.create_analysis()
+        sa.add_single_element_contingency('NHV1_NHV2_1', 'NHV1_NHV2_1')
```
geofjamg, line 468: I have same concern with sensitivity factors per contingency that is not yet interfaced in Python.
Instead of having the enum on Python side I would have only rely on a contingencies array with the following convention:
  - None (default value) -> all
  - empty -> none
  - non empty array -> specific for theses contingencies

@EtienneLt @sylvlecl what do you think?

sylvlecl: Indeed it looks more convenient and efficient to be able to specify a list of contingencies instead of just one contingency.

But for the convention `None` = all and `[]` = N, I don't think it will be clear for the user !
To improve this we propose to:
* keep the enum-based methods
* add convenience methods for the user which will call the enum-based methods, see for example below. I would propose to rename to "add monitored elements" which sounds more intuitive than "add state monitors", what do you think about it ?
```python
sa.add_precontingency_monitored_elements(...)
sa.add_postcontingency_monitored_elements(contingency_ids=[...], ...)
sa.add_monitored_elements(...)
```

geofjamg: Looks nice, ok!

pypowsybl/security.py
```diff
@@ -1,41 +1,44 @@
...
 import _pypowsybl
 from _pypowsybl import ContingencyResult
 from _pypowsybl import LimitViolation
 from _pypowsybl import LimitType
 from _pypowsybl import Side
 from pypowsybl.network import Network
 from pypowsybl.loadflow import Parameters
-from pypowsybl.util import ContingencyContainer
+from pypowsybl.util import ContingencyContainer, create_data_frame_from_series_array
 from prettytable import PrettyTable
```
sylvlecl, line 20: We need to add
```python
from _pypowsybl import ContingencyContextType
```
so that the enum is usable without import "_pypowsybl" (which should remain hidden to the user).

pypowsybl/security.py
```diff
@@ -1,41 +1,44 @@
...
+
 import _pypowsybl
 from _pypowsybl import ContingencyResult
 from _pypowsybl import LimitViolation
 from _pypowsybl import LimitType
 from _pypowsybl import Side
 from pypowsybl.network import Network
 from pypowsybl.loadflow import Parameters
-from pypowsybl.util import ContingencyContainer
+from pypowsybl.util import ContingencyContainer, create_data_frame_from_series_array
```
sylvlecl: I think we should not import `create_data_frame_from_series_array` to that module namespace, otherwise the user will see it when inspecting the `powsybl.security` module methods, whereas it should be here only for internal use.

pypowsybl/security.py
```diff
@@ -77,14 +80,43 @@ def get_table(self):
...
 class SecurityAnalysis(ContingencyContainer):
     def __init__(self, ptr):
         ContingencyContainer.__init__(self, ptr)
 
-    def run_ac(self, network: Network, parameters: Parameters = Parameters(), provider='OpenSecurityAnalysis') -> SecurityAnalysisResult:
+    def run_ac(self, network: Network, parameters: Parameters = Parameters(),
+               provider='OpenSecurityAnalysis') -> SecurityAnalysisResult:
         return SecurityAnalysisResult(_pypowsybl.run_security_analysis(self.ptr, network.ptr, parameters, provider))
 
+    def add_state_monitors(self, contingency_context_type: _pypowsybl.ContingencyContextType,
```
sylvlecl: To be renamed to `add_monitored_elements`, and "convenient" methods to be added for pre/post contingency(see previous comment in the pull request).

pypowsybl/security.py
```diff
@@ -77,14 +80,43 @@ def get_table(self):
...
+                           branch_ids: List[str] = None,
+                           voltage_level_ids: List[str] = None,
+                           three_windings_transformer_ids: List[str] = None,
+                           contingencies=None):
+
+        if three_windings_transformer_ids is None:
```
sylvlecl, line 138: we can directly use empty lists as default arguments, no need for `None` + if statement

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -886,4 +898,114 @@ public static void removeVariant(IsolateThread thread, ObjectHandle networkHandl
...
+                .map(threeWindingsTransformerResult -> new ThreeWindingsTransformerResultContext(threeWindingsTransformerResult, null))
+                .collect(Collectors.toList());
+            result.getPostContingencyResults().forEach(postContingencyResult -> {
+                postContingencyResult.getThreeWindingsTransformerResult()
+                    .forEach(threeWindingsTransformerResult ->
+                        threeWindingsTransformerResults.add(new ThreeWindingsTransformerResultContext(threeWindingsTransformerResult,
+                            postContingencyResult.getContingency().getId())));
+            });
+            return new SeriesPointerArrayBuilder<>(threeWindingsTransformerResults)
+                .addStringSeries("contingency", true, ThreeWindingsTransformerResultContext::getContingency)
```
sylvlecl: to be consistent with other parts of the api, I propose to add `_id` suffix

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -886,4 +898,114 @@ public static void removeVariant(IsolateThread thread, ObjectHandle networkHandl
...
+                .addStringSeries("branch_id", true, ThreeWindingsTransformerResultContext::getThreeWindingsTransformerId)
```
sylvlecl: 3WT are not in the "branch" catefory, so maybe we can use `transformer_id` here

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -886,4 +898,114 @@ public static void removeVariant(IsolateThread thread, ObjectHandle networkHandl
...
+            List<BusResultContext> busResults = result.getPreContingencyResult()
+                .getPreContingencyBusResults().stream()
+                .map(busResult -> new BusResultContext(busResult, null))
+                .collect(Collectors.toList());
+            result.getPostContingencyResults().forEach(postContingencyResult -> {
+                postContingencyResult.getBusResults()
+                    .forEach(busResult -> busResults.add(new BusResultContext(busResult, postContingencyResult.getContingency().getId())));
+            });
+            return new SeriesPointerArrayBuilder<>(busResults)
+                .addStringSeries("voltage_level_id", true, BusResultContext::getVoltageLevelId)
```
sylvlecl, line 975: we need to add `contingency_id` in first position here too

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -886,4 +898,114 @@ public static void removeVariant(IsolateThread thread, ObjectHandle networkHandl
...
+            List<BranchResultContext> branchResults = result.getPreContingencyResult()
+                    .getPreContingencyBranchResults().stream()
+                    .map(branchResult -> new BranchResultContext(branchResult, null))
+                    .collect(Collectors.toList());
+            result.getPostContingencyResults().forEach(postContingencyResult -> {
+                postContingencyResult.getBranchResults()
+                    .forEach(branchResult -> branchResults.add(new BranchResultContext(branchResult, postContingencyResult.getContingency().getId())));
+            });
+            return new SeriesPointerArrayBuilder<>(branchResults)
+                .addStringSeries("contingency", true, BranchResultContext::getContingency)
```
sylvlecl: let's add `_id` here too

sylvlecl (changes requested):

sylvlecl:
Apart from the above comments about the code, the feature seems to work fine :)

sylvlecl (approved):

PR #141, opened 2021-06-22
Refactor init parameter

yichen88:

**Other information**:


cpp/src/bindings.cpp
```diff
@@ -147,36 +146,28 @@ PYBIND11_MODULE(_pypowsybl, m) {
...
-                parameters->write_slack_bus = writeSlackBus;
-                parameters->distributed_slack = distributedSlack;
-                parameters->balance_type = balanceType;
-                parameters->dc_use_transformer_ratio = dcUseTransformerRatio;
-                parameters->countries_to_balance = pypowsybl::copyVectorStringToCharPtrPtr(countriesToBalance);
-                parameters->countries_to_balance_count = countriesToBalance.size();
-                parameters->connected_component_mode = connectedComponentMode;
+            .def(py::init([]() {
+                 auto parameters = new load_flow_parameters();
+                 load_flow_parameters* config_params = pypowsybl::readParameters();
```
sylvlecl: There will be a small memory leak here!
you'll need to delete this at the end (or you can embed it in a smart pointer so that it does the job for you :) ).

Other option : maybe the `readParameters` could be an `updateParameters` instead which would take an existing `parameters` as input and update it on the java side ?

yichen88:
After this implemation, it raises another issue for unit-test.
If a new user has already his own ~/.itools/config.yml, the `unit-test` could be failed.
Solution 1: in unit-test, we provided a clean loadflow parameter.
Solution 2: an options on constructor of _pypowsylb LoadFlowParamater to whether read config or not. If not, we provided all default value(then we repeated again)

yichen88:
After this implemation, it raises another issue for unit-test.
If a new user has already his own ~/.itools/config.yml, the `unit-test` could be failed.
Solution 1: in unit-test, we provided a clean loadflow parameter.
Solution 2: an options on constructor of _pypowsylb LoadFlowParamater to whether read config or not. If not, we provided all default value(then we repeated again)

geofjamg:
> After this implemation, it raises another issue for unit-test.
> If a new user has already his own ~/.itools/config.yml, the `unit-test` could be failed.
> Solution 1: in unit-test, we provided a clean loadflow parameter.
> Solution 2: an options on constructor of _pypowsylb LoadFlowParamater to whether read config or not. If not, we provided all default value(then we repeated again)

To solve that we have a powsybl-config-test as an alternative to powsybl-config-classic just for unit testing that ensure independency to system config

geofjamg (changes requested):
I don't like this new design we still need to be able to write 
```python 
pp.loadflow.Parameters(distributed_slack=True)
```

I think we should just use a global variable to deactivate the read of config.yml (and switch to default contructor in Java) as it is only for unit testing.
So in unit tests we will only need to start with
```python
pp.set_config_read(False)
```

sylvlecl:
Just some notes about remaining work, as discussed this morning:

- remove `readFromConfig` parameter from constructor and use the global variable instead to decide if we should load from config or not
- as we mix objects from java and c++, we need to be careful about how we free memory of parameters : in order to make it simple, the update of parameters must take place on java side, and the cleanup implemented on java side only
- we need to have optional arguments in the constructor of parameters: in order to achieve this, we may use `py::object` as arguments, or a `py::kwargs` (maybe less convenient for the user though ?)

cpp/src/bindings.cpp
```diff
@@ -159,36 +162,44 @@ PYBIND11_MODULE(_pypowsybl, m) {
...
+                 if (countriesToBalance.is_none()) {
+                     parameters->countries_to_balance = config_params->countries_to_balance;
+                     parameters->countries_to_balance_count = config_params->countries_to_balance_count;
+                 } else {
+                     std::vector<std::string> countries = countriesToBalance.cast<std::vector<std::string>>();
+                     parameters->countries_to_balance = pypowsybl::copyVectorStringToCharPtrPtr(countries);
+                     parameters->countries_to_balance_count = countries.size();
+                 }
+                 parameters->connected_component_mode = connectedComponentMode.is_none() ? config_params->connected_component_mode : connectedComponentMode.cast<pypowsybl::ConnectedComponentMode>();
+                 delete config_params;
```
sylvlecl: So here, as `config_params` has been allocated by java code, it must be deallocated in java code too.
It will cause issues if the content of pointers have not been **copied** to `parameters` : it's only an issue for the `char** countries_to_balance` field.

cpp/src/bindings.cpp
```diff
@@ -159,36 +162,44 @@ PYBIND11_MODULE(_pypowsybl, m) {
...
+                 parameters->twt_split_shunt_admittance = twtSplitShuntAdmittance.is_none() ? config_params->twt_split_shunt_admittance : twtSplitShuntAdmittance.cast<bool>();
+                 parameters->simul_shunt = simulShunt.is_none() ? config_params->simul_shunt : simulShunt.cast<bool>();
+                 parameters->read_slack_bus = readSlackBus.is_none() ? config_params->read_slack_bus : readSlackBus.cast<bool>();
+                 parameters->write_slack_bus = writeSlackBus.is_none() ? config_params->write_slack_bus : writeSlackBus.cast<bool>();
+                 parameters->distributed_slack = distributedSlack.is_none() ? config_params->distributed_slack : distributedSlack.cast<bool>();
+                 parameters->balance_type = balanceType.is_none() ? config_params->balance_type : balanceType.cast<pypowsybl::BalanceType>();
+                 parameters->dc_use_transformer_ratio = dcUseTransformerRatio.is_none() ? config_params->dc_use_transformer_ratio : dcUseTransformerRatio.cast<bool>();
+                 if (countriesToBalance.is_none()) {
+                     parameters->countries_to_balance = config_params->countries_to_balance;
```
sylvlecl: See previous comment : here we must copy the content, otherwise have an ownership issue and therefore a memory issue.

sylvlecl (approved):

Issue #143, opened 2021-06-23
[Security analysis] Return violations as a dataframe


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**

Violations are reported in a tree of python objects, themselves mapped on C structures.

* **What is the expected behavior?**

It will be simpler and more consistent with other parts of the python API to return them as a pandas dataframe.

The dataframe index needs to contain the contingency ID (empty string for pre-contingency violations).
To be confirmed: should we make it a multi-indexed dataframe, with for example information about the subject of the violation for other columns ? Not sure it has an added value for now.

* **What is the motivation / use case for changing the behavior?**

* From the implementation point of view, simplify the mapping of limit violation results from java to python
* From the user point of view, having a more consistent API, largely based on pandas dataframes

geofjamg:
Maybe the violation type as a second index? We often need to get for instance only current violation.

EtienneLt:
> Maybe the violation type as a second index? We often need to get for instance only current violation.

will a triple index with violation id contingency id and violation type do the job ?

sylvlecl:
Why not but we must keep the contingency ID first to have a tree-like structure.

The "issue" is that in any case, even a tuple (`contingency_id`, `violation_type`, `subject_id`) will not be unique : for example you can have violations on sides 1 and 2 of a line, or on multiple buses in a voltage level.

I am not enough used to using pandas to estimate if it has an added value to add `subject_id` to the index columns, instead of having it in a "normal" column. 
Actually you can also filter on normal columns, so the added value is not clear to me. Maybe when selecting on an index columns we have better performance?

If anyone knows ... :)

sylvlecl:
Implemented in #145

PR #145, opened 2021-06-24
Return violations as dataframe

EtienneLt:

#143

### DIFF
--- a/java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
+++ b/java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
@@ -181,7 +181,7 @@ public final class PyPowsyblApiLib {
     @CEntryPoint(name = "createEurostagTutorialExample1Network")
     public static ObjectHandle createEurostagTutorialExample1Network(IsolateThread thread, ExceptionHandlerPointer exceptionHandlerPtr) {
         return doCatch(exceptionHandlerPtr, () -> {
-            Network network = EurostagTutorialExample1Factory.create();
+            Network network = EurostagTutorialExample1Factory.createWithFixedCurrentLimits();
             return ObjectHandles.getGlobal().create(network);
         });
     }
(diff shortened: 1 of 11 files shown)


pypowsybl/security.py
```diff
@@ -40,47 +42,18 @@
...
-                               "{:.1f}".format(limit_violation.limit),
-                               limit_violation.limit_name,
-                               limit_violation.acceptable_duration,
-                               limit_violation.limit_reduction,
-                               "{:.1f}".format(limit_violation.value),
-                               limit_violation.side.name])
-        return table
+        return self.results.query('contingency_id != \'\'')
+
+    def get_limit_violations(self):
```
sylvlecl: The "pythonic way" for getters will be to use a read only property as above:
```python
@property
def limit_violations(self):
```

pypowsybl/security.py
```diff
@@ -40,47 +42,18 @@
...
     def __init__(self, result):
         self.ptr = result
-        results = _pypowsybl.get_security_analysis_result(result)
-        self._post_contingency_results = {}
-        for result in results:
-            if result.contingency_id:
-                self._post_contingency_results[result.contingency_id] = result
-            else:
-                self._pre_contingency_result = result
+        self.results = _create_data_frame_from_series_array(_pypowsybl.get_security_analysis_result(result))
```
sylvlecl: I think for now we must **also** keep the "old" result, because it carries the information about the validity of the computation for each contingency: if we keep only the violations dataframe we lose it!

pypowsybl/security.py
```diff
@@ -40,47 +42,18 @@
...
 
     @property
     def pre_contingency_result(self):
-        return self._pre_contingency_result
+        return self.results.query('contingency_id == \'\'')
```
sylvlecl: in those properties we should still return the old results

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -499,10 +507,28 @@ public static ObjectHandle runSecurityAnalysis(IsolateThread thread, ObjectHandl
...
+            List<LimitViolationContext> limitViolations = result.getPreContingencyResult().getLimitViolationsResult().getLimitViolations()
+                .stream().map(limitViolation -> new LimitViolationContext("", limitViolation)).collect(Collectors.toList());
+            result.getPostContingencyResults()
+                .forEach(postContingencyResult -> limitViolations.addAll(postContingencyResult.getLimitViolationsResult()
+                    .getLimitViolations().stream()
+                    .map(limitViolation -> new LimitViolationContext(postContingencyResult.getContingency().getId(), limitViolation))
+                    .collect(Collectors.toList())));
+            return new SeriesPointerArrayBuilder<>(limitViolations)
+                    .addStringSeries("contingency_id", true, LimitViolationContext::getContingencyId)
+                    .addStringSeries("violation_id", true, LimitViolation::getSubjectId)
```
sylvlecl: we should name it "subject_id" like on java side, because it's not an identifier for the violation: we can have multiple violations for the same subject

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -499,10 +507,28 @@ public static ObjectHandle runSecurityAnalysis(IsolateThread thread, ObjectHandl
...
+                    .addStringSeries("violation_name", p -> Objects.toString(p.getSubjectName(), ""))
```
sylvlecl: same let's keep subject_name

sylvlecl (changes requested):

sylvlecl (approved):

PR #162, opened 2021-06-29
Dump network to string

yichen88:

**Other information**:

### DIFF
--- a/cpp/src/bindings.cpp
+++ b/cpp/src/bindings.cpp
@@ -97,10 +97,10 @@ PYBIND11_MODULE(_pypowsybl, m) {
           py::arg("file_name"), py::arg("file_content"),py::arg("parameters"));
 
     m.def("dump_network", &pypowsybl::dumpNetwork, "Dump network to a file in a given format", py::arg("file"),
-          py::arg("format"), py::arg("parameters"), py::arg("parameters"));
+          py::arg("format"), py::arg("parameters"));
 
     m.def("dump_network_to_string", &pypowsybl::dumpNetworkToString, "Dump network in a given format",
-          py::arg("format"), py::arg("parameters"), py::arg("parameters"));
+          py::arg("format"), py::arg("parameters"));
 
     m.def("reduce_network", &pypowsybl::reduceNetwork, "Reduce network", py::arg("network"), py::arg("v_min"), py::arg("v_max"),
           py::arg("ids"), py::arg("vls"), py::arg("depths"), py::arg("with_dangling_lines"));
--- a/cpp/src/bindings.cpp
+++ b/cpp/src/bindings.cpp
@@ -96,11 +96,11 @@ PYBIND11_MODULE(_pypowsybl, m) {
     m.def("load_network_from_string", &pypowsybl::loadNetworkFromString, "Load a network from a string",
           py::arg("file_name"), py::arg("file_content"),py::arg("parameters"));
 
-    m.def("dump_network", &pypowsybl::dumpNetwork, "Dump network to a file in a given format", py::arg("file"),
-          py::arg("format"), py::arg("parameters"));
+    m.def("dump_network", &pypowsybl::dumpNetwork, "Dump network to a file in a given format", py::arg("network"),
+          py::arg("file"),py::arg("format"), py::arg("parameters"));
 
     m.def("dump_network_to_string", &pypowsybl::dumpNetworkToString, "Dump network in a given format",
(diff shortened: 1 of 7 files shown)


java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -291,6 +291,35 @@ public static void dumpNetwork(IsolateThread thread, ObjectHandle networkHandle,
...
+    public static CCharPointer dumpNetworkToString(IsolateThread thread, ObjectHandle networkHandle, CCharPointer format,
+                                                   CCharPointerPointer parameterNamesPtrPtr, int parameterNamesCount,
+                                                   CCharPointerPointer parameterValuesPtrPtr, int parameterValuesCount,
+                                                   ExceptionHandlerPointer exceptionHandlerPtr) {
+        return doCatch(exceptionHandlerPtr, () -> {
+            Network network = ObjectHandles.getGlobal().get(networkHandle);
+            String formatStr = CTypeUtil.toString(format);
+            Properties parameters = createParameters(parameterNamesPtrPtr, parameterNamesCount, parameterValuesPtrPtr, parameterValuesCount);
+            MemDataSource dataSource = new MemDataSource();
+            var exporter = Exporters.getExporter(formatStr);
```
geofjamg, line 304: Check exporter exist

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -291,6 +291,35 @@ public static void dumpNetwork(IsolateThread thread, ObjectHandle networkHandle,
...
+            exporter.export(network, parameters, dataSource);
+            try {
+                var exts = dataSource.listNames(".*?");
+                if (exts.size() != 1) {
+                    throw new PowsyblException("Datasource file error");
```
geofjamg: This is not really an error but a non handled case. We probably can do it later (not easy?) we should add a FIXME to explain it.

pypowsybl/network.py
```diff
@@ -59,6 +59,16 @@ def dump(self, file: str, format: str = 'XIIDM', parameters: dict = {}):
         """
         _pypowsybl.dump_network(self.ptr, file, format, parameters)
 
+    def dump_to_string(self, format: str = 'XIIDM', parameters: dict = {}):
+        """Save a network to a file using a specified format.
```
geofjamg: a string

geofjamg (changes requested):

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
geofjamg: et also a clear exception message that explain that will only support string export for single file format

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -291,6 +291,35 @@ public static void dumpNetwork(IsolateThread thread, ObjectHandle networkHandle,
...
+                }
+                String ext = Iterables.getOnlyElement(exts).substring(1);
+                try (InputStream is = new ByteArrayInputStream(dataSource.getData(null, ext));
```
sylvlecl: I think you can use straight the method which takes the full name as input

cpp/src/bindings.cpp
```diff
@@ -99,6 +99,9 @@ PYBIND11_MODULE(_pypowsybl, m) {
     m.def("dump_network", &pypowsybl::dumpNetwork, "Dump network to a file in a given format", py::arg("file"),
           py::arg("format"), py::arg("parameters"), py::arg("parameters"));
 
+    m.def("dump_network_to_string", &pypowsybl::dumpNetworkToString, "Dump network in a given format",
+          py::arg("format"), py::arg("parameters"), py::arg("parameters"));
```
sylvlecl: I think there is one unnecessary "parameters" argument here

pypowsybl/network.py
```diff
@@ -59,6 +59,16 @@ def dump(self, file: str, format: str = 'XIIDM', parameters: dict = {}):
         """
         _pypowsybl.dump_network(self.ptr, file, format, parameters)
 
+    def dump_to_string(self, format: str = 'XIIDM', parameters: dict = {}):
+        """Save a network to a string using a specified format.
+
+        :param format: format to save the network
```
geofjamg: Could you port it with google style ?

geofjamg (approved):

PR #172, opened 2021-07-05
add test networks and refactor some java code

EtienneLt:
Signed-off-by: Etienne Lesot <etienne.lesot@rte-france.com>

Refactoring and tests:
- it adds some examples of network containing batteries and dangling lines, and tests for them
- java code structure: it separates network API from the rest of the API, which starts to grow

### DIFF
--- a/cpp/src/bindings.cpp
+++ b/cpp/src/bindings.cpp
@@ -52,6 +52,10 @@ PYBIND11_MODULE(_pypowsybl, m) {
 
     m.def("create_four_substations_node_breaker_network", &pypowsybl::createFourSubstationsNodeBreakerNetwork, "Create an 4-substation example network");
 
+    m.def("create_battery_network", &pypowsybl::createBatteryNetwork, "Create an example network with batteries");
+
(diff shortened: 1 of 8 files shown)


cpp/src/bindings.cpp
```diff
@@ -52,6 +52,10 @@ PYBIND11_MODULE(_pypowsybl, m) {
 
     m.def("create_four_substations_node_breaker_network", &pypowsybl::createFourSubstationsNodeBreakerNetwork, "Create an 4-substation example network");
 
+    m.def("create_battery_network", &pypowsybl::createBatteryNetwork, "Create an example network with batteries");
```
yichen88, line 55: check this https://github.com/powsybl/pypowsybl/pull/114
preferred use import xml in tests

sylvlecl: Thanks @yichen88 :
just to log what has been done : Etienne has left the creation methods, but "hidden" from the normal user using a leading underscore.

It allows to use those tests networks from unit test while not exposing them directly to the user, which was the initial goal I think : @geofjamg we can change it in the future if really needed.

sylvlecl (approved):

PR #180, opened 2021-07-08
Add network metadata

yichen88:

**Other information**:

### DIFF
--- a/cpp/src/pypowsybl.cpp
+++ b/cpp/src/pypowsybl.cpp
@@ -246,12 +246,11 @@ SeriesArray* createImporterParametersSeriesArray(const std::string& format) {
     return new SeriesArray(callJava<array*>(::createImporterParametersSeriesArray, (char*) format.data()));
 }
 
-main_attributes* getMainAttributes(const JavaHandle& network) {
-    return callJava<main_attributes*>(::getMainAttributes, network);
-}
-
-void freeMainAttributes(main_attributes& main_attributes) {
-    callJava(::freeMainAttributes, &main_attributes);
+std::shared_ptr<network_metadata> getNetworkMetadata(const JavaHandle& network) {
+    network_metadata* attributes = callJava<network_metadata*>(::getNetworkMetadata, network);
+    return std::shared_ptr<network_metadata>(attributes, [](network_metadata* ptr){
+        callJava(::freeNetworkMetadata, ptr);
+    });
 }
 
 JavaHandle loadNetwork(const std::string& file, const std::map<std::string, std::string>& parameters) {
--- a/cpp/src/pypowsybl.cpp
+++ b/cpp/src/pypowsybl.cpp
@@ -246,6 +246,14 @@ SeriesArray* createImporterParametersSeriesArray(const std::string& format) {
     return new SeriesArray(callJava<array*>(::createImporterParametersSeriesArray, (char*) format.data()));
 }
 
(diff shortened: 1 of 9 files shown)


cpp/src/pypowsybl.cpp
```diff
@@ -246,6 +246,14 @@ SeriesArray* createImporterParametersSeriesArray(const std::string& format) {
     return new SeriesArray(callJava<array*>(::createImporterParametersSeriesArray, (char*) format.data()));
 }
 
+main_attributes* getMainAttributes(const JavaHandle& network) {
```
sylvlecl: here we should return a `shared_ptr` that will call `freeMainAttributes` on destruction to free memory

I would prefer "network_metadata" than "main_attributes" to be consistent with other places and because "main" sounds like other data is less important

java/src/main/java/com/powsybl/python/PyPowsyblApiHeader.java
```diff
@@ -64,6 +66,57 @@ private PyPowsyblApiHeader() {
...
+        ptr.setId(CTypeUtil.toCharPtr(networkMetadata.get("id")));
+        ptr.setName(CTypeUtil.toCharPtr(networkMetadata.get("name")));
+        ptr.setSourceFormat(CTypeUtil.toCharPtr(networkMetadata.get("sourceFormat")));
+        ptr.setForecastDistance(Integer.parseInt(networkMetadata.get("forecastDistance")));
+        ptr.setCaseDate(CTypeUtil.toCharPtr(networkMetadata.get("caseDate")));
+        return ptr;
+    }
+
+    static void freeMainAttributePointer(NetworkMetadataPointer networkMetadataPointer) {
+        UnmanagedMemory.free(networkMetadataPointer.getId());
```
geofjamg: Be careful, maybe I am wrong but in all the other places in pypowsybl, I didn't released strings because I think this is done on python side when python string that wrap c side allocated mem are garbage collected.

sylvlecl: After some reading, it seems to me that python strings are always built by copying the input C string (see [python C API doc on strings](https://docs.python.org/3/c-api/unicode.html#creating-and-accessing-unicode-strings), and [pybind11](https://pybind11.readthedocs.io/en/stable/advanced/cast/strings.html#returning-c-strings-to-python) doc).

So maybe we have some memory leaks here!

sylvlecl: Note: if it's really released by python we could have some issues anyway, because graalvm doc states that memory allocated through the graalvm API needs to be deallocated the same way !

geofjamg: I already tried to release strings and it always end up with a crash

sylvlecl: So I did a test in order to confirm or not the leak, see #181 : it does confirm it! 
Probably the crashes you obtained were caused by something else ? (for example at first I had a "free(): invalid pointer" error because my deallocation code was wrong, I released the char** at adress i instead of the char*)

geofjamg: Oups, probably, I don't remenber what was my use case, it was at the very early stage of the pypowsybl prototyping.
Thanks for the fix #181

sylvlecl (approved):

PR #181, opened 2021-07-09
Fix strings memory leaks

sylvlecl:

Bugfix

String allocated by java code are not correctly freed when not used anymore.

**Info**
In order to confirm the leak, I tested the following code on a large network (like 100000 switches):
```python
import pypowsybl.network as pn

network = pn.load('larger_network.xiidm')

for i in range(1, 100):
    df = network.get_switches()
```

Prior to the fix, I can see the memory usage linearly growing, whereas it remains stable after the fix.

### DIFF
--- a/cpp/src/bindings.cpp
+++ b/cpp/src/bindings.cpp
@@ -30,11 +30,6 @@ py::array seriesAsNumpyArray(const series& series) {
     return py::array(py::dtype::of<T>(), series.data.length, series.data.ptr, py::cast(series));
 }
 
-template<typename F>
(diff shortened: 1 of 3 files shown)


sylvlecl:
So we need to fix all the missing "free", but the one in the series is probably the largest leak.

sylvlecl:
Note that after that fix, there are still some unnecessary string copies. In some cases :
- `char*` is first copied to `std::string` in the c to c++ layer
- then `std::string` is copied to python string in the binding layer

We could remove one of those 2 copies, if we identify it as a performance issue.

Issue #148, opened 2021-06-25
Support for per-unit


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the expected behavior?**

We could propose a per-unit view of the network, in order to facilitate the creation of inputs for computation algorithms.

Alternatives could be : 
```python
# propose per-method per-unit
network.get_lines(per_unit=True, sn=100)
network.update_lines(df, per_unit=True, sn=100)

# maybe better: propose a per-unit view of the network:
pu_network = network.per_unit(sn=100)
```

* **What is the motivation / use case for changing the behavior?**

facilitate the creation of inputs for computation algorithms:
most computation tools will need to per-unit the network values in order to have homogeneous orders of magnitudes.
That transformation can be common and not obvious to implement without digging deep into the framework modelling.

PR #193, opened 2021-07-15
Per unit view of the network

EtienneLt:
Signed-off-by: Etienne Lesot <etienne.lesot@rte-france.com>

#148

**Other information**:


pypowsybl/network.py
```diff
@@ -94,6 +100,15 @@ def __setstate__(self, state):
         n = _pypowsybl.load_network_from_string('tmp.xiidm', xml, {})
         self._handle = n
 
+    def activate_per_unit(self):
```
sylvlecl: In order to better separate the code from the base network, and also to be able to manipulate at the same time a per-united view and the "normal" network, I would propose to return a new object here, like a `PerUnitNetwork` which would be a subclass of `Network`:
this way all the per-unit code would go into that class.

Then you could use it this way:
```python
network = pn.load(...)
pu_view = network.per_unit_view(sn=100)
network.get_lines()  # still returns lines not per-united
pu_view.get_lines()  # returns the per-united lines data
```

pypowsybl/network.py
```diff
@@ -94,6 +100,15 @@ def __setstate__(self, state):
...
+        self.per_unit = True
+
+    def desactivate_per_unit(self):
```
sylvlecl: With the separate "view" approaches, no need anymore to disable per-unit or change the nominal S:
you just use the base network or ask for a view per-united with another nominal S.

pypowsybl/perunitnetwork.py
```diff
@@ -0,0 +1,392 @@
...
+        df = df.drop(['nominal_v', 'voltage_level_id'], axis=1)
+        return self.update_elements(_pypowsybl.ElementType.STATIC_VAR_COMPENSATOR, df)
+    
+    def update_hvdc_lines(self, df: pd.DataFrame):
+        """ Update HVDC lines with a ``Pandas`` data frame.
+
+        Args:
+            df (DataFrame): the ``Pandas`` data frame
+        """
+        df['active_power_setpoint'] *= self.sn
```
sylvlecl: in all update methods:
this kind of call will change the user dataframe "in place", which we should probably not do, because he will not expect it.
For example if you call twice this update method with the same dataframe, the second time it will do something weird !

EtienneLt:
ratio for 2 windings transformers is missing

pypowsybl/perunitnetwork.py
```diff
@@ -0,0 +1,615 @@
...
+        dangling_lines['p0'] /= self.sn
+        dangling_lines['q0'] /= self.sn
+        dangling_lines = _pd.merge(dangling_lines, self.get_voltage_levels()['nominal_v'],
+                                   left_on='voltage_level_id', right_index=True)
+        dangling_lines['i'] /= self.sn * 10 ** 3 / (
+                self.sqrt3 * dangling_lines['nominal_v'])
+        dangling_lines['r'] /= dangling_lines['nominal_v'] ** 2 / self.sn
+        dangling_lines['x'] /= dangling_lines['nominal_v'] ** 2 / self.sn
+        dangling_lines['g'] /= dangling_lines['nominal_v'] ** 2 / self.sn
+        dangling_lines['b'] /= dangling_lines['nominal_v'] ** 2 / self.sn
```
Hadrien-Godard: It seems incorrect for g and b. Per unitage for admittance cannot be the same as for impedance. It should be dangling_lines['g'] /= self.sn / dangling_lines['nominal_v'] ** 2. Same for b.

pypowsybl/perunitnetwork.py
```diff
@@ -0,0 +1,615 @@
...
+
+        Args:
+            df (DataFrame): the ``Pandas`` data frame
+        """
+        keys = df.columns
+        to_update = df.copy()
+        if 'p0' in keys:
+            to_update['p0'] *= self.sn
+        if 'q0' in keys:
+            to_update['q0'] *= self.sn
```
Hadrien-Godard: It seems that the updating of r, x, g and b is missing.

pypowsybl/perunitnetwork.py
```diff
@@ -0,0 +1,615 @@
...
+    def update_hvdc_lines(self, df: _pd.DataFrame):
+        """ Update HVDC lines with a ``Pandas`` data frame.
+
+        Args:
+            df (DataFrame): the ``Pandas`` data frame
+        """
+        keys = df.columns
+        to_update = df.copy()
+        if 'active_power_setpoint' in keys:
+            to_update['active_power_setpoint'] *= self.sn
```
Hadrien-Godard: It seems that the updating of r is missing.

EtienneLt: there is no r in static var compensator am I wrong ? but it seems that b field is missing and not only for per unitage.

Hadrien-Godard: Sorry, the comment was for an updated r missing for HVDC lines.

Hadrien-Godard (changes requested):
Hi, I have put some comments on missing updated values or on wrong per unitage for dangling lines. It also seems that the "i" field is not concerned by updating. Is it on purpose ?
I approve the PR as long as the comments are adressed.

EtienneLt:
> Hi, I have put some comments on missing updated values or on wrong per unitage for dangling lines. It also seems that the "i" field is not concerned by updating. Is it on purpose ? I approve the PR as long as the comments are adressed.

"i" is computed with a loadflow, it can not be updated

Hadrien-Godard (approved):

pypowsybl/perunitnetwork.py
```diff
@@ -0,0 +1,636 @@
...
+        Returns:
+            a lines data frame
+        """
+        lines = self.get_elements(_pypowsybl.ElementType.LINE)
+        lines['p1'] /= self.sn
+        lines['p2'] /= self.sn
+        lines['q1'] /= self.sn
+        lines['q2'] /= self.sn
+        lines = _pd.merge(lines, self.get_voltage_levels()['nominal_v'],
+                          left_on='voltage_level1_id', right_index=True)
```
annetill: This should be the nominal voltage on side 2 here.

pypowsybl/perunitnetwork.py
```diff
@@ -0,0 +1,636 @@
...
+                                              left_on='voltage_level2_id', right_index=True)
+        two_windings_transformers['i1'] /= self.sn * 10 ** 3 / (
+                self.sqrt3 * two_windings_transformers['nominal_v'])
+        two_windings_transformers['i2'] /= self.sn * 10 ** 3 / (
+                self.sqrt3 * two_windings_transformers['nominal_v'])
+        two_windings_transformers['r'] /= two_windings_transformers['nominal_v'] ** 2 / self.sn
+        two_windings_transformers['x'] /= two_windings_transformers['nominal_v'] ** 2 / self.sn
+        two_windings_transformers['g'] /= self.sn / two_windings_transformers['nominal_v'] ** 2
+        two_windings_transformers['b'] /= self.sn / two_windings_transformers['nominal_v'] ** 2
+        two_windings_transformers['rated_u1'] /= two_windings_transformers['nominal_v']
```
annetill: Need @sylvlecl opinion here: for me ratedU1 is with the nominal voltage on side 1 and ratedU2 is with the nominal voltage en side 2. It is the only way to have a correct rho in the pi model.

pypowsybl/perunitnetwork.py
```diff
@@ -0,0 +1,636 @@
...
+            a 3 windings transformers data frame
+        """
+        three_windings_transformers = self.get_elements(_pypowsybl.ElementType.THREE_WINDINGS_TRANSFORMER)
+        three_windings_transformers['p1'] /= self.sn
+        three_windings_transformers['q1'] /= self.sn
+        three_windings_transformers['p2'] /= self.sn
+        three_windings_transformers['q2'] /= self.sn
+        three_windings_transformers['p3'] /= self.sn
+        three_windings_transformers['q3'] /= self.sn
+        three_windings_transformers = _pd.merge(three_windings_transformers, self.get_voltage_levels()['nominal_v'],
```
annetill: For each leg, you have to choose the nominal voltage of the side 2 of the leg itself. In our modeling of three windings transformer, for all legs, it is the bus star nominal voltage. So here you have to per-unit with the nominal voltage of the bus star which is indeed the ratedU0 of the three windings transformer. @sylvlecl do you agree ?

pypowsybl/perunitnetwork.py
```diff
@@ -0,0 +1,636 @@
...
+            'nominal_v'] ** 2
+        three_windings_transformers['b2'] /= self.sn / three_windings_transformers[
+            'nominal_v'] ** 2
+        three_windings_transformers['r3'] /= three_windings_transformers['nominal_v'] ** 2 / self.sn
+        three_windings_transformers['x3'] /= three_windings_transformers['nominal_v'] ** 2 / self.sn
+        three_windings_transformers['g3'] /= self.sn / three_windings_transformers[
+            'nominal_v'] ** 2
+        three_windings_transformers['b3'] /= self.sn / three_windings_transformers[
+            'nominal_v'] ** 2
+        three_windings_transformers['rated_u1'] /= three_windings_transformers['nominal_v']
```
annetill: Same comment: for me ratedU1 should be with the nominal voltage of the voltage level 1, ratedU2 with nominal voltage of voltage level 2, ratedU3 with the nominal voltage of voltage level 3. RatedU0 is per-unit using ratedU0 so it is equal to 1. RatedU0 is missing in your PR indeed.

pypowsybl/perunitnetwork.py
sylvlecl: Indeed ! thank you

pypowsybl/perunitnetwork.py
sylvlecl: Yes.
Note that the use of rate_u0 is somewhat arbitrary and "uncommon" compared to everywhere else where we use nominal voltages of the network for per-uniting, so we should mention it in the docstring.

annetill: It is just that we have decided to put star bus at a nominal voltage equals to ratedU0. And yes, it is arbitrary because this bus is fictitious indeed.

pypowsybl/perunitnetwork.py
EtienneLt: on a line nominal voltage is the same on both sides isn't it ?

sylvlecl (approved):

Issue #212, opened 2021-08-12
Network merge


geofjamg:
* **Do you want to request a *feature* or report a *bug*?**
feature

* **What is the motivation / use case for changing the behavior?**
We should be able to merge networks like in Java (See com.powsybl.iidm.network.Network.merge method).
The API should look like:
```python
n1 = pp.network.load('n1.uct')
n2 = pp.network.load('n2.uct')
n3 = pp.network.load('n3.uct')
n1.merge(n2, n3)
```
So with a merge method that accepts one or several networks to absorb.

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

geofjamg:
Solved by #214

PR #214, opened 2021-08-13
Merge networks api

obrix:
- [x ] The commit message follows our guidelines
- [x ] Tests for the changes have been added (for bug fixes / features)

This is the implementation for issue https://github.com/powsybl/pypowsybl/issues/212 

No api to merge networks was available.

An api to merge networks is available, this is the core api wrapped and working the same way.


obrix:
I used a specific C struct to pass an array of ObjectHandles from C++ to java. As discussed with @geofjamg there may be a more direct way to do this by defining and using a Java structure extending PointerBase with  @CPointerTo(VoidPointer.class) annotation. I will test that soon to see if this works. The solution provided here is operational though.

java/src/main/java/com/powsybl/python/PyPowsyblApiHeader.java
```diff
@@ -52,6 +57,17 @@ private PyPowsyblApiHeader() {
         void setLength(int length);
     }
 
+    /* Import of a C function pointer type. */
+    interface GetElementFunctionPointer extends CFunctionPointer {
```
geofjamg: We can remove it?

geofjamg (changes requested):

tests/test_network.py
```diff
@@ -524,6 +524,21 @@ def test_create_network(self):
         n = pp.network.create_ieee118()
         self.assertEqual('ieee118cdf', n.id)
 
+    def test_network_merge(self):
+        n1 = pp.network.load(str(TEST_DIR.joinpath('empty-network.xml')))
```
geofjamg: We have to find a better test case, with mergeable networks.

java/src/main/java/com/powsybl/python/PyPowsyblApiHeader.java
obrix: Yes, I removed it.

geofjamg (approved):

Issue #210, opened 2021-08-05
Clean up public API


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**

_(See comments below, the goal of the issue has slightly changed, but I leave the initial goal here in order not make the discussion senseless)_

Features are only accessible by importing their implementation from inner modules, for example `pypowsybl.network`, `pypowsybl.loadflow`.

* **What is the expected behavior?**

It seems recommended in python (cannot find the reference anymore, will update) to have as much as possible the API in one root package. Indeed, when you used libs such as `numpy` or `pandas` you more or less never need to import submodules.

We could try to follow the same path by importing all widely used classes/methods in `pypowsybl/__init__.py`.

This is open to discussion.

* **What is the motivation / use case for changing the behavior?**

Ease of use / conformance to the language best practices.

For example it would avoid to import multiple modules when you only want to read a file and run a loadflow.

geofjamg:
If you look at some other successful libraries like tensorflow and pytorch there is many subpackages that are use as a namespace like in pypowsybl (this has been my reference design)

sylvlecl:
Yes, the issue is really to discuss pros/cons, it may not be a good fit for pypowsybl.

In numpy and pandas everything is centered around 1 object basically (array and dataframe), so it makes sense to have one main module. When you have multiple very distinct features (like loadflow and dynamic simulation), it may not be that meaningful.

sylvlecl:
But, in any case, the important point is that we should be more careful than we are today about what we expose in our "public" modules, be it only the root package or other modules.

For example, today in `pypowsybl/network.py` we have those imports:
```python
import _pypowsybl
import sys
from _pypowsybl import ElementType
from _pypowsybl import PyPowsyblError
from typing import List
from typing import Set

import pandas as pd
import datetime

from pypowsybl.util import create_data_frame_from_series_array
```
It means that `sys`, `List`, `Set`, `pd` and `datetime` are exposed in the public API of `pypowsybl.network` module. It means that your IDE will propose autocompletion for them, for instance. We should really remove all those artifacts !

Even in libs like tensorflow with submodules, each module is carefully populated with only the relevant names. It's achieved with a combination of `_` prefixed names, of methods and classes selectively imported from internal modules, etc.

sylvlecl:
I close that issue following #217, but it's something we need to keep in mind on the long run.

[...]

PR #217, opened 2021-08-25
clean up public api

EtienneLt:

#210

### DIFF
--- a/pypowsybl/security.py
+++ b/pypowsybl/security.py
@@ -4,38 +4,38 @@
 # iicense, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 #
-from typing import Union, Dict
+from typing import Union as _Union, Dict as _Dict
 from pypowsybl.util import create_data_frame_from_series_array as _create_data_frame_from_series_array
 
-from typing import List
+from typing import List as _List
 
 import _pypowsybl
-from _pypowsybl import ContingencyResult
-from _pypowsybl import LimitViolation
-from _pypowsybl import ContingencyContextType
-from pypowsybl.network import Network
+from _pypowsybl import ContingencyResult as ContingencyResult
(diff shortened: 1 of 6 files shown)


pypowsybl/security.py
```diff
@@ -4,38 +4,38 @@
...
 
-from typing import List
+from typing import List as _List
 
 import _pypowsybl
-from _pypowsybl import ContingencyResult
-from _pypowsybl import LimitViolation
-from _pypowsybl import ContingencyContextType
-from pypowsybl.network import Network
+from _pypowsybl import ContingencyResult as _ContingencyResult
```
sylvlecl: Actually that class is intentionally imported into the `pypowsybl.security` namespace.
The user must be able to see it without importing the native module `_pypowsybl`: instead it must be available in `pypowsybl.security`.

pypowsybl/security.py
```diff
@@ -4,38 +4,38 @@
...
+from _pypowsybl import LimitViolation as _LimitViolation
```
sylvlecl: Same, it must be available in `pypowsybl.security` namespace.

pypowsybl/security.py
```diff
@@ -4,38 +4,38 @@
...
+from _pypowsybl import ContingencyContextType as _ContingencyContextType
```
sylvlecl: Same, it must be available in `pypowsybl.security` namespace.

pypowsybl/security.py
```diff
@@ -4,38 +4,38 @@
...
+from pypowsybl.network import Network as _Network
 from pypowsybl.loadflow import Parameters
```
sylvlecl: That one should not be exposed here, it's already available in `pypowsybl.loadflow`.

sylvlecl (changes requested):
Visibility to be changed for a very few classes

sylvlecl (approved):

Issue #207, opened 2021-07-29
Get and set regulating terminal for generator


vincentbarbesant:
* **Do you want to request a *feature* or report a *bug*?**
A feature

* **What is the current behavior?**
The dataframe for generators does not return which terminal the voltage regulation of the generator is regulating
```
Data columns (total 12 columns):
 #   Column                Non-Null Count  Dtype  
---  ------                --------------  -----  
 0   energy_source         2768 non-null   object 
 1   target_p              2768 non-null   float64
 2   max_p                 2768 non-null   float64
 3   min_p                 2768 non-null   float64
 4   target_v              2768 non-null   float64
 5   target_q              2768 non-null   float64
 6   voltage_regulator_on  2768 non-null   bool   
 7   p                     2768 non-null   float64
 8   q                     1435 non-null   float64
 9   voltage_level_id      2768 non-null   object 
 10  bus_id                2768 non-null   object 
 11  genreCvg              2768 non-null   object 
```
* **What is the expected behavior?**
I want to be able to get and set the regulating terminal (possibly a remote one) for any generator

* **Please tell us about your environment:**
  - PowSyBl Version: 0.8.0
  - OS Version: Windows 10 Entreprise 64-bit (10.0, Build 18363) (18362.19h1_release.190318-1202)

sylvlecl:
For the setter, we need to decide what will be the input : 

1.  any identifiable (generator, load, busbar) --> in that case the corresponding terminal will be used
2.  in bus breaker topology, this will not be very natural : we would like to be able to regulate a specific bus

However, 2. is a design issue in `powsybl-core`, not specific to pypowsybl, so we can go for 1. and handle it later.

sylvlecl:
Implemented in #218 for node breaker networks only.

Issue #296 has been created for bus breaker implementation, and to discuss the corresponding design issues.

PR #218, opened 2021-08-26
Enable to set regulating terminal for generators, in node breaker topology

EtienneLt:

#207


sylvlecl (approved):
For me the implementation is OK.

However it differs from what we have done for tap changers, where we have used a bus instead (`regulating_bus_id`).

@geofjamg : do you think it's consistent with IIDM model to use the connectable ID as the "regulating terminal" identifier ? The javadoc says that a terminal is the connection point of an equipment, so for me it makes sense.

sylvlecl:
For now, only node breaker will be supported for this feature.
- we will only allow to set busbar sections and injection terminals as the regulation location

Bus breaker raises the following issues:

- should we expose the ID of the connectable associated to the regulated terminal, or the corresponding connectable bus ID
- for modification, should we allow connectable IDs ? Should we allow buses IDs ? If we allow both, it will give weird results because the getter will not return the same thing that was set for both cases.
- if we allow to set buses, pypowsybl does not yet allow to easily access the list of "bus breaker" buses, so it will make the feature hard to use
- if we don't, we cannot regulate voltage on buses where only branches are connected

PR #221, opened 2021-09-10
[WIP] Possibility to get buses from the bus breaker view

EtienneLt:

Feature

In all tables, we can chose between buses from the bus view or from the bus-breaker view.
For example:
```python
buses_bus_breaker_view = network.get_buses(bus_breaker_view=True)
```

### DIFF
--- a/java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
+++ b/java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
@@ -37,29 +37,34 @@ import static com.powsybl.dataframe.MappingUtils.ifExistsInt;
 public final class NetworkDataframes {
 
     private static final Map<DataframeElementType, NetworkDataframeMapper> MAPPERS = createMappers();
+    private static final Map<DataframeElementType, NetworkDataframeMapper> BUS_BREAKER_VIEW_MAPPERS = createMappersBusBreakerView();
 
     private NetworkDataframes() {
     }
 
     public static NetworkDataframeMapper getDataframeMapper(DataframeElementType type) {
-        return MAPPERS.get(type);
+        return getDataframeMapper(type, false);
+    }
+
+    public static NetworkDataframeMapper getDataframeMapper(DataframeElementType type, boolean busBreakerView) {
+        return busBreakerView ? BUS_BREAKER_VIEW_MAPPERS.get(type) : MAPPERS.get(type);
     }
 
     private static Map<DataframeElementType, NetworkDataframeMapper> createMappers() {
         Map<DataframeElementType, NetworkDataframeMapper> mappers = new EnumMap<>(DataframeElementType.class);
-        mappers.put(DataframeElementType.BUS, buses());
-        mappers.put(DataframeElementType.LINE, lines());
-        mappers.put(DataframeElementType.TWO_WINDINGS_TRANSFORMER, twoWindingTransformers());
-        mappers.put(DataframeElementType.THREE_WINDINGS_TRANSFORMER, threeWindingTransformers());
-        mappers.put(DataframeElementType.GENERATOR, generators());
-        mappers.put(DataframeElementType.LOAD, loads());
-        mappers.put(DataframeElementType.BATTERY, batteries());
-        mappers.put(DataframeElementType.SHUNT_COMPENSATOR, shunts());
+        mappers.put(DataframeElementType.BUS, buses(false));
+        mappers.put(DataframeElementType.LINE, lines(false));
+        mappers.put(DataframeElementType.TWO_WINDINGS_TRANSFORMER, twoWindingTransformers(false));
+        mappers.put(DataframeElementType.THREE_WINDINGS_TRANSFORMER, threeWindingTransformers(false));
+        mappers.put(DataframeElementType.GENERATOR, generators(false));
+        mappers.put(DataframeElementType.LOAD, loads(false));
+        mappers.put(DataframeElementType.BATTERY, batteries(false));
+        mappers.put(DataframeElementType.SHUNT_COMPENSATOR, shunts(false));
         mappers.put(DataframeElementType.NON_LINEAR_SHUNT_COMPENSATOR_SECTION, shuntsNonLinear());
-        mappers.put(DataframeElementType.DANGLING_LINE, danglingLines());
-        mappers.put(DataframeElementType.LCC_CONVERTER_STATION, lccs());
-        mappers.put(DataframeElementType.VSC_CONVERTER_STATION, vscs());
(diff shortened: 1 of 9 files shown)


java/src/main/java/com/powsybl/dataframe/network/NetworkDataFramesBusBreakerView.java
```diff
@@ -0,0 +1,512 @@
...
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.ToDoubleFunction;
+import java.util.function.ToIntFunction;
+import java.util.stream.Stream;
+
+import static com.powsybl.dataframe.MappingUtils.ifExistsDouble;
+import static com.powsybl.dataframe.MappingUtils.ifExistsInt;
+
+public final class NetworkDataFramesBusBreakerView {
```
geofjamg: This class add a lot of code duplication, I would keep everything in  NetworkDataFrames with 2 BUS_BREAKER_VIEW_MAPPER in addition to MAPPERS passing a boolean to createMappers(boolean busBreakerView) and all elements related methods generators(boolean busBreakerView)

tests/test_network.py
```diff
@@ -121,26 +121,35 @@ def test_get_network_element_ids(self):
 
     def test_buses(self):
         n = pp.network.create_eurostag_tutorial_example1_network()
-        buses = n.get_buses()
+        buses_bus_breaker_view = n.get_buses()
```
geofjamg: this is not bus_breaker_view by default?

EtienneLt: yes by default it is not busbreakerview but I can change it

java/src/main/java/com/powsybl/dataframe/network/NetworkDataFramesBusBreakerView.java
EtienneLt: I don't know how I could pass the boolean to createmapper() because it is created once, Am I wrong ?

EtienneLt: You mean that mappers will be recreated each time we call get_elements() ?

geofjamg: No you can just have 2 caches map, one for busview and an other one for busbreakerview:
```java
    public static NetworkDataframeMapper getDataframeMapper(DataframeElementType type, boolean busBreakerView) {
        return busBreakerView ? BUS_BREAKER_VIEW_MAPPERS.get(type) : MAPPERS.get(type);
    }
```

java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
```diff
@@ -178,15 +202,17 @@ static NetworkDataframeMapper generators() {
                 .doubles("q", getQ(), setQ())
                 .doubles("i", g -> g.getTerminal().getI())
                 .strings("voltage_level_id", getVoltageLevelId())
-                .strings("bus_id", g -> getBusId(g.getTerminal()))
+                .strings("bus_id", g -> getBusId(g.getTerminal()), g -> getBusIdWithBusBreakerView(g.getTerminal()), busBreakerView)
```
sylvlecl: The test on which getter to use would better take place here instead of adding a method to the builder:
you could probably define something like 
```java
<some return type> busGetter(boolean busBreakerView) {
    return busBreakerView ? g -> getBusId(g.getTerminal()) : g -> getBusBreakerBusId(g.getTerminal()));
}
```

Then you can just use that method everywhere:
```java
    .strings("bus_id", busGetter(busBreakerView))
```

java/src/main/java/com/powsybl/dataframe/BaseDataframeMapperBuilder.java
```diff
@@ -84,6 +84,14 @@ public B strings(String name, Function<U, String> value) {
         return strings(name, value, null);
     }
 
+    public B strings(String name, Function<U, String> value, Function<U, String> value2, boolean getSecondValue) {
```
sylvlecl: It's not really the best place for that test, see below, we should decide which getter to use before calling that method

java/src/main/java/com/powsybl/python/PyPowsyblNetworkApiLib.java
```diff
@@ -320,9 +320,11 @@ public static void removeVariant(IsolateThread thread, ObjectHandle networkHandl
 
     @CEntryPoint(name = "createNetworkElementsSeriesArray")
     public static PyPowsyblApiHeader.ArrayPointer<PyPowsyblApiHeader.SeriesPointer> createNetworkElementsSeriesArray(IsolateThread thread, ObjectHandle networkHandle,
-                                                                                                                     PyPowsyblApiHeader.ElementType elementType, PyPowsyblApiHeader.ExceptionHandlerPointer exceptionHandlerPtr) {
+                                                                                                                     PyPowsyblApiHeader.ElementType elementType, boolean busBreakerView,
+                                                                                                                     PyPowsyblApiHeader.ExceptionHandlerPointer exceptionHandlerPtr) {
         return doCatch(exceptionHandlerPtr, () -> {
-            NetworkDataframeMapper mapper = NetworkDataframes.getDataframeMapper(convert(elementType));
+            NetworkDataframeMapper mapper;
+            mapper = NetworkDataframes.getDataframeMapper(convert(elementType), busBreakerView);
```
sylvlecl: we should keep only one line here

pypowsybl/network.py
```diff
@@ -214,75 +214,78 @@ def get_elements_ids(self, element_type: _pypowsybl.ElementType, nominal_voltage
...
         Returns:
             a network elements data frame for the specified element type
         """
-        series_array = _pypowsybl.create_network_elements_series_array(self._handle, element_type)
+        series_array = _pypowsybl.create_network_elements_series_array(self._handle, element_type, bus_breaker_view)
         return create_data_frame_from_series_array(series_array)
 
-    def get_buses(self) -> pd.DataFrame:
-        return self.get_elements(_pypowsybl.ElementType.BUS)
+    def get_buses(self, bus_breaker_view=False) -> pd.DataFrame:
```
sylvlecl: With that new argument to all methods, I'm afraid we need to update all the docstrings!!

sylvlecl (changes requested):
Suggestion to change an implementation detail + update docstrings

sylvlecl:
In standby for the moment:
although the possibility to get info from the bus breaker view seems interesting, there is no real demand for now, and it's not obvious that the best solution is to add a parameter to all methods.

sylvlecl:
Closing it, this should be replaced by #320

Issue #225, opened 2021-09-17
Enable update of multi-indexed dataframes


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature/refactoring.

* **What is the current behavior?**

The `DataframeMapper` API on the java side does not allow to update multi-indexed dataframes, such as current limits or tap changers steps.

* **What is the expected behavior?**

The user should be able to update those data like others.

That requires to refactor a little the `DataframeMapper` update design. For now, the update take series as inputs (`DoubleIndexedSeries`, `IntIndexedSeries`, `IndexedSeries<String>`), which only expose `String` as their index. 

Instead, the update method could take as input a full table (a list of named columns).
Then, it's the role of the mapper to be able to map each line onto a network element, based on the columns he wants.

* **What is the motivation / use case for changing the behavior?**

Being able to update network elements that cannot be updated today, such as current limits and tap changers steps.

sylvlecl:
Implemented in #234

PR #234, opened 2021-09-29
Manage multi index dataframes

EtienneLt:

#225


cpp/src/bindings.cpp
```diff
@@ -23,6 +22,55 @@ void bindArray(py::module_& m, const std::string& className) {
...
 }
 
+array* createArray(py::list columnsValues, py::list columnsNames, py::list columnsTypes, int columnSize, int columnsNumber) {
+    array* dataframe = new array();
+    dataframe->ptr = new series[columnsNumber];
+    series* columns = new series[columnsNumber];
+    for (int indice = 0 ; indice < columnsNumber ; indice ++ ) {
+        series* column = columns + indice;
+        py::str name = (py::str) columnsNames[indice];
+        column->name = strdup((char*) const_cast<char*>(((std::string) name).c_str()));
```
sylvlecl: we don't need the `const_cast` anymore with the string copy (`strdup`).

cpp/src/bindings.cpp
```diff
@@ -23,6 +22,55 @@ void bindArray(py::module_& m, const std::string& className) {
...
+        } else if (properType == 2 || properType == 3) {
+            std::vector<int> values = py::cast<std::vector<int>>(columnsValues[indice]);
+            column->data.ptr = pypowsybl::copyVectorInt(values);
+        }
+    }
+    dataframe->ptr = columns;
+    return dataframe;
+}
+
+void updateNetworkElementsWithSeries(pypowsybl::JavaHandle network, py::list columnsValues, py::list columnsNames, py::list columnsTypes, py::list indexList, int columnSize, int columnsNumber, element_type elementType, int indexCount) {
```
sylvlecl: Actually we could simplify a little the code by using c++ type here instead of py::list, when we know exactly what is expected:
- columnsNames and indexList --> `const std::vector<std::string>&`
- columnsTypes --> `const std::vector<int>&`

Only for the values it's necessary to use a `py::list` because values in it are heterogeneous

cpp/src/bindings.cpp
```diff
@@ -23,6 +22,55 @@ void bindArray(py::module_& m, const std::string& className) {
...
+            pypowsybl::deleteCharPtrPtr((char**) columns->data.ptr, columnSize);
+        } else if (columns->type == 1) {
+            delete (double*) columns->data.ptr;
+        } else if (columns->type == 2) {
+            delete (int*) columns->data.ptr;
+        }
+        delete columns->name;
+    }
+    delete dataframe->ptr;
+    delete dataframe;
```
sylvlecl: We need to handle the case where the update will throw an exception, so that the .
For that the best option is probably to wrap the `array*` in a `std::shared_ptr<array>`  which will be destroyed when we leave the scope (exception or not).

cpp/src/bindings.cpp
```diff
@@ -431,17 +479,13 @@ PYBIND11_MODULE(_pypowsybl, m) {
...
-          py::arg("network"), py::arg("element_type"), py::arg("series_name"), py::arg("ids"), py::arg("values"),
-          py::arg("element_count"));
-
-    m.def("update_network_elements_with_double_series", &pypowsybl::updateNetworkElementsWithDoubleSeries, "Update network elements for a given element type with a double series",
-          py::arg("network"), py::arg("element_type"), py::arg("series_name"), py::arg("ids"), py::arg("values"),
-          py::arg("element_count"));
+    m.def("get_index_type", &pypowsybl::getIndexType, "Get index type integer for a given element type, index_name or index in the dataframe",
+            py::arg("element_type"), py::arg("series_name"), py::arg("index"));
+    
+    m.def("update_network_elements_with_series", ::updateNetworkElementsWithSeries, "Update network elements for a given element type with a series",
```
sylvlecl, line 477: If this one is not needed anymore we should delete it

cpp/src/bindings.cpp
```diff
@@ -23,6 +22,55 @@ void bindArray(py::module_& m, const std::string& className) {
...
 
+array* createArray(py::list columnsValues, py::list columnsNames, py::list columnsTypes, int columnSize, int columnsNumber) {
+    array* dataframe = new array();
+    dataframe->ptr = new series[columnsNumber];
+    series* columns = new series[columnsNumber];
+    for (int indice = 0 ; indice < columnsNumber ; indice ++ ) {
+        series* column = columns + indice;
+        py::str name = (py::str) columnsNames[indice];
+        column->name = strdup((char*) const_cast<char*>(((std::string) name).c_str()));
+        column->index = 0;
```
sylvlecl: actually here we could set it to true/false according to the index list, that way on the java side we will know which series is an index without a separate list.

cpp/src/bindings.cpp
```diff
@@ -23,6 +22,55 @@ void bindArray(py::module_& m, const std::string& className) {
...
+    char** charIndexList = pypowsybl::copyVectorStringToCharPtrPtr(py::cast<std::vector<std::string>>(indexList));
+    pypowsybl::updateNetworkElementsWithSeries(network, dataframe, charIndexList, columnSize, columnsNumber, elementType, indexCount);
+    pypowsybl::deleteCharPtrPtr(charIndexList, indexCount);
+    for (int indice = 0 ; indice < columnsNumber ; indice ++ ) {
+        series* columns = ((series*) dataframe->ptr) + indice;
+        if (columns->type == 0) {
+            pypowsybl::deleteCharPtrPtr((char**) columns->data.ptr, columnSize);
+        } else if (columns->type == 1) {
+            delete (double*) columns->data.ptr;
+        } else if (columns->type == 2) {
```
sylvlecl: boolean case is missing

cpp/src/bindings.cpp
```diff
@@ -23,6 +22,55 @@ void bindArray(py::module_& m, const std::string& className) {
...
+    return dataframe;
+}
+
+void updateNetworkElementsWithSeries(pypowsybl::JavaHandle network, py::list columnsValues, py::list columnsNames, py::list columnsTypes, py::list indexList, int columnSize, int columnsNumber, element_type elementType, int indexCount) {
+    array* dataframe = createArray(columnsValues, columnsNames, columnsTypes, columnSize, columnsNumber);
+    char** charIndexList = pypowsybl::copyVectorStringToCharPtrPtr(py::cast<std::vector<std::string>>(indexList));
+    pypowsybl::updateNetworkElementsWithSeries(network, dataframe, charIndexList, columnSize, columnsNumber, elementType, indexCount);
+    pypowsybl::deleteCharPtrPtr(charIndexList, indexCount);
+    for (int indice = 0 ; indice < columnsNumber ; indice ++ ) {
+        series* columns = ((series*) dataframe->ptr) + indice;
```
sylvlecl: maybe `column` (not plural) for clarity

sylvlecl (commented):
First comments for the C++ side

java/src/main/java/com/powsybl/dataframe/network/NetworkDataframeMapperBuilder.java
```diff
@@ -28,10 +29,21 @@ public NetworkDataframeMapperBuilder() {
         this.addProperties = false;
     }
 
+    @FunctionalInterface
+    public interface TriFunction<T, U, I> {
```
sylvlecl: We should make that class more specific and explicit:
 - give it an explicit name like `ItemGetter`
 - remove the `T` and `U` parameters which will always be `Network` and `UpdatingDataframe`
 - also name the method more explicitly like `getItem`
 - name the parameters more explicitly, for example `lineNumber` for the last parameter

java/src/main/java/com/powsybl/dataframe/BaseDataframeMapperBuilder.java
```diff
@@ -22,6 +25,7 @@
 
     protected Function<T, List<U>> itemsProvider;
     protected BiFunction<T, String, U> itemGetter;
+    protected NetworkDataframeMapperBuilder.TriFunction<T, UpdatingDataframe, U> itemMultiIndexGetter;
```
sylvlecl: That class is supposed to be more generic than just for `Network` (for example it's used also for security analysis results): any network specific class should not appear here!

java/src/main/java/com/powsybl/dataframe/BaseDataframeMapperBuilder.java
```diff
@@ -22,6 +25,7 @@
 
     protected Function<T, List<U>> itemsProvider;
     protected BiFunction<T, String, U> itemGetter;
```
sylvlecl: We should remove that one, since it is replaced by the new more generic getter.

In order to keep the existing `itemGetter` method, you need to change its implementation so that it creates the new kind of getter fro the older simpler one.

java/src/main/java/com/powsybl/dataframe/update/UpdatingDataframe.java
```diff
@@ -0,0 +1,48 @@
...
+
+    void addStringSerie(Serie stringSerie);
+
+    void addDoubleSerie(Serie doubleSerie);
+
+    void addColumnName(String name, String type);
+
+    List<Pair<String, String>> getOrderColumnNames();
+
+    Double getDoubleValue(String column, int index);
```
sylvlecl: we should use native types for `double` and `int`, for performance

java/src/main/java/com/powsybl/dataframe/update/UpdatingDataframe.java
```diff
@@ -0,0 +1,48 @@
+package com.powsybl.dataframe.update;
+
+import org.apache.commons.lang3.tuple.Pair;
+import java.util.List;
+import java.util.Set;
+
+public interface UpdatingDataframe {
+
+    void addIndex(String index);
```
sylvlecl: The best design would be not to have the setters/adders methods in the interface, we only need the getters which will be used by the mapper.
The setters/adders need only to be known at build time, where we know the actual implementation class.

java/src/main/java/com/powsybl/dataframe/update/UpdatingDataframe.java
```diff
@@ -0,0 +1,48 @@
+package com.powsybl.dataframe.update;
+
+import org.apache.commons.lang3.tuple.Pair;
+import java.util.List;
+import java.util.Set;
+
+public interface UpdatingDataframe {
```
sylvlecl, line 16: Missing author/license on all new files

sylvlecl (commented):
Some more comments for the java side (still in progress :) )

java/src/main/java/com/powsybl/python/update/UpdatingDataframeImpl.java
```diff
@@ -0,0 +1,155 @@
...
+ */
+
+import com.powsybl.dataframe.update.Serie;
+import com.powsybl.dataframe.update.UpdatingDataframe;
+import com.powsybl.python.CTypeUtil;
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.*;
+
+public class UpdatingDataframeImpl implements UpdatingDataframe {
```
sylvlecl: we can name it something like `CUpdatingDataframe` so that it's clear that it's a wrapper around C data (more specific thant just "impl")

java/src/main/java/com/powsybl/dataframe/update/Serie.java
```diff
@@ -0,0 +1,10 @@
+package com.powsybl.dataframe.update;
+
+public interface Serie<T> {
```
sylvlecl: typo : I know it's strange but the series word always has a "s" !

then: that interface does not seem to be used in the "dataframe" package , but is only part of the C implementation --> it should go into "python" instead

java/src/main/java/com/powsybl/dataframe/update/UpdatingDataframe.java
```diff
@@ -0,0 +1,48 @@
...
+
+    void addColumnName(String name, String type);
+
+    List<Pair<String, String>> getOrderColumnNames();
+
+    Double getDoubleValue(String column, int index);
+
+    Double getDoubleValue(int column, int index);
+
+    Double getDoubleValue(String columnName, int column, int index);
```
sylvlecl: at least for those methods with the 2 arguments, we would need some javadoc so that other developers understand what it does (not obvious from the name and arguments)

java/src/main/java/com/powsybl/dataframe/update/UpdatingDataframe.java
```diff
@@ -0,0 +1,48 @@
...
+
+    Integer getIntValue(String column, int index);
+
+    Integer getIntValue(int column, int index);
+
+    Integer getIntValue(String columnName, int column, int index);
+
+    int getSize();
+
+    Set<String> getDoubleSeriesNames();
```
sylvlecl: I think those 3 methods are not useful in the interface

java/src/main/java/com/powsybl/dataframe/update/UpdatingDataframe.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.apache.commons.lang3.tuple.Pair;
+import java.util.List;
+import java.util.Set;
+
+public interface UpdatingDataframe {
+
+    void addIndex(String index);
+
+    List<String> getIndexList();
```
sylvlecl: See comment on the c++ side: actually there we have a boolean that says which column is an index:
I think it would be more natural to have a boolean for each column which says if it's part of the index or not, instead of that list.

java/src/main/java/com/powsybl/python/PyPowsyblNetworkApiLib.java
```diff
@@ -328,81 +330,46 @@ public static void removeVariant(IsolateThread thread, ObjectHandle networkHandl
...
-
-            @Override
-            public String getId(int index) {
-                return CTypeUtil.toString(elementIdPtrPtr.read(index));
-            }
-
-            @Override
-            public double getValue(int index) {
-                return valuePtr.read(index);
+    public static UpdatingDataframeImpl createDataframe(PyPowsyblApiHeader.ArrayPointer<PyPowsyblApiHeader.SeriesPointer> dataframe, int columnsNumber, int elementCount, CCharPointerPointer indexList, int indexCount) {
```
sylvlecl: not very important but it's more usual to use the interface as return type (that way we know that nothing specific to that implementation will be used after).

java/src/main/java/com/powsybl/python/update/DoubleSerie.java
```diff
@@ -0,0 +1,37 @@
+/**
+ * Copyright (c) 2021, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.python.update;
+/**
+ * @author Etienne Lesot {@literal <etienne.lesot at rte-france.com>}
```
sylvlecl: javadoc misplaced

java/src/main/java/com/powsybl/python/update/IntSerie.java
```diff
@@ -0,0 +1,37 @@
+/**
+ * Copyright (c) 2021, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.python.update;
+/**
+ * @author Etienne Lesot {@literal <etienne.lesot at rte-france.com>}
```
sylvlecl: javadoc misplaced

java/src/main/java/com/powsybl/python/update/UpdatingDataframeImpl.java
```diff
@@ -0,0 +1,155 @@
...
+    private final int size;
```
sylvlecl: we should probably name it more explicitly like `lineCount`

java/src/main/java/com/powsybl/dataframe/update/UpdatingDataframe.java
```diff
@@ -0,0 +1,48 @@
...
+
+    void addIntSerie(Serie intSerie);
+
+    void addStringSerie(Serie stringSerie);
+
+    void addDoubleSerie(Serie doubleSerie);
+
+    void addColumnName(String name, String type);
+
+    List<Pair<String, String>> getOrderColumnNames();
```
sylvlecl: Here I would suggest to use the class `SeriesMetadata` which already exists to carry the name/type/index information of a series, instead of the pair.
Also as it's a list, no need to include "ordered" in the name, it's redundant

java/src/main/java/com/powsybl/python/update/UpdatingDataframeImpl.java
```diff
@@ -0,0 +1,155 @@
...
+        this.stringSeries.put(stringSerie.getName(), (StringSerie) stringSerie);
+    }
+
+    @Override
+    public void addDoubleSerie(Serie doubleSerie) {
+        this.doubleSeries.put(doubleSerie.getName(), (DoubleSerie) doubleSerie);
+    }
+
+    public Serie getSeriesOfDataframeWithIndex(int indice) {
+        switch (orderColumns.get(indice).getRight()) {
```
sylvlecl: it would be simpler and more efficient to store the series also in a list, on construction, to just retrieve them with 
```java
series.get(indice)
```

pypowsybl/network.py
```diff
@@ -419,24 +421,38 @@ def update_elements(self, element_type: _pypowsybl.ElementType, df: _pd.DataFram
             element_type (ElementType): the element type
             df (DataFrame): the ``Pandas`` data frame
         """
+        index_names = []
+        columns_names = []
+        columns_values = []
+        columns_types = []
+        column_size = len(df.index.values)
+        index_count = 0
```
sylvlecl, line 1080: we don't need it, just call `len(index_names)` at the end.

Note: maybe it would be simpler to pass a list of booleans, that way we provide to the c++ side lists which have all the same size:
names, values, types, is_index

setup.py
```diff
@@ -93,7 +93,7 @@ def build_extension(self, ext):
                       '-DPYTHON_EXECUTABLE=' + sys.executable,
                       '-DEXTRA_JARS=' + extra_jars]
 
-        cfg = 'Debug' if self.debug else 'Release'
+        cfg = 'Debug' # if self.debug else 'Release'
```
sylvlecl: we must not forget to remove this before merge

pypowsybl/network.py
```diff
@@ -419,24 +421,38 @@ def update_elements(self, element_type: _pypowsybl.ElementType, df: _pd.DataFram
...
+        for index_name in df.index.names:
+            index_column = []
+            for x in df.index.values:
```
sylvlecl: not sure what this for loop is meant to do: I think we should remove it ? It seems that it will impact performances (we are making copies of all the values in a new list, which will not be efficient for numeric types)

pypowsybl/network.py
```diff
@@ -419,24 +421,38 @@ def update_elements(self, element_type: _pypowsybl.ElementType, df: _pd.DataFram
...
+                if type(x) == str or type(x) == int or type(x) == bool or type(x) == float:
+                    index_column.append(x)
+                else:
+                    index_column.append(x[index_count])
+            if index_name is None:
+                index_name = ''
+            columns_names.append(index_name)
+            index_names.append(index_name)
+            columns_types.append(_pypowsybl.get_index_type(element_type, index_name, index_count))
+            columns_values.append(np.array(index_column))
```
sylvlecl: here we should simply use `index.values.to_numpy()` I think for simple index, then for multi index it seems that we can have access to the colums by `df.index.levels[num].to_numpy()`. Not sure the `to_numpy` are necessary actually.

pypowsybl/network.py
```diff
@@ -419,24 +421,38 @@ def update_elements(self, element_type: _pypowsybl.ElementType, df: _pd.DataFram
...
-                                                                      series.values, len(series))
-            elif series_type == 0:
-                _pypowsybl.update_network_elements_with_string_series(self._handle, element_type, series_name,
-                                                                      df.index.values,
-                                                                      series.values, len(series))
-            else:
-                raise _PyPowsyblError(
-                    f'Unsupported series type {series_type}, element type: {element_type}, series_name: {series_name}')
+            columns_types.append(series_type)
+            columns_values.append(np.array(series))
```
sylvlecl: I think we can just use `series.values` as before : by default, the array constructor makes a copy which we don't want for performance purpose.

cpp/src/pypowsybl.cpp
```diff
@@ -202,6 +202,18 @@ char** copyVectorStringToCharPtrPtr(const std::vector<std::string>& strings) {
     return charPtrPtr;
 }
 
+int* copyVectorInt(const std::vector<int>& ints) {
```
sylvlecl, line 205: Note for later (last thing to do):  we'll need to try to get rid of those copies

cpp/src/bindings.cpp
sylvlecl: We still need to do this :)

pypowsybl/network.py
```diff
@@ -552,9 +552,10 @@ def get_non_linear_shunt_compensator_sections(self) -> _DataFrame:
             A dataframe of non linear model shunt compensators sections.
         """
         return self.get_elements(_pypowsybl.ElementType.NON_LINEAR_SHUNT_COMPENSATOR_SECTION)
+
     def get_linear_shunt_compensator_sections(self) -> _DataFrame:
         """
-        Get a dataframe of shunt compensators sections for linear model.
+        Get shunt compensators sections for linear model as a ``Pandas`` data frame.
```
sylvlecl: To do: revert to docstring of main branch

java/src/main/java/com/powsybl/dataframe/update/UpdatingDataframe.java
```diff
@@ -0,0 +1,57 @@
...
+import com.powsybl.dataframe.SeriesMetadata;
+
+import java.util.List;
+
+/**
+ * @author Etienne Lesot {@literal <etienne.lesot at rte-france.com>}
+ */
+public interface UpdatingDataframe {
+
+    List<SeriesMetadata> getColumnNames();
```
sylvlecl: To be renamed in consistence with the return type

cpp/src/bindings.cpp
```diff
@@ -23,6 +22,51 @@ void bindArray(py::module_& m, const std::string& className) {
             }, py::keep_alive<0, 1>());
 }
 
+array* createArray(py::list columnsValues, const std::vector<std::string>& columnsNames, const std::vector<int>& columnsTypes, int columnSize, int columnsNumber, const std::vector<int>& indexBool) {
```
sylvlecl: last arg could be a vector of booleans

java/src/main/java/com/powsybl/dataframe/AbstractDataframeMapper.java
```diff
@@ -49,31 +50,36 @@ public void createDataframe(T object, DataframeHandler dataframeHandler) {
...
-        }
-    }
-
-    @Override
-    public void updateStringSeries(T object, String seriesName, IndexedSeries<String> values) {
-        SeriesMapper<U> series = seriesMappers.get(seriesName);
-        for (int i = 0; i < values.getSize(); i++) {
-            series.updateString(getItem(object, values.getId(i)), values.getValue(i));
-        }
+    private U getItemWithChosenIndex(int index, T object, UpdatingDataframe updatingDataframe) {
```
sylvlecl: I think we can remove that intermediate method

java/src/test/java/com/powsybl/dataframe/network/NetworkDataframesTest.java
```diff
@@ -85,9 +84,9 @@ void generators() {
         assertThat(series.get(3).getDoubles())
             .containsExactly(607);
 
-        NetworkDataframes.generators().updateDoubleSeries(network, "target_p", createInput(List.of("GEN"), 500));
+        // NetworkDataframes.generators().updateDoubleSeries(network, "target_p", createInput(List.of("GEN"), 500));
```
sylvlecl: Let's delete it or adapt it

java/src/main/java/com/powsybl/python/update/CUpdatingDataframe.java
```diff
@@ -0,0 +1,112 @@
...
+public class CUpdatingDataframe implements UpdatingDataframe {
+    private final int lineCount;
+    private final List<SeriesMetadata> columns = new ArrayList<>();
+    private final Map<String, Series> seriesMap = new HashMap<>();
+
+    public CUpdatingDataframe(int lineCount) {
+        this.lineCount = lineCount;
+    }
+
+    public void addColumnName(SeriesMetadata seriesMetadata) {
```
sylvlecl: Addition of series and metadata are always done at the same time:
would be simpler to have 1 method only

java/src/main/java/com/powsybl/python/update/CUpdatingDataframe.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    public List<SeriesMetadata> getColumnNames() {
+        return columns;
+    }
+
+    public void addSeries(Series series) {
+        this.seriesMap.put(series.getName(), series);
+    }
+
+    public Series getSeries(int indice) {
```
sylvlecl: Can probably be private

java/src/main/java/com/powsybl/python/update/CUpdatingDataframe.java
```diff
@@ -0,0 +1,111 @@
...
+    public double getDoubleValue(int columnNumber, int index) {
+        return ((DoubleSeries) getSeries(columnNumber)).getValues().read(index);
+    }
+
+    @Override
+    public int getLineCount() {
+        return lineCount;
+    }
+
+    private boolean containsColumnName(String columnName, SeriesDataType type) {
```
sylvlecl, line 104: It will not be very efficient to traverse the list for each row:
maybe we should have a method like `OptionalInt findColumn(String name, SeriesDataType type)` to find the column number only once for each column.

cpp/src/pypowsybl.cpp
sylvlecl: Tried to directly cast the python object to a pointer, to be able to postpone its deletion and avoid the copy, but id did not work (refused by pybind).
Let's postpone that issue, it may not matter very much...

java/src/main/java/com/powsybl/dataframe/AbstractDataframeMapper.java
```diff
@@ -49,31 +50,36 @@ public void createDataframe(T object, DataframeHandler dataframeHandler) {
...
+                    switch (column.getType()) {
+                        case STRING:
+                            series.updateString(getItemWithChosenIndex(i, object, updatingDataframe),
+                                    updatingDataframe.getStringValue(seriesName, i));
+                            break;
+                        case DOUBLE:
+                            series.updateDouble(getItemWithChosenIndex(i, object, updatingDataframe),
+                                    updatingDataframe.getDoubleValue(seriesName, i));
+                            break;
+                        case INT:
```
sylvlecl, line 68: Looks like boolean case is missing, but it did not break the test?

java/src/main/java/com/powsybl/python/PyPowsyblNetworkApiLib.java
```diff
@@ -328,81 +333,43 @@ public static void removeVariant(IsolateThread thread, ObjectHandle networkHandl
...
-            @Override
-            public double getValue(int index) {
-                return valuePtr.read(index);
+    public static UpdatingDataframe createDataframe(PyPowsyblApiHeader.ArrayPointer<PyPowsyblApiHeader.SeriesPointer> dataframe, int columnsNumber, int elementCount) {
+        CUpdatingDataframe updatingDataframe = new CUpdatingDataframe(elementCount);
+        for (int i = 0; i < columnsNumber; i++) {
+            PyPowsyblApiHeader.SeriesPointer seriesPointer = dataframe.getPtr().addressOf(i);
+            String name = CTypeUtil.toString(seriesPointer.getName());
+            switch (seriesPointer.getType()) {
+                case 0:
```
sylvlecl: Let's reuse the constants in CDataframeHandler (should we move them?).

cpp/src/bindings.cpp
```diff
@@ -23,6 +22,51 @@ void bindArray(py::module_& m, const std::string& className) {
...
+        } else if (type == 2 || type == 3) {
+            std::vector<int> values = py::cast<std::vector<int>>(columnsValues[indice]);
+            column->data.ptr = pypowsybl::copyVectorInt(values);
+        }
+    }
+    dataframe->ptr = columns;
+    return dataframe;
+}
+
+void updateNetworkElementsWithSeries(pypowsybl::JavaHandle network, py::list columnsValues, const std::vector<std::string>& columnsNames, const std::vector<int>& columnsTypes, const std::vector<int>& indexBool, int columnSize, int columnsNumber, element_type elementType) {
```
sylvlecl: `columnSize` and `columnsNumber` seem redundant with the other parameters (we know the size of columnsNames for example) : let's remove them everywhere.

sylvlecl (changes requested):
More minor fixes and suggestions

java/src/main/java/com/powsybl/dataframe/AbstractDataframeMapper.java
EtienneLt: I put the boolean case in the integer case

case INT_SERIES_TYPE:
                case BOOLEAN_SERIES_TYPE:
                    updatingDataframe.addSeries(new IntSeries(name, elementCount,
                                    (CIntPointer) seriesPointer.data().getPtr()),
                            new SeriesMetadata(seriesPointer.isIndex(), name, false, SeriesDataType.INT));
                    break;

sylvlecl: Ok, added a throw for default, just in case.

sylvlecl (approved):

Issue #213, opened 2021-08-12
Sensitivity analysis: improve factor configuration


geofjamg:
* **Do you want to request a *feature* or report a *bug*?**
feature

* **What is the expected behavior?**
With this current Python API, factors defined by  set_branch_flow_factor_matrix are calculated for the pre-contingency state and for **ALL** post contingency states (defined by add_single_element_contingency).

```python
analysis = pp.sensitivity.create_dc_analysis()
analysis.add_single_element_contingency('NHV1_NHV2_1')
analysis.add_single_element_contingency('NHV1_NHV2_2')
analysis.set_branch_flow_factor_matrix(branches_ids=['NHV1_NHV2_1', 'NHV1_NHV2_2'], variables_ids=['LOAD'])
result = analysis.run(network)
```

`set_branch_flow_factor_matrix` become `add_branch_flow_factor_matrix` because we will be able to define several matrices per simulation.

In addition, we should be able to define factors that will be calculated only for pre-contingency state:
```python
analysis.add_precontingency_branch_flow_factor_matrix(branches_ids=['NHV1_NHV2_1', 'NHV1_NHV2_2'], variables_ids=['LOAD'])
```

And also factors that will be calculated only post contingency states and for a subset of contingencies:
```python
analysis.add_postcontingency_branch_flow_factor_matrix(branches_ids=['NHV1_NHV2_1', 'NHV1_NHV2_2'], variables_ids=['LOAD'], contingency_ids=['NHV1_NHV2_2'])
``` 

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #244, opened 2021-10-07
Sensitivity analysis: improve factor configuration 

obrix:

This PR is a proposed implementation for issue #213 

This a refactor of the sensitivity analysis API to handle multiple factor matrix for various contingency contexts. 

A single sensitivity matrix for branch flow we supported.

Multiple named sensitivity matrix are now supported with various contingency contexts.

**Other information**:


pypowsybl/sensitivity.py
```diff
@@ -249,11 +319,62 @@ def set_branch_flow_factor_matrix(self, branches_ids: _List[str], variables_ids:
                 branch_data_frame_index.append(TO_REMOVE)
             else:
                 raise _PyPowsyblError(f'Unsupported factor variable type {type(variable_id)}')
+        return (flatten_variables_ids, branch_data_frame_index)
 
-        _pypowsybl.set_branch_flow_factor_matrix(self._handle, branches_ids, flatten_variables_ids)
```
geofjamg, line 254: @obrix for backward compatibility, I think we should:
 - keep matrix_id optional so that if not specified we add factors to "default" matrix
 - keep set_branch_flow_factor_matrix that forward to add_.. with default matrix and deprecate it in the API

geofjamg (changes requested):

pypowsybl/sensitivity.py
obrix: I added the modifications.

cpp/src/bindings.cpp
```diff
@@ -475,6 +475,15 @@ PYBIND11_MODULE(_pypowsybl, m) {
     m.def("set_branch_flow_factor_matrix", &pypowsybl::setBranchFlowFactorMatrix, "Add a branch_flow factor matrix to a sensitivity analysis",
```
geofjamg: Not used anymore?

cpp/src/bindings.cpp
```diff
@@ -492,13 +501,19 @@ PYBIND11_MODULE(_pypowsybl, m) {
             });
 
     m.def("get_branch_flows_sensitivity_matrix", &pypowsybl::getBranchFlowsSensitivityMatrix, "Get sensitivity analysis result matrix for a given contingency",
-          py::arg("sensitivity_analysis_result_context"), py::arg("contingency_id"));
+              py::arg("sensitivity_analysis_result_context"), py::arg("matrix_id"), py::arg("contingency_id"));
+
+    m.def("get_precontingency_branch_flows_sensitivity_matrix", &pypowsybl::getPreContingencyBranchFlowsSensitivityMatrix, "Get sensitivity analysis result matrix for a given contingency",
```
geofjamg: Why do we need 3 different methods get_branch_flows_sensitivity_matrix, get_precontingency_branch_flows_sensitivity_matrix, get_postcontingency_branch_flows_sensitivity_matrix ?
Matrix are uniquely identified by its ID, so no need to specify the contingency context.

docs/reference/sensitivity.rst
```diff
@@ -46,7 +46,7 @@ In AC mode, you can define voltage sensitivities, in addition to flows sensitivi
    :nosignatures:
    :toctree: api/
 
-    SensitivityAnalysis.set_branch_flow_factor_matrix
+    SensitivityAnalysis.add_branch_flow_factor_matrix
```
geofjamg, line 49: There is other methods to add?

java/src/main/java/com/powsybl/python/SensitivityAnalysisContext.java
```diff
@@ -28,26 +29,107 @@
...
+        int getColumnCount() {
+            return columnIds.size();
+        }
+    }
+
+    private final Map<String, MatrixInfo> branchFlowFactorsMatrix = new HashMap<>();
+
+    private MatrixInfo busVoltageFactorsMatrix;
+
+    private final Map<String, MatrixInfo> preContingencyBranchFlowFactorMatrix = new HashMap<>();
```
geofjamg: Why do we need 3 different maps branchFlowFactorsMatrix, preContingencyBranchFlowFactorMatrix, postContingencyBranchFlowFactorMatrix ?

java/src/main/java/com/powsybl/python/SensitivityAnalysisContext.java
```diff
@@ -28,26 +29,107 @@
...
 
     void setBranchFlowFactorMatrix(List<String> branchesIds, List<String> variablesIds) {
-        this.branchFlowBranchesIds = Objects.requireNonNull(branchesIds);
-        this.branchFlowVariablesIds = Objects.requireNonNull(variablesIds);
+        addBranchFlowFactorMatrix("default", branchesIds, variablesIds);
+    }
+
+    void addBranchFlowFactorMatrix(String matrixId, List<String> branchesIds, List<String> variablesIds) {
+        MatrixInfo info = new MatrixInfo(ContingencyContextType.ALL, SensitivityFunctionType.BRANCH_ACTIVE_POWER, branchesIds, variablesIds, Collections.emptyList());
+        branchFlowFactorsMatrix.put(matrixId, info);
```
geofjamg: An ID collision check is missing

java/src/main/java/com/powsybl/python/SensitivityAnalysisContext.java
```diff
@@ -64,84 +146,162 @@ void setBusVoltageFactorMatrix(List<String> busVoltageIds, List<String> targetVo
         return injection;
     }
 
+    List<MatrixInfo> prepareMatrices() {
+        List<MatrixInfo> matrices = new ArrayList<>();
+        int offsetData = 0;
+        int offsetColumns = 0;
+
+        for (MatrixInfo matrix : branchFlowFactorsMatrix.values()) {
```
geofjamg, line 152: This code would be simpler if just one map.

java/src/main/java/com/powsybl/python/SensitivityAnalysisContext.java
```diff
@@ -64,84 +146,162 @@ void setBusVoltageFactorMatrix(List<String> busVoltageIds, List<String> targetVo
...
+
+        for (MatrixInfo matrix : postContingencyBranchFlowFactorMatrix.values()) {
+            matrix.setOffsetData(offsetData);
+            matrix.setOffsetColumn(offsetColumns);
+            matrices.add(matrix);
+            offsetData += matrix.getColumnCount() * matrix.getRowCount();
+            offsetColumns += matrix.getColumnCount();
+        }
+
+        if (busVoltageFactorsMatrix != null) {
```
geofjamg, line 160: We could have support same kind of feature for voltage sensitivity. Later?

java/src/main/java/com/powsybl/python/SensitivityAnalysisContext.java
```diff
@@ -64,84 +146,162 @@ void setBusVoltageFactorMatrix(List<String> busVoltageIds, List<String> targetVo
...
+        double[][] referencesByContingencyIndex = new double[contingencies.size()][totalColumnsCount];
+
+        NavigableMap<Integer, MatrixInfo> factorIndexMatrixMap = new TreeMap<>();
+        for (MatrixInfo m : matrices) {
+            factorIndexMatrixMap.put(m.getOffsetData(), m);
+        }
+
+        SensitivityValueWriter valueWriter = (factorContext, contingencyIndex, value, functionReference) -> {
+            int factorIndex = factorContext;
+            MatrixInfo m = factorIndexMatrixMap.floorEntry(factorIndex).getValue();
```
geofjamg, line 278: This is the most critical part for performance, because done for each of the huge number of factor. Maybe ok because small numer of matrices?

java/src/main/java/com/powsybl/python/SensitivityAnalysisResultContext.java
```diff
@@ -19,13 +19,13 @@
  */
 public class SensitivityAnalysisResultContext {
 
-    private final int flowRowCount;
+    Map<String, SensitivityAnalysisContext.MatrixInfo> branchFlowFactorsMatrix;
```
geofjamg: private final

java/src/main/java/com/powsybl/python/SensitivityAnalysisResultContext.java
```diff
@@ -56,20 +59,42 @@ public SensitivityAnalysisResultContext(int flowRowCount, int flowColCount, int
...
+        return createDoubleMatrix(() -> getValues(""), m.getOffsetData(), m.getRowCount(), m.getColumnCount());
+    }
+
+    public PyPowsyblApiHeader.MatrixPointer createPostContingencyBranchFlowsSensitivityMatrix(String matrixId, String contingency) {
+        SensitivityAnalysisContext.MatrixInfo m = this.postContingencyFactorMatrix.get(matrixId);
+        return createDoubleMatrix(() -> getValues(contingency), m.getOffsetData(), m.getRowCount(), m.getColumnCount());
+    }
+
+    public PyPowsyblApiHeader.MatrixPointer createBranchFlowsSensitivityMatrix(final String contingencyId) {
+        return createBranchFlowsSensitivityMatrix("default", contingencyId);
```
geofjamg: Maybe the concept of "default" should only be on python side?

pypowsybl/sensitivity.py
```diff
@@ -99,45 +99,121 @@ class DcSensitivityAnalysisResult:
...
         self._handle = result_context_ptr
         self.result_context_ptr = result_context_ptr
         self.branches_ids = branches_ids
         self.branch_data_frame_index = branch_data_frame_index
+        self.precontingency_branches_ids = precontingency_branches_ids
+        self.precontingency_branch_data_frame_index = precontingency_branch_data_frame_index
+        self.postcontingency_branches_ids = postcontingency_branches_ids
+        self.postcontingency_branch_data_frame_index = postcontingency_branch_data_frame_index
+
+    def post_process_branch_flows(self, flows: _pd.DataFrame, branch_ids: _List[str], branch_data_frame_index: _List[str]) -> _Optional[_pd.DataFrame]:
```
geofjamg: Not used

pypowsybl/sensitivity.py
```diff
@@ -99,45 +99,121 @@ class DcSensitivityAnalysisResult:
...
+    def get_postcontingency_branch_flows_sensitivity_matrix(self, matrix_id: str = 'default', contingency_id: str = None) -> _Optional[_pd.DataFrame]:
+        """ Get the matrix of branch flows sensitivities on the post contingency.
+        Args:
+            matrix_id (str, mandatory): ID of the matrix
+            contingency_id (str, optional): ID of the contingency
+        Returns:
+            the matrix of branch flows sensitivities
+        """
+        m = _pypowsybl.get_postcontingency_branch_flows_sensitivity_matrix(self.result_context_ptr, matrix_id, '' if contingency_id is None else contingency_id)
+        if m is None:
```
geofjamg: Lot's of duplicated code

geofjamg (changes requested):

java/src/main/java/com/powsybl/python/SensitivityAnalysisContext.java
obrix: I think usage should probably a few big matrices rather than lots of small matrices, in this cas this has minimal impact on performance. No way to enforce what usage will be though.

java/src/main/java/com/powsybl/python/SensitivityAnalysisResultContext.java
obrix: In fact this dates back to the first implementation last year. It should be cleaned.

java/src/main/java/com/powsybl/python/SensitivityAnalysisContext.java
obrix: Matrices are put in a unique container in prepareMatrices method. I cannot see any reason for the separated maps. There may have been one earlier but I can't remember. This will be cleaned.

geofjamg (approved):

Issue #238, opened 2021-10-01
Update methods simpler overloads


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the current behavior?**

Update methods only accept `pandas.DataFrame` as input, which can be a little cumbersome 
for simple use cases, for example when you only want to update one value. For example,
a minimalist use case will look like :  
```python
network.update_generators(pd.DataFrame(index=['GEN'], data={'target_p': [ 350 ] }))
```

* **What is the expected behavior?**

We could provide simpler overloads which would not take a dataframe as input, but a list of named arguments:
```python
network.update_generators(id='GEN', target_p=350)
network.update_generators(id=['GEN1', 'GEN2'], target_p=[350, 400])
network.update_generators(id=['GEN1', 'GEN2'], target_p=some_numpy_array)
```

* **What is the motivation / use case for changing the behavior?**

Ease of use, in particular for simple use cases.

sylvlecl:
Note:
to implement this, we should use keyword arguments, and transform keywords arguments in dataframes if no dataframe is provided as the first argument.

Like:
```python
    def update_elements(self, element_type: _pypowsybl.ElementType, df: _DataFrame = None, **kwargs):
        if df is None:
            ... transform kwargs in dataframe

```

sylvlecl:
Implemented in #251

PR #251, opened 2021-10-13
Simplier update methods

EtienneLt:

#238


pypowsybl/network.py
```diff
@@ -1062,7 +1062,7 @@ def get_reactive_capability_curve_points(self) -> _DataFrame:
         """
         return self.get_elements(_pypowsybl.ElementType.REACTIVE_CAPABILITY_CURVE_POINT)
 
-    def update_elements(self, element_type: _pypowsybl.ElementType, df: _DataFrame):
+    def update_elements(self, element_type: _pypowsybl.ElementType, df: _DataFrame = None, **kwargs):
```
sylvlecl, line 1066: We need tests !! :)

pypowsybl/network.py
```diff
@@ -1078,31 +1078,46 @@ def update_elements(self, element_type: _pypowsybl.ElementType, df: _DataFrame):
...
-            columns_types.append(series_type)
-            columns_values.append(series.values)
-            index_bool.append(0)
+        line_to_update = 0
+        if df is None:
+            for key, value in kwargs.items():
+                columns_names.append(key)
+                index_bool.append(int(_pypowsybl.is_index(element_type, key)))
+                columns_types.append(_pypowsybl.get_series_type(element_type, key))
+                if type(value) == list:
```
sylvlecl: It looks like the best way now to detect if it's a sequence (more generic than list), is to:
```python
if isinstance(value, collections.Sequence):
```
If it's already one, I think we don't need to wrap it in a numpy array (it just adds an intermediate):
if it's already a numpy array very good, else it's not worth to copy it and pybind should still be able to transform it to a std::vector

Issue #102, opened 2021-05-17
Plugin system for network data tables


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the current behavior?**

The data series available for each type of equipment are "hard coded" in pypowsybl.

* **What is the expected behavior?**

In order to make available third party IIDM extensions data from python, we could add a plugin system so that extensions provider can define new columns to be appended to equipments dataframes.

Extensions providerq will need to implement a service for that, a minimal spec would be:
```java
interface NetworkExtensionSeriesProvider {
    String getExtensionName();
    DataframeElementType getElementType();
    void addSeries(NetworkDataframeMapperBuilder builder);
}
```

At the time of mappers creation, we will look for all the series provider for the current equipment type through the `ServiceLoader`.

**Note:**
The order of columns should be stable between executions, we could ensure that by sorting providers by their extension name.

* **What is the motivation / use case for changing the behavior?**

Make IIDM extensions data available to python users.

sylvlecl:
Good first candidates to benefit from that feature would be the following extensions:
- `GeneratorActivePowerControl`
- `HvdcAngleDroopActivePowerControl`
- `HvdcOperatorActivePowerRange`

Note that, since those classes are in powsybl-core, we could implement this without any plugin mechanism.
However, it makes sense to treat all extensions equals, and that's a good opportunity to test that mechanism.

sylvlecl:
@CBiasuzzi raised the issue of "what to provide when extensions are absent".

For example, if generator GEN1 has a GeneratorActivePowerControl but GEN2 does not have it, what should be the content of columns for GEN2 ?

For `float` and `string` it's kind of OK: we have `nan` and empty string.
But for `bool` and `int` we don't have that kind of special values. Of course this is a broader issue than just for extensions, as already mentioned in issue #79 .

**An overview of missing values support in numpy and pandas:**
Basically, in both libraries, there is some support of missing values through boolean masks.
- numpy: [masked arrays](https://numpy.org/doc/stable/reference/maskedarray.html)
- pandas: for example [IntegerArray](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.arrays.IntegerArray.html#pandas.arrays.IntegerArray) for integers, which is implemented as a masked array too. This goes together with the special value `pd.NA`. However, all this is mentioned as experimental in the documentation.
[...]

PR #277, opened 2021-12-01
Network extension series (Plugin system for network data tables enhancement)

CBiasuzzi:

#102

**Other information**:


pypowsybl/network.py
```diff
@@ -3018,6 +3027,14 @@ def create_eurostag_tutorial_example1_with_power_limits_network() -> Network:
     """
     return _create_network('eurostag_tutorial_example1_with_power_limits')
 
+def create_eurostag_tutorial_example1_with_apc_extension() -> Network:
```
sylvlecl: Probably better to hide more this method, for example in the unit test file, since it's really more targetted at unit testing than as a case for users.

sylvlecl (approved):
It look good enough to be merged at least as a first support, what do you think @CBiasuzzi  ?

Do you intend to change something more?

If not, we only need to add the new methods in the API reference in documentation.

geofjamg (approved):

pypowsybl/network.py
CBiasuzzi: I've removed that method from the APIs: the test now calls the generic _create_network func, directly.
I also added references to the two methods get_extension and get_extensions_names in the "Network element access" and  "Miscellaneous network functions" sections of the Network API reference; 
Do you think they deserve a dedicated section, instead (e.g. "Network elements extensions access") ?
That being said, I'm fine with its current status.

sylvlecl: It looks good this way, thanks :)

Issue #142, opened 2021-06-22
Integration with python logging API


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**

Java logging configuration is "hardcoded" in a logback config file, the only thing the user can do is setting the log level to `DEBUG` by calling the method `pypowsybl.set_debug_mode`.

* **What is the expected behavior?**

We could try to integrate java logs to python logging framework.
In order to achieve this, a solution would be to implement SLF4J API so that loggers actually send logs to python loggers.

_**As of now, it's not clear if it's worth the effort**_

* **What is the motivation / use case for changing the behavior?**

Giving the hand to python user on powsybl logs, using its usual logging API.

geofjamg:
Design proposition:
 - On Java side, use logback-classic to store logs with a new appender:
Example:
```java
import ch.qos.logback.classic.spi.ILoggingEvent;
import ch.qos.logback.core.AppenderBase;
import org.slf4j.LoggerFactory;
import ch.qos.logback.classic.Logger;

import java.util.ArrayList;
import java.util.Arrays;
import java.util.List;

public class LoggerTest {

    private static final Logger LOGGER = (Logger) LoggerFactory.getLogger(LoggerTest.class);

    public static class ListAppender<E> extends AppenderBase<E> {

        private final List<E> list = new ArrayList<>();

        protected synchronized void append(E e) {
            list.add(e);
        }

        public synchronized List<E> getListAndClear() {
            List<E> listCopy = new ArrayList<>(list);
            list.clear();
            return listCopy;
        }
    }

    public static void main(String[] args) {
        // create and start a ListAppender
        ListAppender<ILoggingEvent> listAppender = new ListAppender<>();
        listAppender.start();

        // add the appender to the logger
        LOGGER.addAppender(listAppender);

        LOGGER.info("toto {}", 1);
        LOGGER.info("tutu {}", "sss");

        // JUnit assertions
        List<ILoggingEvent> logsList = listAppender.getListAndClear();
        System.out.println(logsList.get(0).getMessage());
        System.out.println(logsList.get(0).getFormattedMessage());
        System.out.println(logsList.get(0).getLevel());
        System.out.println(Arrays.toString(logsList.get(0).getArgumentArray()));
        System.out.println(logsList.get(0).getLoggerName());
        System.out.println(logsList.get(0).getTimeStamp());
   }
}
```
[...]

PR #280, opened 2021-12-17
Integration with python logging implementation

obrix:
- [X ] The commit message follows our guidelines

Yes this is an implementation for https://github.com/powsybl/pypowsybl/issues/142

This integrate java logging with logback with python logging from pypowsybl

**Other information**:


java/src/main/java/com/powsybl/python/CustomAppender.java
```diff
@@ -0,0 +1,48 @@
+package com.powsybl.python;
```
geofjamg, line 7: Licence

java/src/main/java/com/powsybl/python/CustomAppender.java
```diff
@@ -0,0 +1,48 @@
...
+
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.AppenderBase;
+import ch.qos.logback.core.encoder.Encoder;
+
+import java.nio.charset.StandardCharsets;
+
+public class CustomAppender extends AppenderBase<ILoggingEvent> {
```
geofjamg, line 18: Author

java/src/main/java/com/powsybl/python/CustomAppender.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private Encoder<ILoggingEvent> encoder;
+
+    @Override
+    protected void append(final ILoggingEvent e) {
+        PyPowsyblApiLib.Callback logMessage = (PyPowsyblApiLib.Callback) PyPowsyblApiLib.loggerCallback;
+        logMessage.invoke(logbackLevelToPythonLevel(e.getLevel()), CTypeUtil.toCharPtr(new String(this.getEncoder().encode(e), StandardCharsets.UTF_8)));
+    }
+
+    private int logbackLevelToPythonLevel(Level l) {
```
geofjamg: You should move this along with the invert method from PyPowsyblLib into a dedicated class

java/src/main/java/com/powsybl/python/CustomAppender.java
```diff
@@ -0,0 +1,48 @@
...
+import java.nio.charset.StandardCharsets;
+
+public class CustomAppender extends AppenderBase<ILoggingEvent> {
+
+    private Encoder<ILoggingEvent> encoder;
+
+    @Override
+    protected void append(final ILoggingEvent e) {
+        PyPowsyblApiLib.Callback logMessage = (PyPowsyblApiLib.Callback) PyPowsyblApiLib.loggerCallback;
+        logMessage.invoke(logbackLevelToPythonLevel(e.getLevel()), CTypeUtil.toCharPtr(new String(this.getEncoder().encode(e), StandardCharsets.UTF_8)));
```
geofjamg: This is not correct, the full log is serialized into a string, we should send to python all the log detail
 - date
 - message
 - logger

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -58,6 +60,10 @@
 @CContext(Directives.class)
 public final class PyPowsyblApiLib {
 
+    static boolean readConfig = true;
```
geofjamg: ?

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -687,4 +693,43 @@ public static void addMonitoredElements(IsolateThread thread, ObjectHandle secur
     public static void freeString(IsolateThread thread, CCharPointer string, ExceptionHandlerPointer exceptionHandlerPtr) {
         doCatch(exceptionHandlerPtr, () -> UnmanagedMemory.free(string));
     }
+
+    interface Callback extends CFunctionPointer {
+        @InvokeCFunctionPointer
+        void invoke(int level, CCharPointer string);
```
geofjamg: Should more detailed to really dispatch to python by providing all de individual fields

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -687,4 +693,43 @@ public static void addMonitoredElements(IsolateThread thread, ObjectHandle secur
...
+
+    @CEntryPoint(name = "setLogLevel")
+    public static void setLogLevel(IsolateThread thread, int logLevel, ExceptionHandlerPointer exceptionHandlerPtr) {
+        doCatch(exceptionHandlerPtr, () -> {
+            Logger rootLogger = (Logger) LoggerFactory.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME);
+            rootLogger.setLevel(pythonLevelToLogbackLevel(logLevel));
+        });
+    }
+
+    private static Level pythonLevelToLogbackLevel(int l) {
```
geofjamg: See previous comment

pypowsybl/__init__.py
```diff
@@ -31,6 +33,15 @@
     "sensitivity"
 ]
 
+#Setup a default logger that is the root logger with default log level set to ERROR
+_pypowsybl.set_logger(logging.getLogger())
+logging.getLogger().setLevel(level=logging.ERROR)
+
+def set_logger(logger: Logger) -> None:
```
geofjamg: I don't think we should allow and bother the user to set the logger
As a first version of logging
  - log everything from Java into a "powsybl" logger (logging.getLogger('powsybl'). the problem with this approach is that we lost all Java logger infos (= java packages)
  - for a future dev, we should be able to create a python logger automatically for each Java logger

pypowsybl/__init__.py
```diff
@@ -31,6 +33,15 @@
...
+    _pypowsybl.set_logger(logger)
+
+def get_logger() -> Logger:
+    return _pypowsybl.get_logger()
 
 def set_debug_mode(debug: bool = True) -> None:
```
geofjamg: to remove

tests/test_logging.py
```diff
@@ -0,0 +1,54 @@
+#
+# Copyright (c) 2020, RTE (http://www.rte-france.com)
```
geofjamg: 2022

tests/test_logging.py
```diff
@@ -0,0 +1,54 @@
...
+
+class LoggingTestCase(unittest.TestCase):
+
+    def setUp(self):
+        pp.set_config_read(False)
+
+    def test_loglevel_lf(self):
+
+        #Build a PyPow logger to handle logs and install it
+        pypowLogger = logging.getLogger("PyPow")
```
geofjamg: Should be done automatically

cpp/src/pypowsybl.h
```diff
@@ -310,6 +313,42 @@ std::vector<std::vector<SeriesMetadata>> getNetworkElementCreationDataframesMeta
...
+    py::object getLogger() {
+      return this->mlogger;
+    }
+
+    bool loggerInitialized() {
+      return this->initialized;
+    }
+
+  private:
+    static CppToPythonLogger* mSingleton;
```
geofjamg: we prefer _xxx convention

cpp/src/pypowsybl.h
```diff
@@ -13,6 +13,9 @@
 #include <memory>
 #include <stdexcept>
 #include "pypowsybl-api.h"
+#include <pybind11/pybind11.h>
+
+namespace py = pybind11;
```
geofjamg: Until now we tried to have a pypowsybl namespace free of pybind11 code and all python binding is in bindings.cpp

cpp/src/pypowsybl.cpp
```diff
@@ -689,4 +697,33 @@ void createElement(pypowsybl::JavaHandle network, dataframe_array* dataframes, e
     pypowsybl::callJava<>(::createElement, network, elementType, dataframes);
 }
 
+void CppToPythonLogger::logFromJava(int level, char* pArg) {
+
+    py::gil_scoped_acquire acquire;
+    if(CppToPythonLogger::get()->loggerInitialized())
+    {
```
geofjamg: On C++, could you try to use same style as the existing code?

cpp/src/pypowsybl.cpp
```diff
@@ -689,4 +697,33 @@ void createElement(pypowsybl::JavaHandle network, dataframe_array* dataframes, e
     pypowsybl::callJava<>(::createElement, network, elementType, dataframes);
 }
 
+void CppToPythonLogger::logFromJava(int level, char* pArg) {
```
geofjamg: Is it the right place for this method? A free function?

cpp/src/pypowsybl.cpp
```diff
@@ -60,6 +60,14 @@ template<typename F, typename... ARGS>
 void callJava(F f, ARGS... args) {
     GraalVmGuard guard;
     exception_handler exc;
+
+    //Explicitly update log level on java side
+    if( CppToPythonLogger::get()->loggerInitialized() ) {
```
geofjamg: There is no sync between "loggerInitialized" and "get". Could cause a crash if no luck?

geofjamg (changes requested):

java/src/main/java/com/powsybl/python/CustomAppender.java
obrix: You lose the formatting from logback encoder then. But can still format it from python. Is this what you have in mind ? Specifiying a custom formatter from python ?

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
obrix: This was present before on the main branch but is no longer used. Probably an error from the multiple merging and rebasing...

cpp/src/pypowsybl.h
obrix: Let's move everything logging related in a pylogging.h / pylogging.cpp.

pypowsybl/__init__.py
obrix: Done.

geofjamg (approved):

Issue #285, opened 2022-01-13
[Perfs] Release the GIL when calling long java processing, to enable multi-threading


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the current behavior?**

Python GIL is never released when performing long lasting processings, which does not allow to carry out multi-threaded programs.

For example, the following code actually executes the four reads sequentially:
```python
import pypowsybl.network as pn
from multiprocessing.pool import ThreadPool

def read_network(num):
    print('Reading num {}'.format(num))
    return pn.load('/path/to/network.xiidm')

with ThreadPool(4) as pool:
    pool.map(read_network, range(4))
```

* **What is the expected behavior?**

We should be able to import files or run loadflow in parallel, for example.

`pybind11` provides `py::gil_scoped_release release;` for that purpose.

* **What is the motivation / use case for changing the behavior?**

Being able to use multi-threading to scale processing.

sylvlecl:
Closing but we need to keep this in mind for future methods

PR #287, opened 2022-01-13
Release GIL for expansive operations

sylvlecl:

#285


sylvlecl:
Interesting stuff @geofjamg :

With that evolution, we can indeed execute multithreaded code such as this one (4 imports in //):
```python
import pypowsybl.network as pn
from multiprocessing.pool import ThreadPool

def read_network(num):
    print('Reading num {}'.format(num))
    return pn.load('/path/to/network.xiidm')

with ThreadPool(4) as pool:
    pool.map(read_network, range(4))
```
**BUT**, it does not scale that well, from 40s with 1 thread to 25 with 2 threads and no more gain for 3/4 threads.

However, I tried to build with G1 garbage collector (graalvm EE), and it gives very different results.
It scales very well, going to 10s with 4 threads as could be expected.

Issue #271, opened 2021-11-10
Dataframes elements selection


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the current behavior?**

We can only retrieve the whole tables of network elements, not a subset (for example only for 1 generator among the 100 generators of a network).
This can hurt performances for some use cases.

* **What is the expected behavior?**

Allowing to define for which elements we want to retrieve data. In order to do this, the user will need to provide values for all the index columns. For most types of elements, this will only be a list of ids, but this could also be a multi-index:
```python
gens = network.get_generators(id=['G1', 'G2'])
steps = network.get_ratio_tap_changer_step(index=pd.MultiIndex(...))
# or using 2 lists
steps = network.get_ratio_tap_changer_step(id=['T1', 'T2'], position=[1, 1])
```

* **What is the motivation / use case for changing the behavior?**

Mainly performance improvements when using only a small part of network elements, for example when getting the position of on switch in the whole network.

PR #311, opened 2022-02-08
Dataframes elements selection

CBiasuzzi:

#271

**Other information**:


pypowsybl/network.py
```diff
@@ -390,16 +391,41 @@ def get_elements(self, element_type: _pypowsybl.ElementType, all_attributes: boo
...
+                if is_multi_index:
+                    columns_values.append(sel_index.get_level_values(index_name))
+                else:
+                    columns_values.append(sel_index.values)
+                columns_names.append(index_name)
+                columns_types.append(metadata_by_name[index_name].type)
+                is_index.append(False)
+            sel_index_array = _pypowsybl.create_dataframe(columns_values, columns_names, columns_types, is_index)
+        else:
+            sel_index_array = _pypowsybl.create_dataframe([[]], [''], [1], [False])
```
sylvlecl: Is it possible to simply use a `None` here ?
I guess it should translate to null on C and java sides.

pypowsybl/network.py
```diff
@@ -369,14 +369,15 @@ def get_elements_ids(self, element_type: _pypowsybl.ElementType, nominal_voltage
                                                    main_connected_component, main_synchronous_component,
                                                    not_connected_to_same_bus_at_both_sides)
 
-    def get_elements(self, element_type: _pypowsybl.ElementType, all_attributes: bool = False, attributes: _List[str] = []) -> _DataFrame:
+    def get_elements(self, element_type: _pypowsybl.ElementType, all_attributes: bool = False, attributes: _List[str] = [], sel_index: _pd.Index=None) -> _DataFrame:
```
sylvlecl: I'm not a big fan of the "sel_index" name :), because I am not sure what it stands for. But is there a better alternative ?

Maybe we could name it:
 - `elements`? meaning "those are the elements we want to get". It's not great for other methods like `get_generators(elements=***)`, but if we name it after the element type, like `get_generators(generators=***)` ? it will get too long for some types, so I would keep "elements".
 - `filter` ? meaning I want to filter only those elements. But maybe it's too vague.

In the end I think "elements" is OK, this way "attribute" word is clearly linked to columns while "element" word is clearly linked to rows: tell me what you think about it!

sylvlecl: About the type `pd.index` : it would be great to allow any array-like input at least for simple index (not multi index).

pypowsybl/network.py
```diff
@@ -390,16 +391,41 @@ def get_elements(self, element_type: _pypowsybl.ElementType, all_attributes: boo
         elif len(attributes) > 0:
             filter_attributes = _pypowsybl.FilterAttributesType.SELECTION_ATTRIBUTES
 
-        series_array = _pypowsybl.create_network_elements_series_array(self._handle, element_type, filter_attributes, attributes)
+        if sel_index is not None:
```
sylvlecl: is it possible to share that code with the one used in "update" ?

java/src/main/java/com/powsybl/dataframe/network/AbstractNetworkDataframeMapper.java
```diff
@@ -38,6 +43,35 @@ public void createDataframe(Network network, DataframeHandler dataframeHandler,
         mappers.stream().forEach(mapper -> mapper.createSeries(items, dataframeHandler));
     }
 
+    protected List<T> getFilteredItems(Network network, DataframeFilter dataframeFilter) {
+        if (dataframeFilter.getRowsIds().size() == 0) {
```
sylvlecl: If possible it would be better to differentiate null (no filter) from 0, maybe in some use cases the user could be surprised if he requests a 0 size array (maybe inside some business logic) and gets all the network.

java/src/main/java/com/powsybl/dataframe/network/AbstractNetworkDataframeMapper.java
```diff
@@ -38,6 +43,35 @@ public void createDataframe(Network network, DataframeHandler dataframeHandler,
...
+            return getItems(network);
+        }
+        return getItems(network).stream().filter(item -> filterItem(item, dataframeFilter)).collect(Collectors.toList());
```
sylvlecl: I wonder what would be the perf impact on doing a search by IDs instead of a filter, for small selections:
do you think you could make a small test for this?

Like trying to get 1 switch by its ID with this approach on one side, and with a "network.getSwitch(id)" approach on the other side, and time both approaches, to check if it can be worth it to do something "smart" about it in the future or not ? (not necessarily in that PR)

sylvlecl: (note: timing it from the python side, not from a java test for ex., to have "real" timing)

pypowsybl/network.py
CBiasuzzi: changed to use None; it can actually be checked on Java side using the ArrayPointer isNonNull method

java/src/main/java/com/powsybl/dataframe/network/AbstractNetworkDataframeMapper.java
CBiasuzzi: differentiated null filter  from empty filter: now an empty filter (a 0 size array) returns no elements

pypowsybl/network.py
CBiasuzzi: you re right: 'elements' sounds better; also changed the single-index get_ cases to take a List parameter instead of a pd.Index

sylvlecl: Ok!
About the type, sorry to ask for an additional change: the best option would be to accept a broader range of inputs.

For example, in updates, we accept `ArrayLike` (more or less any primitive object, list, or numpy array) and convert them to a numpy array (see `_to_array`). I don't know if pandas indexes are array-like, if not we could even accept a `Union[ArrayLike, pd.Index]`.

sylvlecl: This way we can choose between multiple options:
```python
network.get_generators(elements='GEN')  # for simple use cases
network.get_generators(elements=['GEN1', 'GEN2'])
network.get_generators(elements=other_dataframe.index)  # For ex. using the index of filtered dataframe
```

CBiasuzzi: revised to accept ArrayLike parameters

pypowsybl/network.py
CBiasuzzi: I've moved the common code to a shared function

java/src/main/java/com/powsybl/dataframe/network/AbstractNetworkDataframeMapper.java
CBiasuzzi: searching by id (a single switch id) is roughly 30 times faster than applying a filter
(360 times faster than a pandas filter applied on the whole elements set, python-side). 
The latest commit contains the search-by-ids  implementation.

sylvlecl (approved):

Issue #301, opened 2022-02-02
Use pandas.Categorical for enum types


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**

Enum attributes are represented as strings in the provided dataframes, which is not memory efficient and not very descriptive.

* **What is the expected behavior?**

pandas proposes the `pd.Categorical` type for objects with only a few valid values.

* **What is the motivation / use case for changing the behavior?**

Improved memory efficiency and semantics.

About memory efficiency, see on a large network:
```python
>>> network.get_generators().memory_usage(deep=True)
...
energy_source           316067
...

>>> network.get_generators().astype('category').memory_usage(deep=True)
...
energy_source             5652
```

We can also improve the java -> python performances by transferring "codes" (integers) instead of names, and create the series from the underlying codes:
```python
series = pd.Series(pd.Categorical.from_codes(data, dtype=EnergySource))
```

sylvlecl:
Note that this will probably require to add some information about the enum type in the series metadata, in order to know which categorical type to use on python side.

PR #325, opened 2022-02-22
[WIP] add categorical for enum columns

EtienneLt:

#301


java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
```diff
@@ -231,7 +231,7 @@ static NetworkDataframeMapper loads() {
         return NetworkDataframeMapperBuilder.ofStream(Network::getLoadStream, getOrThrow(Network::getLoad, "Load"))
                 .stringsIndex("id", Load::getId)
                 .strings("name", l -> l.getOptionalName().orElse(""))
-                .enums("type", LoadType.class, Load::getLoadType)
+                .enums("type", LoadType.class, l -> l.getLoadType().ordinal())
```
sylvlecl, line 234: Impl detail: maybe we could move the `ordinal()` call inside the function

java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
```diff
@@ -552,7 +552,13 @@ private static NetworkDataframeMapper substations() {
                 .strings("name", s -> s.getOptionalName().orElse(""))
                 .strings("TSO", Substation::getTso, Substation::setTso)
                 .strings("geo_tags", substation -> String.join(",", substation.getGeographicalTags()))
-                .enums("country", Country.class, s -> s.getCountry().orElse(null), Substation::setCountry)
+                .enums("country", Country.class, s -> {
+                    if (s.getCountry().isPresent()) {
```
sylvlecl: syntax: let's use `orElse`

pypowsybl/utils/dataframes.py
```diff
@@ -95,8 +96,12 @@ def _create_c_dataframe(df: DataFrame, series_metadata: List[_pp.SeriesMetadata]
         if series_name not in metadata_by_name:
             raise ValueError(f'No column named {series_name}')
         series = df[series_name]
-        series_type = metadata_by_name[series_name].type
+        metadata = metadata_by_name[series_name]
+        series_type = metadata.type
         columns_types.append(series_type)
-        columns_values.append(series.values)
+        if series_type == 4:
+            columns_values.append(pd.Categorical(series, categories=_pp.get_enum_values(metadata.enum_class)).codes)
```
sylvlecl, line 103: We could have a global map of those `CategoricalDType` to avoid to recreate them each time. It would be initialized on import.

pypowsybl/util.py
```diff
@@ -67,3 +76,7 @@ def create_data_frame_from_series_array(series_array: _pypowsybl.SeriesArray) ->
     else:
         index = _pd.MultiIndex.from_arrays(index_data, names=index_names)
     return _pd.DataFrame(series_dict, index=index)
+
+
+def get_enum_values(enum_class: str) -> List[str]:
```
sylvlecl, line 81: We could have a `get_categorical_dtype()` instead which would return the categorical dtype.
Maybe we could also have a `get_categorical_dtypes()` wjhich return the list of all supported dtypes.

java/src/test/java/com/powsybl/dataframe/DataframeMapperBuilderTest.java
```diff
@@ -24,20 +25,15 @@
  */
 class DataframeMapperBuilderTest {
 
-    private static enum Color {
-        RED,
```
sylvlecl, line 28: Actually, probably in the future we should enable to extend the list of supported enums via the `ServiceLoader` mechanism, in order to support this in extensions.

sylvlecl (commented):
A set of comments for minor implementation improvements.

We need to be careful about what could be a breaking change: 
for example empty countries will now be `np.nan`s, which could break existing code.

We will also probably need to update the elements creation part.

PR #330, opened 2022-03-08
Overriding default providers

EtienneLt:

Allow to define which provider should be used by default, for loadflow, security analysis, and sensitivity analysis.

OpenLoadFlow is the default loadflow provider and if it needs to be changed it has to be each time a loadflow is called 

the default loadflow provider can now be set.
each time a loadflow or a analysis is launched a log with the name of the loadflow provider is sent.


java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -220,6 +232,11 @@ private static LoadFlowParametersPointer convertToLoadFlowParametersPointer(Load
             Network network = ObjectHandles.getGlobal().get(networkHandle);
             LoadFlowParameters parameters = createLoadFlowParameters(dc, loadFlowParametersPtr);
             String providerStr = CTypeUtil.toString(provider);
+            if (providerStr.equals("")) {
+                providerStr = PyPowsyblConfiguration.getLoadFlowProvider();
+            }
+            Logger rootLogger = (Logger) LoggerFactory.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME);
```
sylvlecl: We should not log it in the root logger but in the class logger (arg to be replaced with PyPowsyblApiLib.class), same for other logs.

pypowsybl/__init__.py
```diff
@@ -56,3 +56,9 @@ def is_config_read() -> bool:
 
 def print_version() -> None:
     print(_pypowsybl.get_version_table())
+
+def set_loadflow_provider(provider: str) -> None:
```
sylvlecl: I think we should be more specific by naming method `set_default_` and `get_default_...`.

Also it deserves some docstrings, and to be added in the API reference in the doc :) (loadflow.rst in reference directory)

pypowsybl/loadflow.py
```diff
@@ -187,7 +187,7 @@ def __repr__(self) -> str:
                f")"
 
 
-def run_ac(network: _Network, parameters: Parameters = None, provider : str = 'OpenLoadFlow') -> _List[ComponentResult]:
+def run_ac(network: _Network, parameters: Parameters = None, provider: str = '') -> _List[ComponentResult]:
```
sylvlecl, line 190: We need to update the docstring: OpenLoadFlow is not anymore the default, depending on the new config (same for run_dc).

sylvlecl (changes requested):
Naming + doc

sylvlecl (approved):

Issue #336, opened 2022-03-15
Interface active power and apparent power limits


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the current behavior?**

Only current limits can be displayed.

* **What is the expected behavior?**

We should be able to display / modify all kinds of limits on branches: active power, apparent power, current.

The best option is probably to "replace" the current limits dataframe with a more generic operational_limits dataframe:

```python
network.get_operation_limits()
```
Which would give, for example, with different types of limits and different type of elements:

| element_id | element_type | side | type | name | value | acceptable_duration | is_fictitious |
| -- | -- | -- | -- | -- | -- | -- | -- |
| 'my_line' | LINE | ONE | CURRENT | '20min' | 1000 | 1200 | False |
| 'my_line' | LINE | ONE | CURRENT | '10min' | 1500 | 600 | False |
| 'my_line' | LINE | TWO | APPARENT_POWER | '20min' | 1000 | 1200 | False |
| 'my_line' | LINE | TWO | ACTIVE_POWER | '20min' | 1000 | 1200 | False |
| ... | ... | ... | ... | ... | ... | ... | ... |
| 'my_transfo_3' | T3T | ONE | CURRENT | '20min' | 1000 | 1200 | False |
| 'my_transfo_3' | T3T | TWO | CURRENT | '20min' | 1000 | 1200 | False |
| 'my_transfo_3' | T3T | THREE | CURRENT | '20min' | 1000 | 1200 | False |
| ... | ... | ... | ... | ... | ... | ... | ... |
| 'my_dangling_line' |  DL |  | CURRENT | '20min' | 1000 | 1200 | False |

* **What is the motivation / use case for changing the behavior?**

For example, matpower import creates apparent power limits, which are not accessible with pypowsybl.

sylvlecl:
For the modification part, I propose to implement it as on the java side:
creating limits of a given type will actually replace the previous ones of that type.

```python
# will delete previous current limits at side 1
# and create a new one
network.create_operational_limits(element_id='my_line', element_type='LINE', side='ONE',
        type='CURRENT', name='20min', value=1000, acceptable_duration=1200)
```

Permanent limits could be defined by using `np.inf` value for acceptable_duration:
```python
network.create_operational_limits(element_id='my_line', element_type='LINE', side='ONE',
        type='CURRENT', value=1000, acceptable_duration=np.inf)
```
[...]

PR #340, opened 2022-03-18
Create a dataframe with all operational limits

EtienneLt:

#336


cpp/src/bindings.cpp
```diff
@@ -184,6 +184,7 @@ PYBIND11_MODULE(_pypowsybl, m) {
             .value("PHASE_TAP_CHANGER", element_type::PHASE_TAP_CHANGER)
             .value("REACTIVE_CAPABILITY_CURVE_POINT", element_type::REACTIVE_CAPABILITY_CURVE_POINT)
             .value("CURRENT_LIMITS", element_type::CURRENT_LIMITS)
+            .value("LIMITS", element_type::LIMITS)
```
sylvlecl: Let's use the same word as in powsybl, "operational limits"

java/src/main/java/com/powsybl/python/TemporaryCurrentLimitContext.java
```diff
@@ -0,0 +1,73 @@
...
+package com.powsybl.python;
+
+import com.powsybl.iidm.network.Branch;
+
+import java.util.Objects;
+
+/**
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ */
+public class TemporaryCurrentLimitContext {
```
sylvlecl: Just naming: I know it's already used in some other places, but "context" is not really appropriate, maybe we could use "data" instead.

Context is more appropriate when you define a kind of environement, for ex. a set of data, that will be used for a specific purpose: for example SecurityAnalysisContext makes sense, because we define the context in which the security analysis will be run.

pypowsybl/network.py
```diff
@@ -2451,8 +2455,24 @@ def get_current_limits(self, all_attributes: bool = False, attributes: _List[str
         Returns:
             all current limits on the network
         """
+        warnings.warn("get_current_limits is deprecated, use get_limits instead", DeprecationWarning)
         return self.get_elements(ElementType.CURRENT_LIMITS, all_attributes, attributes)
 
+    def get_limits(self, all_attributes: bool = False, attributes: _List[str] = None) -> _DataFrame:
```
sylvlecl: `get_operational_limits` instead

sylvlecl (commented):
Just some naming comments, looks good but will do a more in-depth review later :)

java/src/main/java/com/powsybl/python/PyPowsyblApiHeader.java
```diff
@@ -343,7 +343,8 @@ private PyPowsyblApiHeader() {
         RATIO_TAP_CHANGER,
         PHASE_TAP_CHANGER,
         REACTIVE_CAPABILITY_CURVE_POINT,
-        CURRENT_LIMITS;
+        CURRENT_LIMITS,
```
sylvlecl: We could probably delete this value and guarantee non-regression on python side

pypowsybl/network.py
```diff
@@ -2492,8 +2496,24 @@ def get_current_limits(self, all_attributes: bool = False, attributes: _List[str
         Returns:
             all current limits on the network
         """
+        warnings.warn("get_current_limits is deprecated, use get_operational_limits instead", DeprecationWarning)
         return self.get_elements(ElementType.CURRENT_LIMITS, all_attributes, attributes)
 
+    def get_operational_limits(self, all_attributes: bool = False, attributes: _List[str] = None) -> _DataFrame:
+        """
+        Get the list of all limits on the network paired with their equipment id.
```
sylvlecl: We should say something about conventions for infinite duration etc.

tests/test_network.py
```diff
@@ -18,6 +18,8 @@
 import util
 import tempfile
 
+from pypowsybl._pypowsybl import NONE, CURRENT, DANGLING_LINE
```
sylvlecl: Would be better to not import things from native `_pypowsybl` module.

sylvlecl:
@EtienneLt : 
I performed small changes:
 - removed the multi-index which I think complicates things for little added value
 - reordered columns in consequence
 - removed the "CURRENT_LIMITS" enum value (the method still exists but uses the new one)
 - finally, I removed the use of the `Inf` value because it changes the type of the column to float, which can lead to mistakes

PR #341, opened 2022-03-21
Create limits

EtienneLt:

#336


java/src/main/java/com/powsybl/dataframe/network/adders/LimitsDataframeAdder.java
```diff
@@ -0,0 +1,195 @@
...
+        }
+    }
+
+    private LoadingLimitsAdder getAdder(Network network, String elementId, String type, String side, IdentifiableType identifiableType) {
+        LoadingLimitsAdder adder;
+        switch (identifiableType) {
+            case LINE:
+            case TWO_WINDINGS_TRANSFORMER:
+                switch (type) {
+                    case "CURRENT":
```
sylvlecl: It would be better to use switch/case with enums

java/src/main/java/com/powsybl/dataframe/network/adders/LimitsDataframeAdder.java
```diff
@@ -0,0 +1,195 @@
...
+                        if (side.equals("ONE")) {
+                            adder = network.getBranch(elementId).newCurrentLimits1();
```
sylvlecl: We shoudl try to avoid to search twice for the element in the network

java/src/main/java/com/powsybl/dataframe/network/adders/LimitsDataframeAdder.java
```diff
@@ -0,0 +1,195 @@
...
+            SeriesMetadata.booleans("is_fictitious")
+    );
+
+    @Override
+    public List<List<SeriesMetadata>> getMetadata() {
+        return Collections.singletonList(METADATA);
+    }
+
+    @Override
+    public void addElement(Network network, List<UpdatingDataframe> dataframes, int index) {
```
sylvlecl: We should try to change the interface to satisfy the new need of indexing before looping on lines, instead of only partially implement the existing interface and work around it by using directly the implementation in some places

sylvlecl (commented):
Some comments about implementation details on java side:
I will try to handle them

Issue #338, opened 2022-03-17
Network elements removal


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the expected behavior?**

We should be able to remove elements of the network.

* **What is the motivation / use case for changing the behavior?**

We can now create elements, but not remove them, which still does not allow all kinds of network modifications.

**Implementation** 

For single-indexed elements, the API would be straight forward:
```python
network.remove_generators(id=['GEN12', 'GEN56'])
```
Note that those methods must still be as flexible as possible, as other methods, accepting any array-like or dataframe as input.

For multi-indexed elements, it's less obvious. This includes:
 - current limits
 - ratio tap changer steps
 - phase tap changer steps
 - non linear shunt compensator sections
 - reactive capability curve points

Do we want to support the removal of only one of those objects ?

For now, creation is not supported for those kinds of elements, so I propose to not handle them at all,
in a first evolution.
If he really needs to modify those elements, the use will have to re-crate the "containing" element, for
example the transformer, for transformers steps.

EtienneLt:
solved with #347

PR #347, opened 2022-03-25
allow to remove element from network

EtienneLt:

#338


java/src/main/java/com/powsybl/python/PyPowsyblNetworkApiLib.java
```diff
@@ -398,6 +400,21 @@ public static void updateNetworkElementsWithSeries(IsolateThread thread, ObjectH
...
 
+    @CEntryPoint(name = "removeNetworkElements")
+    public static void removeNetworkElements(IsolateThread thread, ObjectHandle networkHandle, CCharPointerPointer cElementIds,
+                                             int elementCount, PyPowsyblApiHeader.ExceptionHandlerPointer exceptionHandlerPtr) {
+        doCatch(exceptionHandlerPtr, () -> {
+            Network network = ObjectHandles.getGlobal().get(networkHandle);
+            List<String> elementIds = CTypeUtil.toStringList(cElementIds, elementCount);
+            elementIds.forEach(elementId -> {
+                List<Connectable> connectables = network.getConnectableStream()
+                        .filter(connectable -> connectable.getId().equals(elementId))
```
sylvlecl: As you have suggested, we should not limit ourselves to connectables.

Trying to define the list of other identifiables we want to remove:
 - substations
 - voltage levels
 - switches
 - HVDC lines

java/src/main/java/com/powsybl/python/PyPowsyblNetworkApiLib.java
```diff
@@ -398,6 +400,21 @@ public static void updateNetworkElementsWithSeries(IsolateThread thread, ObjectH
...
     }
 
+    @CEntryPoint(name = "removeNetworkElements")
+    public static void removeNetworkElements(IsolateThread thread, ObjectHandle networkHandle, CCharPointerPointer cElementIds,
+                                             int elementCount, PyPowsyblApiHeader.ExceptionHandlerPointer exceptionHandlerPtr) {
+        doCatch(exceptionHandlerPtr, () -> {
+            Network network = ObjectHandles.getGlobal().get(networkHandle);
+            List<String> elementIds = CTypeUtil.toStringList(cElementIds, elementCount);
+            elementIds.forEach(elementId -> {
+                List<Connectable> connectables = network.getConnectableStream()
```
sylvlecl: For each element ID there can be only ONE identifiable --> no need to use a list.

Actually the more efficient will be to use `network.getIdentifiable(String id)`, in order to avoid linear searches.

sylvlecl (approved):

Issue #328, opened 2022-03-07
Modifying/adding properties on network elements


nicolaslhuillier:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
Though optional attributes can be retrieve in element dataframes (via get_XXX() functions, e.g. 'get_generators(True)'),
it is currently only possible to modify IIDM attributes (via update_XXX() functions).
Attempt to modify an existing property YYY raises the error message :"no column named YYY"
 
* **What is the expected behavior?**
It would be very useful to be able to modify attributes that are not part of core IIDM (e.g. "property" or "extension").
If the attribute does not exist, it could even be added as a new property.

* **What is the motivation / use case for changing the behavior?**
The use-case is to attach custom information to network elements in order to ease further processing (e.g. time-series mapping)

sylvlecl:
For the handling of arbitrary properties, it would be better to not mix the creation of properties with `update_***` methods, because it will be error-prone: the user can easily end up creating new columns whereas he wanted to update an existing one.

For this reason, I would propose to have a separate method:
```python
network.add_elements_properties(id=['GEN1', 'GEN2'], my_prop=['value1', 'value2'])
```

Then for removal:
```python
network.remove_elements_properties(properties=['my_prop'], id=['GEN1', 'GEN2'])
```

Eventually, it raises the question of properties handling in getters:
to keep consistent, we could remove those columns from the usual `get_***` methods,
and have a unique `get_elements_properties` method, similarly to extensions handling.

Obviously this would be a breaking change. However this would clarify an ambiguity around properties,
where users may be surprised that some data are strings (see #199). It would also avoid
the inflation of dataframe columns.

Any opinion on this proposition @geofjamg @nicolaslhuillier @EtienneLt @vincentbarbesant ?

geofjamg:
Fully agree with the proposition but only worried about the breaking change.
Maybe as an intermediate version properties could be duplicated in both the element table and the property one ? And then remove them from element table next version?
With a warning when gettting a table with properties to force the user to migrate?

nicolaslhuillier:
I also agree it is better to distinguish between "update" and "add".
I understand the point on "getters", but from a user's perspective it is more convenient if iidm attributes, properties and extensions all come directly in the same dataframe (without the need of a merge)

sylvlecl:
Indeed the consequence of removing from the existing table would be to force the user to have wrapper methods such as:

```python
[...]

PR #371, opened 2022-04-19
add the possibility to create and remove properties on network elements

EtienneLt:

#328


docs/reference/network.rst
```diff
@@ -172,6 +172,17 @@ Network variants may be used to manage multiple states of the network efficientl
    Network.remove_variant
    Network.get_variant_ids
 
+Network properties management
+---------------------------
+
+Network properties can be added and removed for all network's elements.
```
sylvlecl: I would rather move it to the existing section about elements update, I don't think it deserves a whole new section

pypowsybl/network.py
```diff
@@ -3772,6 +3774,49 @@ def get_extension(self, extension_name: str) -> _DataFrame:
...
+        Args:
+            df: the properties to be created or updated. The index has to be `id`
+                    identifying the network elements.
+            kwargs: the properties to be added as string.
+                    Arguments can be single values or any type of sequence.
+                    In the case of sequences, all arguments must have the same length.
+        """
+        if df is None:
+            df = _adapt_properties_kwargs(**kwargs)
+        if (df.isnull().sum().sum() > 0):
```
sylvlecl: We can probably use `.any()` instead of `.sum()` to get a straight boolean (and save some computation).

pypowsybl/network.py
```diff
@@ -3772,6 +3774,49 @@ def get_extension(self, extension_name: str) -> _DataFrame:
...
+            raise _pp.PyPowsyblError("dataframe can not contain NaN values")
+        for series_name in df.columns.values:
+            df[series_name] = df[series_name].astype(str)
+        c_df = _create_properties_c_dataframe(df)
+        _pp.add_network_element_properties(self._handle, c_df)
+
+    def remove_elements_properties(self, df: _DataFrame = None, **kwargs: _ArrayLike) -> None:
```
sylvlecl: For this specific method, I think we should be more specific about arguments:
we must have one "id" and one "properties" arguments, and each one can be a string or sequence of strings.

Dataframe does not seem appropriate since we will never have multiple columns: it is too generic!

PS:
Actually there was a misunderstanding about what it should do, see comment in the unit test

pypowsybl/network.py
```diff
@@ -3772,6 +3774,49 @@ def get_extension(self, extension_name: str) -> _DataFrame:
         """
         return _create_data_frame_from_series_array(_pp.create_network_elements_extension_series_array(self._handle, extension_name))
 
+
+    def add_elements_properties(self, df: _DataFrame = None, **kwargs: _ArrayLike) -> None:
+        """
+        add network element properties with data provided as a :class:`~pandas.DataFrame` or as named arguments.
```
sylvlecl: Detail: for the docs we should start with a capital letter ;)

java/src/main/java/com/powsybl/python/PyPowsyblNetworkApiLib.java
```diff
@@ -555,6 +552,50 @@ public static void freeDataframesMetadata(IsolateThread thread, DataframesMetada
...
+    public static void addNetworkElementProperties(IsolateThread thread, ObjectHandle networkHandle, DataframePointer properties, ExceptionHandlerPointer exceptionHandlerPtr) {
+        doCatch(exceptionHandlerPtr, () -> {
+            UpdatingDataframe propertiesDataframe = createDataframe(properties);
+            Network network = ObjectHandles.getGlobal().get(networkHandle);
+            for (int i = 0; i < Objects.requireNonNull(propertiesDataframe).getLineCount(); i++) {
+                for (SeriesMetadata column : propertiesDataframe.getSeriesMetadata()) {
+                    if (!column.isIndex()) {
+                        String seriesName = column.getName();
+                        if (column.getType() == SeriesDataType.STRING) {
+                            String data = propertiesDataframe.getStringValue(seriesName, i)
```
sylvlecl: I just changed a little the `UpdatingDataframe` methods:
this will still work, but it would be better to get the column only once (as a `StringSeries`), and then re-use it for each row.

java/src/main/java/com/powsybl/python/PyPowsyblNetworkApiLib.java
```diff
@@ -555,6 +552,50 @@ public static void freeDataframesMetadata(IsolateThread thread, DataframesMetada
...
+                                    .orElseThrow(() -> new PowsyblException("id is missing"));
+                            String id = propertiesDataframe.getStringValue("id", 0, i)
```
sylvlecl: Same comment as above about getting the "id" column only once as a `StringSeries`.

tests/test_network.py
```diff
@@ -1247,6 +1247,40 @@ def test_get_empty_attributes(self):
...
+        pd.testing.assert_frame_equal(network.get_lines(attributes=['prop1', 'prop2', 'prop3']), expected, check_dtype=False)
+        removed_properties = pd.DataFrame.from_records(index='id', data=[
+            {'id': 'GEN', 'properties': 'prop1'},
+            {'id': 'GEN', 'properties': 'prop2'}
+        ])
+        network.remove_elements_properties(removed_properties)
+        columns = network.get_generators(all_attributes=True).columns
+        self.assertTrue('prop1' not in columns and 'prop2' not in columns and 'prop3' not in columns)
+        network.remove_elements_properties(id=['NHV1_NHV2_2'], properties=['prop3'])
+        network.remove_elements_properties(id=['NHV1_NHV2_1', 'NHV1_NHV2_1'], properties=['prop2', 'prop1'])
```
sylvlecl: Actually the behaviour I expected was that
```python
network.remove_elements_properties(id='NHV1_NHV2_1', properties=['prop2', 'prop1'])
```
would remove both properties on the line NHV1_NHV2_1.

And
```python
network.remove_elements_properties(id=['E1', 'E2', 'E3'], properties=['prop2', 'prop1'])
```
would remove prop2 and prop1 for the 3 elements E1, E2, and E3.

sylvlecl (commented):
Details + discussion about removal behaviour

sylvlecl (approved):

PR #374, opened 2022-04-21
Migrate to PowSyBl dependencies 1.1.0 (core 4.8.0)

geofjamg:

**Other information**:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,110 @@ on:
       - 'v*.*.*'
 
 jobs:
+  manylinux_build:
+    name: Build linux ${{ matrix.python.name }} wheel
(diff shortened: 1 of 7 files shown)


sylvlecl:
The issue seems to be this one:
https://github.com/actions/setup-python/issues/279

geofjamg:
> The issue seems to be this one:
> [actions/setup-python#279](https://github.com/actions/setup-python/issues/279)

This workaround seems to work very well. I also tested to fix math native (with a 1.2.1) so that MacOS amd 64 lib is compiled on MacOS 10 instead of 11 (like for math native 1.1.0) and it works too. I think we should keep MacOS 11 migration to another PR later and stay on 10 until build is supported by github actions.

Issue #358, opened 2022-04-13
[Loadflow] Implementation-specific loadflow parameters


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the current behavior?**

Implementation-specific parameters, for example `OpenLoadFlowParameters`, cannot be provided through the python interface.

* **What is the expected behavior?**

We can provide a map of properties that will be loaded as implementation specific parameters:
```python
import pypowsybl.loadflow as lf
params = lf.Parameters()
params.provider_parameters = { 'maxIteration': '50' }

# OpenLoadFlowParameters will be loaded from provider_parameters
lf.run_ac(network, parameters=params, provider='OpenLoadFlow') 
```

At least in a first implementation, parameters should be a `dict[str, str]`.

Also, we should provide the list of parameter names for each loadflow provider:
```python
>>> lf.get_provider_parameters_names()  # parameters names for default provider
['maxIteration', ... ]
>>> lf.get_provider_parameters_names('OpenLoadFlow)  # parameters names for the specified provider
['maxIteration', ... ]
```

PR #377, opened 2022-04-22
Implement specific loadflow parameters

EtienneLt:

#358


cpp/src/bindings.cpp
```diff
@@ -114,6 +114,8 @@ std::shared_ptr<load_flow_parameters> initLoadFlowParameters() {
         parameters->countries_to_balance = pypowsybl::copyVectorStringToCharPtrPtr(configCountries);
         parameters->countries_to_balance_count = configCountries.size();
         parameters->connected_component_mode = config_params->connected_component_mode;
+        parameters->specific_parameters_keys_count = config_params->specific_parameters_keys_count;
```
sylvlecl: Naming: see in the issue, I think "provider_parameters" will be a better name everywhere than "specific_parameters" because we don't know with respect to what they are specific.

pypowsybl/loadflow.py
```diff
@@ -168,6 +172,12 @@ def __init__(self,
             self.countries_to_balance = countries_to_balance
         if connected_component_mode is not None:
             self.connected_component_mode = connected_component_mode
+        if specific_parameters is not None:
```
sylvlecl: For the user we should expose the parameters as a dict.
The simplest way is probably to provide a property getter and a property setter for this dict, which will use the underlying lists.
This will not allow to modify the parameters in place, for ex. like that:
```python
parameters.provider_parameters['prop'] = ['val'] # not working
```
But defining the whole dict will be possible like in the constructor:
```python
parameters.provider_parameters = { 'prop': 'val'}
```
and getting the parameters will return a dict:
```python
>>> parameters.provider_parameters
{ 'prop': 'val'}
```

The underlying lists should be hidden as protected members (leading `_`).

sylvlecl: Actually, maybe the best option would be break the link between the C struct and the python class:
instead of having `Parameters` inherit `_pypowsybl.LoadFlowParameters`, it should be an independent class.

When we get parameters from java, we would perform a conversion from the second to the first (in particular transforming 2 lists to dict), and when we send parameters to java, we will do the opposite.

This way we could support a straight dict in the exposed  `Parameters` class.

sylvlecl (changes requested):
Naming + I think we should change how the dict is exposed to python user

sylvlecl (approved):

Issue #376, opened 2022-04-22
Core dump while logging


geofjamg:
* **Do you want to request a *feature* or report a *bug*?**
bug

* **What is the current behavior?**
When interrupting a long processing that is logging is we press ctrl-c to stop the processing we end up with a core dump:
```python
DEBUG:powsybl:2 windings transformer 'XXX': , temporary limit should be greater than permanent limit
DEBUG:powsybl:2 windings transformer 'YYY': , temporary limit should be greater than permanent limit
DEBUG:powsybl:2 windings transformer 'ZZZZ': , temporary limit should be greater than permanent limit
DEBUG:powsybl:2 windings transformer 'UUUU': , temporary limit should be greater than permanent limit
^Cterminate called after throwing an instance of 'pybind11::error_already_set'
  what():  KeyboardInterrupt: 

At:
  /usr/lib/python3.8/logging/__init__.py(1485): log
  /home/jamgotchiangeo/PycharmProjects/pypowsybl/pypowsybl/network.py(3978): load
  <stdin>(1): <module>

Abandon (core dumped)
```

* **What is the expected behavior?**
Same behaviour as without logging

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #378, opened 2022-04-26
Fix for issue #376

obrix:
Try fixing the interruption of powsybl process badly quitting the python interpreter by properly detecting the interruption and ending the java process by throwing an error.

Signed-off-by: Bertrand Rix <bertrand.rix@artelys.com>

Issue #376 

**Other information**:


sylvlecl:
I have been thinking a little more about that issue:

1. Finally, I think we should keep the logback behaviour of "swallowing" exceptions.
This is because a typical use-case of loggers is to be called inside `catch` statements, where we should avoid to rethrow something (in particular in `doCatch` method, otherwise the process crashes again).

2. Then, we should probably still avoid to swallow the keyboard interrupt. In order to achieve this, we could try to restore the python exception at the end of the java call.
Note that this will not actually interrupt the java work, but this is the same behaviour as today without logging.

3. Finally, if we wanted to actually enable the user to interrupt long java calls, it would require a lot more work:
  - translating the python interrupt into java `Thread.interrup()`
  - more importantly, long java functions of powsybl such as loading a network would need to check for interrupts so that it actually does something

obrix:
@sylvlecl Great !

> Finally, I think we should keep the logback behaviour of "swallowing" exceptions.

I think this is still the case in the provided solution

> Then, we should probably still avoid to swallow the keyboard interrupt. In order to achieve this, we could try to restore the python exception at the end of the java call.
Note that this will not actually interrupt the java work, but this is the same behaviour as today without logging.

This cannot be done through a classic exception from AppenderBase.append, this is the reason I used an Error which inherit from Throwable and thus go through AppenderBase.doAppend exception filter. This has the effect of interrupting the long process immediatly. If we want to continue the process and "restore the python exception at the end of the java call" this become more complicated.
Maybe from AppenderBase.append where I throw the error I can instead raise a flag in PypowsyblApiLib that will be used to throw an exception from Util.doCatch after the run. This should implement a behavior similar to what you are describing, not interrupting the process but ending the call with an exception. What do you think ?

sylvlecl:
> I think this is still the case in the provided solution

Not really since we throw the Error! I mean that after all, it's better to not throw at all (or throw a normal exception that logback will catch, but it's not very useful)

> Maybe from AppenderBase.append where I throw the error I can instead raise a flag in PypowsyblApiLib that will be used to throw an exception from Util.doCatch after the run. This should implement a behavior similar to what you are describing, not interrupting the process but ending the call with an exception. What do you think ?

I think to achieve this we could use the python primitives, and bypass completely the java part:

1. if a `py::error_already_set` is thrown by pybind11 in the c++ callback, we can "restore it" to the python C API representation and continue java work normally:
```cpp
catch(py::error_already_set& err) {
    err.restore();
}
```

2. then at the end of the java call, on c++ side, we can check if a python error has been stored:
```cpp
if (PyErr_Occurred() != nullptr) {
    throw py::error_already_set();
}
```
This way the calling python code would receive the inner exception after the java work has completed.

obrix:
@sylvlecl As discussed I tried the described approach in the last commit. Catching the exception and restoring it let the java process go to its end but PyErr_Occured does not allow to rethrow it at the end. This has the effect of disabling interruption with ctrl+c. Do you see anything that differs from what you tested ?

cpp/src/pylogging.cpp
```diff
@@ -35,12 +35,30 @@ py::object CppToPythonLogger::getLogger() {
...
+    ~save_python_error() {
+        if (PyErr_Occurred() == nullptr) {
+            PyErr_Restore(type, value, trace); 
+        }
+    }
+};
+
 void logFromJava(int level, long timestamp, char* loggerName, char* message) {
+    py::gil_scoped_acquire acquire;
+    save_python_error previousError;  // to keep and restore the previously set exception, if any
```
sylvlecl, line 53: @obrix I added this mechanism to allow to keep the initial exception during the subsequent log calls.

It seems that without this, after the log call which sets the error, the previously set error is cleared (the python error is set at the beginning of the call, but not set anymore at the end ...).

With that additional workaround, the KeyboardInterrupt exception is correctly passed all the way until the end of the java call, and re-thrown thanks to your code.
From python side, we correctly receive it:
```
Traceback (most recent call last):
  File "toto.py", line 8, in <module>
    network = pn.load('/home/leclercsyl/feature_tests/01-test-cases/02-full/pf_20200326T1635Z_20200326T1635Z.xiidm')
  File "/home/leclercsyl/Projects/sources/powsybl/pypowsybl/pypowsybl/network.py", line 4058, in load
    return Network(_pp.load_network(file, parameters))
  File "/usr/lib/python3.8/logging/__init__.py", line 1497, in log
    def log(self, level, msg, *args, **kwargs):
KeyboardInterrupt
``` 

It looks good this way :)

sylvlecl (approved):

cpp/src/pylogging.cpp
obrix: Cool, thank you!

Issue #354, opened 2022-04-12
Generators reactive limits creation and modifications


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the current behavior?**

Reactive limits of generators cannot be changed for existing generators
They cannot be created either when creating generators.

* **What is the expected behavior?**

We need to be able to create and change reactive limits of generators.

If we follow the java API, we can support creation only as a separate step from generators creation.

We could provide either one or 2 methods for the 2 types of reactive limits representations:
```python
network.create_curve_reactive_limits(points_data)
network.create_minmax_reactive_limits(minmax_data)

#or
network.create_reactive_limits(points_data, kind=CURVE)
```

PR #379, opened 2022-04-27
Creates and updates generators' reactive limits

CBiasuzzi:

#354

**Other information**:


java/src/main/java/com/powsybl/dataframe/network/adders/MinMaxReactiveLimitsDataframeAdder.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    @Override
+    public void addElements(Network network, List<UpdatingDataframe> dataframes) {
+        UpdatingDataframe primaryTable = dataframes.get(0);
+        MinMaxReactiveLimitsSeries series = new MinMaxReactiveLimitsSeries(primaryTable);
+        for (int i = 0; i < primaryTable.getRowCount(); i++) {
+            String elementId = series.getElementIds().get(i);
+            Double minQ = series.getMinQs().get(i);
```
sylvlecl: we'd better use primitive `double`

java/src/main/java/com/powsybl/dataframe/network/adders/CurveReactiveLimitsDataframeAdder.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    @Override
+    public void addElements(Network network, List<UpdatingDataframe> dataframes) {
+        UpdatingDataframe primaryTable = dataframes.get(0);
+        CurveReactiveLimitsSeries series = new CurveReactiveLimitsSeries(primaryTable);
+        Map<String, SortedSet<CurvePoint>> curvesPoints = new HashMap<>();
+        for (int i = 0; i < primaryTable.getRowCount(); i++) {
+            String elementId = series.getElementIds().get(i);
+            Integer num = series.getNums().get(i);
```
sylvlecl: we'd better use primitives here too

java/src/main/java/com/powsybl/dataframe/network/adders/CurveReactiveLimitsDataframeAdder.java
```diff
@@ -0,0 +1,152 @@
...
+            Double p = series.getPs().get(i);
+            Double minQ = series.getMinQs().get(i);
+            Double maxQ = series.getMaxQs().get(i);
+            CurvePoint curvePoint = new CurvePoint(num, p, minQ, maxQ);
+            if (!curvesPoints.containsKey(elementId)) {
```
sylvlecl: obviously not very important but using `curvesPoints.computeIfAbsent` would save a search and a line of code :)

java/src/main/java/com/powsybl/dataframe/network/adders/CurveReactiveLimitsDataframeAdder.java
sylvlecl: I think that we could spare the user the necessity to define the number for points:
we could just add them in the order of the table ?

Actually, in the adder implementation, the order is not even used: points will be sorted by P. (it's not really specified in the interface though!)

Also, in similar tables (shunts sections and transfo steps), it's done this way, so it would be more consistent.

sylvlecl (commented):
Mainly, proposition to remove the `num` column for curve points.

java/src/main/java/com/powsybl/dataframe/network/adders/MinMaxReactiveLimitsDataframeAdder.java
CBiasuzzi: fixed. Thanks.

java/src/main/java/com/powsybl/dataframe/network/adders/CurveReactiveLimitsDataframeAdder.java
CBiasuzzi: fixed. Thanks.

java/src/main/java/com/powsybl/dataframe/network/adders/CurveReactiveLimitsDataframeAdder.java
CBiasuzzi: fixed. Thanks.

java/src/main/java/com/powsybl/dataframe/network/adders/CurveReactiveLimitsDataframeAdder.java
CBiasuzzi: column 'num', for points, removed. Indeed it was redundant. Thanks.

sylvlecl (approved):

Issue #385, opened 2022-05-03
Support for pathlib.Path


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**

pypowsybl methods that accept a file path as argument only accept strings.
In particular: pp.network.load, pp.network.Network.dump 

* **What is the expected behavior?**

We should provide straight support for `Path` objects, for example supporting:
```python
from pathlib import Path
dir = Path('/path/to_dir')
network = pp.network.load(dir / 'case.xiidm')
```

* **What is the motivation / use case for changing the behavior?**

Conveniency.

PR #388, opened 2022-05-06
add support for pathlib.path for load and dump network

EtienneLt:

#385


pypowsybl/network.py
```diff
@@ -251,9 +252,12 @@ def dump(self, file: str, format: str = 'XIIDM', parameters: ParamsDict = None)
                 network.dump('network.xiidm.gz')  # produces a gzipped file
                 network.dump('/path/to/network.uct', format='UCTE')
         """
+        if isinstance(file, pathlib.Path):
+            file = file.resolve()
+        file_str = str(file)
```
sylvlecl: I've been digging a little more the doc about path objects:

it turns out we can be more generic than just `pathlib.Path`, by using `os.PathLike` instead, and its `__fspath__` method to get the string representation.

pypowsybl/network.py
```diff
@@ -230,7 +231,7 @@ def connect(self, id: str) -> bool:
     def disconnect(self, id: str) -> bool:
         return _pp.update_connectable_status(self._handle, id, False)
 
-    def dump(self, file: str, format: str = 'XIIDM', parameters: ParamsDict = None) -> None:
+    def dump(self, file: _Union[str, pathlib.Path], format: str = 'XIIDM', parameters: ParamsDict = None) -> None:
```
sylvlecl: Similarly, we can allow path objects for the 2 methods "write single line diagram" and "write network area diagram".

sylvlecl (commented):
Minor improvements suggestions, after more digging of the docs

PR #394, opened 2022-05-11
GLSK import 

obrix:
Introduce glsk file loading and zone creation using loaded data.

**Other information**:


tests/test_glsk.py
```diff
@@ -0,0 +1,37 @@
...
+        self.assertEqual(datetime.datetime(2019, 1, 9, 0, 0), importer.get_gsk_time_interval_end())
+        t = importer.get_gsk_time_interval_start()
+        self.assertEqual(['FFR1AA1 ', 'FFR2AA1 ', 'FFR3AA1 '], importer.get_points_for_country('10YFR-RTE------C', t.timestamp()))
+        self.assertEqual([0.0158, 0.1299, 0.1299], importer.get_glsk_factors('10YFR-RTE------C', t.timestamp()))
+
+    def test_zone(self):
+        n = pp.network.load('../data/simple-eu.uct')
+        importer = pp.glsk.GLSKImporter("../data/glsk_sample.xml")
+        t = importer.get_gsk_time_interval_start()
+        de_generators = importer.get_points_for_country('10YCB-GERMANY--8', t.timestamp())
```
geofjamg: Maybe we should also provide a simpler facade API to be able to write:
```python
zones = pp.sensitivity.create_zones_from_glsk_file(n, '../data/glsk_sample.xm', datetime)
```

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
```diff
@@ -812,4 +814,55 @@ public static void setLogLevel(IsolateThread thread, int logLevel, ExceptionHand
...
+    @CEntryPoint(name = "getGLSKcountries")
+    public static ArrayPointer<CCharPointerPointer> getGLSKcountries(IsolateThread thread, ObjectHandle importerHandle, ExceptionHandlerPointer exceptionHandlerPtr) {
+        return doCatch(exceptionHandlerPtr, () -> {
+            GLSKimportContext importer = ObjectHandles.getGlobal().get(importerHandle);
+            return createCharPtrArray(new ArrayList<>(importer.getCountries()));
+        });
+    }
+
+    @CEntryPoint(name = "getInjectionFactor")
+    public static ArrayPointer<CDoublePointer> getInjectionFactor(IsolateThread thread, ObjectHandle importerHandle, CCharPointer countryPtr, double instant, ExceptionHandlerPointer exceptionHandlerPtr) {
```
geofjamg: instant should be a long?

pypowsybl/glsk.py
```diff
@@ -0,0 +1,30 @@
...
+    def get_countries(self):
+        return _pypowsybl.get_glsk_countries(self.handle)
+
+    def get_points_for_country(self, country, instant):
+        return _pypowsybl.get_glsk_injection_keys(self.handle, country, instant)
+
+    def get_glsk_factors(self, country, instant):
+        return _pypowsybl.get_glsk_factors(self.handle, country, instant)
+
+    def get_points_count_for_country(self, country):
```
geofjamg: To remove?

pypowsybl/glsk.py
```diff
@@ -0,0 +1,30 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+import pypowsybl._pypowsybl as _pypowsybl
+from datetime import datetime
+
+class GLSKImporter:
+    def __init__(self, file):
+        self.handle = _pypowsybl.create_glsk_importer(file)
+
+    def get_gsk_time_interval_start(self):
```
geofjamg: Instead of get_gsk_time_interval_start and get_gsk_time_interval_end we could get a list of timedate

java/pom.xml
```diff
@@ -75,6 +75,7 @@
         <maven-shade-plugin.version>3.2.4</maven-shade-plugin.version>
         <powsybl-dependencies.version>1.1.0</powsybl-dependencies.version>
         <powsybl-math-native.version>1.2.1</powsybl-math-native.version>
+        <powsybl-entsoe.version>1.4.0</powsybl-entsoe.version>
```
geofjamg: You don't need to specify the version, it is imported from powsybl dependencies with the right one

cpp/CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ ExternalProject_Add(native-image
     DEPENDS mvn
     SOURCE_DIR ${PYPOWSYBL_JAVA_BIN_DIR}
     DOWNLOAD_COMMAND ""
-    PATCH_COMMAND $ENV{JAVA_HOME}/bin/native-image ${NATIVE_IMAGE_BUILD_OPTIONS} --class-path ${PYPOWSYBL_JAVA_SRC_DIR}/target/pypowsybl-java.jar${EXTRA_JARS} --no-fallback --allow-incomplete-classpath --shared -H:Name=pypowsybl-java -H:CLibraryPath=${CMAKE_CURRENT_SOURCE_DIR}/src
+    PATCH_COMMAND $ENV{JAVA_HOME}/bin/native-image ${NATIVE_IMAGE_BUILD_OPTIONS} --class-path ${PYPOWSYBL_JAVA_SRC_DIR}/target/pypowsybl-java.jar${EXTRA_JARS} --no-fallback --allow-incomplete-classpath --report-unsupported-elements-at-runtime --shared -H:Name=pypowsybl-java -H:CLibraryPath=${CMAKE_CURRENT_SOURCE_DIR}/src
```
geofjamg: Why do you need this?

geofjamg (changes requested):

cpp/CMakeLists.txt
obrix: It does not build if not provided. I have not looked into it yet.

java/src/main/java/com/powsybl/python/PyPowsyblApiLib.java
obrix: Yes I haven't pushed yet but I switched with long when relevant.

tests/test_glsk.py
obrix: Good idea!

pypowsybl/glsk.py
obrix: I'm not sure we can provide a list ? This is just a wrapping of UCTEGlskDocument that also provide start / end api. Querying data with any timestamp between start / end will work.

geofjamg: you are right, this is just an interval

pypowsybl/sensitivity.py
```diff
@@ -94,6 +95,28 @@ def create_country_zone(network: _Network, country: str,
...
+    # join generators, voltage levels and substations to get generators with countries
+    generators = network.get_generators()
+    generators_with_countries = generators.join(
+        voltage_levels[['substation_id']].join(substations[['country']], on=['substation_id']),
+        on=['voltage_level_id'])
+    shift_keys_by_id = dict(zip(generator_index, shift_keys))
+
+    return Zone(country, shift_keys_by_id)
+
+def create_zones_from_glsk_file(network: _Network, glsk_file: str, instant: datetime):
```
geofjamg: It seems we don't need the network in fact for loading zones

geofjamg: Oups we need, forget my remark

geofjamg (commented):
Looks good. I'm going to test it on real cases.
In the meantime, could you add the documentation?
(and fix the CI)

annetill:
I am not sure to understand all your PR but it seems that the best way to import a GLSK file coming from UCTE format and to convert it in a VariableSet understandable by a sensitivity analysis is to interface this java method:
````
ZonalData<SensitivityVariableSet> ucteGlskProvider = UcteGlskDocument.importGlsk(getClass().getResourceAsStream("/20170322_1844_SN3_FR2_GLSK_test.xml"))
                .getZonalGlsks(network, instant);
````
You have to give the IIDM network as argument to complete the conversion, coming from the UCTE network file associated to the GLSK file you are loading. In that way you will have the good factors associated to the good injections.

**Comment by geofjamg** [comment]
@obrix if you can update your PR, normally you can remove the graalvm option since last powsybl update

**Comment by obrix** [comment]
@geofjamg I rebased the branch on main and removed the report-unsupported-elements-at-runtime option, it works without it now.

File: java/src/main/java/com/powsybl/python/GLSKimportContext.java [hunk 8]
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * @author Bertrand Rix {@literal <bertrand.rix at artelys.com>}
+ */
+public class GLSKimportContext {
+
+    private UcteGlskDocument document;
+
+    public void load(String filename) {
+        try {
+            document = UcteGlskDocument.importGlsk(new FileInputStream(filename));
```
geofjamg: Should be closed?

pypowsybl/sensitivity.py
```diff
@@ -94,6 +96,35 @@ def create_country_zone(network: _Network, country: str,
 
     return Zone(country, shift_keys_by_id)
 
+def create_country_zone_generator(country: str, generator_index: _List[str], shift_keys: _List[float]) -> Zone:
```
geofjamg: IDs could be any injection so we should rename the method and as first param is just a string it could not be a country id. You should better instead just add another constructor to Zone that take 2 list instead of the dict

cpp/src/pypowsybl.h
```diff
@@ -158,6 +158,8 @@ void deleteCharPtrPtr(char** charPtrPtr, int length);
 
 ::zone* createZone(const std::string& id, const std::vector<std::string>& injectionsIds, const std::vector<double>& injectionsShiftKeys);
 
+::zone* createZone(const std::string& id, const JavaHandle& glskImportHandle);
```
geofjamg: Where is this method used?

geofjamg (changes requested):

pypowsybl/glsk.py
```diff
@@ -0,0 +1,30 @@
...
+#
+from typing import List as _List
+from datetime import datetime
+import pypowsybl._pypowsybl as _pypowsybl
+from pypowsybl.network import Network as _Network
+
+#GLSKImporter helper class
+class GLSKImporter:
+    def __init__(self, file: str):
+        self.handle = _pypowsybl.create_glsk_importer(file)
```
sylvlecl: We should hide the java handle from the user: to be renamed to `_handle`

pypowsybl/glsk.py
```diff
@@ -0,0 +1,30 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+from typing import List as _List
+from datetime import datetime
+import pypowsybl._pypowsybl as _pypowsybl
+from pypowsybl.network import Network as _Network
+
+#GLSKImporter helper class
+class GLSKImporter:
```
sylvlecl: From the API point of view: 
for better consistency with network API, maybe we could have a `GLSKDocument` class and a `load` method ?

```python
gslk_document = pp.glsk.load('/path/to/doc')
gslk_document.get_glsk_factors(...)
```

In the current API, it feels like the document will be imported after the creation of the importer, whereas it is actually imported as soon as the object is built.

Also: we should support `os.PathLike` as an input to the loading method, together with `str` (see network.load).

pypowsybl/sensitivity.py
obrix: You mean remove create_country_zone_generator and replace it by a more generic Zone constructor taking the two lists ? That's it ?

geofjamg: Yes an additional constructor

obrix: As discussed I just renamed the function with a more generic name.

java/src/main/java/com/powsybl/python/GLSKimportContext.java
```diff
@@ -0,0 +1,55 @@
...
+import java.io.IOException;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * @author Bertrand Rix {@literal <bertrand.rix at artelys.com>}
+ */
+public class GLSKimportContext {
```
geofjamg: GLSKDocumentContext ?

obrix: done

sylvlecl (approved):

PR #407, opened 2022-05-25
add min max limits for vsc and batteries

EtienneLt:

feature

min_q and max_q attributes are now avalaible for vsc stations and batteries (get, create and update)

### DIFF
--- a/tests/test_network_elements_creation.py
+++ b/tests/test_network_elements_creation.py
@@ -629,17 +629,37 @@ def test_create_limits():
 
 
 def test_create_minmax_reactive_limits():
-    network = pn.create_eurostag_tutorial_example1_network()
+    network = pn.create_four_substations_node_breaker_network()
     network.create_minmax_reactive_limits(pd.DataFrame.from_records(index='id', data=[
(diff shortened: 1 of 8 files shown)


EtienneLt:
I forgot docstring, I'll add that

tests/test_network_elements_creation.py
```diff
@@ -629,17 +629,37 @@ def test_create_limits():
 
 
 def test_create_minmax_reactive_limits():
-    network = pn.create_eurostag_tutorial_example1_network()
+    network = pn.create_four_substations_node_breaker_network()
+    network.create_minmax_reactive_limits(pd.DataFrame.from_records(index='id', data=[
```
sylvlecl, line 633: We need to support also `create_curve_reactive_limits` method for VSCs and batteries.

This makes me think that we should also update `get_reactive_capability_curve_points` to return curves of VSCs and batteries.

Maybe we will need to list this as a breaking change in the release notes, because users code may assume that they only receive generators data.

EtienneLt: maybe it is another pr ?

sylvlecl (approved):

Issue #403, opened 2022-05-24
Slack terminal support


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the current behavior?**

The slack terminal extension is not exposed.

* **What is the expected behavior?**

We should be able to manipulate that extension.
For example, the dataframe could look like:
```python
>>> network.get_extensions('slackTerminal')
       slack_element_id
id
'VL-1'           'BBS-1'
```

For the creation, we may support only a limited set of input types:
- bus bar sections for node breaker topologies
- buses for bus breaker topologies

* **What is the motivation / use case for changing the behavior?**

Being able to explicitly define the slack terminal.

EtienneLt:
done

PR #414, opened 2022-06-02
add slack Terminal extension

EtienneLt:

#403

### DIFF
--- a/java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeAdder.java
+++ b/java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeAdder.java
@@ -56,10 +56,11 @@ public class SlackTerminalDataframeAdder extends AbstractSimpleAdder {
                 }
             } else if (this.busId != null && this.elementId == null) {
                 String busId = this.busId.get(row);
-                if (network.getBusView().getBus(busId) == null) {
+                Bus bus = network.getBusView().getBus(busId);
+                if (bus == null) {
                     throw new PowsyblException("bus id : " + busId + " does not match any existing bus");
                 }
-                SlackTerminal.attach(network.getBusView().getBus(busId));
+                SlackTerminal.attach(bus);
             } else {
                 throw new PowsyblException("only one of element_id or bus_id must be filled");
             }
--- a/java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeAdder.java
+++ b/java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeAdder.java
@@ -1,0 +1,82 @@
+/**
+ * Copyright (c) 2022, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package com.powsybl.dataframe.network.extensions;
+
+import com.powsybl.commons.PowsyblException;
+import com.powsybl.dataframe.SeriesMetadata;
+import com.powsybl.dataframe.network.adders.AbstractSimpleAdder;
+import com.powsybl.dataframe.update.StringSeries;
+import com.powsybl.dataframe.update.UpdatingDataframe;
+import com.powsybl.iidm.network.*;
+import com.powsybl.iidm.network.extensions.SlackTerminal;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * @author Etienne Lesot <etienne.lesot@rte-france.com>
+ */
+public class SlackTerminalDataframeAdder extends AbstractSimpleAdder {
+
+    private static final List<SeriesMetadata> METADATA = List.of(
+            SeriesMetadata.stringIndex("voltage_level_id"),
+            SeriesMetadata.strings("element_id"),
+            SeriesMetadata.strings("bus_id"));
+
+    private static class SlackTerminalSeries {
+        private final StringSeries voltageLevelId;
+        private final StringSeries elementId;
+        private final StringSeries busId;
+
+        SlackTerminalSeries(UpdatingDataframe dataframe) {
+            this.voltageLevelId = dataframe.getStrings("voltage_level_id");
+            this.elementId = dataframe.getStrings("element_id");
+            this.busId = dataframe.getStrings("bus_id");
+        }
+
+        void create(Network network, int row) {
+            VoltageLevel voltageLevel = network.getVoltageLevel(this.voltageLevelId.get(row));
+            if (voltageLevel == null) {
+                throw new PowsyblException("voltage_level_id : " + this.voltageLevelId.get(row) +
+                        " does not correspond to any voltage level");
+            }
+            if (this.elementId != null && this.busId == null) {
+                Identifiable identifiable = network.getIdentifiable(this.elementId.get(row));
+                if (identifiable instanceof Injection) {
+                    Terminal terminal = ((Injection) identifiable).getTerminal();
+                    SlackTerminal.reset(terminal.getVoltageLevel(), terminal);
+                } else if (identifiable instanceof Bus) {
+                    SlackTerminal.attach((Bus) identifiable);
+                } else {
+                    throw new PowsyblException("only injections or configured buses are handled as an \"element_id\"");
+                }
+            } else if (this.busId != null && this.elementId == null) {
+                String busId = this.busId.get(row);
+                if (network.getBusView().getBus(busId) == null) {
+                    throw new PowsyblException("bus id : " + busId + " does not match any existing bus");
+                }
+                SlackTerminal.attach(network.getBusView().getBus(busId));
+            } else {
+                throw new PowsyblException("only one of element_id or bus_id must be filled");
+            }
+        }
+    }
+
+    @Override
+    public void addElements(Network network, UpdatingDataframe dataframe) {
+        SlackTerminalDataframeAdder.SlackTerminalSeries series =
+                new SlackTerminalDataframeAdder.SlackTerminalSeries(dataframe);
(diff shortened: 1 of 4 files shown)


java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeAdder.java
```diff
@@ -0,0 +1,72 @@
...
+        Identifiable<?> identifiable = network.getIdentifiable(targetElement);
+        if (identifiable instanceof Injection) {
+            Terminal terminal = ((Injection<?>) identifiable).getTerminal();
+            if (terminal.getVoltageLevel().getTopologyKind() == TopologyKind.BUS_BREAKER && (!targetElement.equals(id))) {
+                throw new UnsupportedOperationException("Cannot set regulated element to " + regulatedElement +
+                        ": not currently supported for bus breaker topologies.");
+            }
+            return terminal;
+        } else {
+            throw new UnsupportedOperationException("Cannot set regulated element to " + regulatedElement +
```
sylvlecl: We have seen yesterday that there is a `SlackTerminal.attach(bus)` :
after all we should also handle the case where the identifiable is a bus here.

sylvlecl (commented):
We should add bus handling, + missing python unit tests

EtienneLt:
> We should add bus handling, + missing python unit tests

Yes but

java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeAdder.java
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        void create(Network network, int row) {
+            String elementId = this.slackElementId.get(row);
+            Identifiable identifiable = network.getIdentifiable(elementId);
+            if (identifiable instanceof Injection) {
+                Terminal terminal = ((Injection) identifiable).getTerminal();
+                SlackTerminal.reset(terminal.getVoltageLevel(), terminal);
+            } else if (identifiable instanceof Bus) {
+                if (network.getBusBreakerView().getBus(elementId) != null) {
```
sylvlecl: Here the identifiable is already a bus, we don't need to get it again, we just need to cast

java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeAdder.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * @author Etienne Lesot <etienne.lesot@rte-france.com>
+ */
+public class SlackTerminalDataframeAdder extends AbstractSimpleAdder {
+
+    private static final List<SeriesMetadata> METADATA = List.of(
+            SeriesMetadata.stringIndex("id"));
```
sylvlecl: `element_id` will probably be best

java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeAdder.java
```diff
@@ -0,0 +1,60 @@
...
+                Terminal terminal = ((Injection) identifiable).getTerminal();
+                SlackTerminal.reset(terminal.getVoltageLevel(), terminal);
+            } else if (identifiable instanceof Bus) {
+                if (network.getBusBreakerView().getBus(elementId) != null) {
+                    SlackTerminal.attach(network.getBusBreakerView().getBus(elementId));
+                } else {
+                    SlackTerminal.attach(network.getBusView().getBus(elementId));
+                }
+            } else {
+                throw new PowsyblException("slack_element_id should be a bus or an injection");
```
sylvlecl: We should formulate it differently so that the user understands that:
- he can use a busbar section (which is in injection in powsybl, but it's not natural for the user)
- the bus argument will only work for "configured buses", i.e buses of voltage levels defined as bus-breaker breaker topologies

java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeAdder.java
sylvlecl: I would propose the following addition, to make it easier for the user to define a slack BUS (bus of the bus view):
to have, as an alternative, and additional column `bus_id`.
Only one of the 2 columns `element_id` and `bus_id` must be filled.

Only when the `bus_id` column is used, we will get the bus from the bus view of the network.

This way the user has 2 choices:
- 1 "high level" choice where he just provides a bus id
- 1 "low level" choice where he provides the id of an element (which could be a bus but only a "configured" bus, voltage levels described in bus-breaker)

pypowsybl/network.py
```diff
@@ -4123,6 +4156,23 @@ def remove_extensions(self, extension_name: str, ids: _Union[str, _List[str]]) -
             ids = [ids]
         _pp.remove_extensions(self._handle, extension_name, ids)
 
+    def get_slack(self) -> _DataFrame:
```
sylvlecl: After more thoughts, I would propose to remove for now those additional network methods:
as such they have no added value compared to the extension.
It could have some added value if, for example, it was also taking care of removing existing slack buses, but let's postpone this kind of addition until the need arises.

java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeProvider.java
```diff
@@ -0,0 +1,48 @@
...
+    private Stream<SlackTerminal> itemsStream(Network network) {
+        return network.getVoltageLevelStream()
+                .map(vl -> (SlackTerminal) vl.getExtension(SlackTerminal.class))
+                .filter(Objects::nonNull);
+    }
+
+    @Override
+    public NetworkDataframeMapper createMapper() {
+        return NetworkDataframeMapperBuilder.ofStream(this::itemsStream)
+                .stringsIndex("id", st -> st.getTerminal().getConnectable().getId())
```
sylvlecl: I would propose propose the same columns as in the adder

java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeProvider.java
```diff
@@ -0,0 +1,48 @@
...
+                .build();
+    }
+
+    @Override
+    public void removeExtensions(Network network, List<String> ids) {
```
sylvlecl, line 48: Maybe it's weird to expect voltage level IDs here, since in the returned dataframe we don't expose it.
So I think we should either:
- add the voltage_level_id as a new column
- accept here "element" ids and/or bus ids 

But in the second case it's not really clear what is expected, so I propose to keep the voltage_level_id solution ...

sylvlecl (approved):
Some propositions to try to clarify things a little from the user point of view (since "terminal"s are not objects that the python user really knows about ... )

java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeAdder.java
```diff
@@ -0,0 +1,69 @@
...
+            if (voltageLevel == null) {
+                throw new PowsyblException("voltage_level_id : " + this.voltageLevelId.get(row) +
+                        " does not correspond to any voltage level");
+            }
+            if (this.elementId != null && this.busId == null) {
+                Identifiable identifiable = network.getIdentifiable(this.elementId.get(row));
+                Terminal terminal = ((Injection) identifiable).getTerminal();
+                SlackTerminal.reset(terminal.getVoltageLevel(), terminal);
+            } else if (this.busId != null && this.elementId == null) {
+                if (network.getBusView().getBus(this.busId.get(row)) == null) {
```
sylvlecl: Note that in general when you have chained calls like this, it's better to keep the result in a variable instead of calling them again after:
- saves some CPU operations
- the intent is more clear: it avoids to have the reader check that the 2 chained calls are exactly the same
- not here, but in a multi-threaded use case it could be weird: another thread could change the result of the getters between the 2 calls

java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeAdder.java
```diff
@@ -0,0 +1,69 @@
...
+                    throw new PowsyblException("bus id : " + this.busId.get(row) + " does not match any existing bus");
+                }
+                SlackTerminal.attach(network.getBusView().getBus(this.busId.get(row)));
+            } else {
+                throw new PowsyblException("only one of element_id or bus_id have to be filled");
```
sylvlecl: typo "has to be filled" (or actually I think "must" is more appropriate :) )

java/src/main/java/com/powsybl/dataframe/network/extensions/SlackTerminalDataframeAdder.java
```diff
@@ -0,0 +1,69 @@
...
+
+        void create(Network network, int row) {
+            VoltageLevel voltageLevel = network.getVoltageLevel(this.voltageLevelId.get(row));
+            if (voltageLevel == null) {
+                throw new PowsyblException("voltage_level_id : " + this.voltageLevelId.get(row) +
+                        " does not correspond to any voltage level");
+            }
+            if (this.elementId != null && this.busId == null) {
+                Identifiable identifiable = network.getIdentifiable(this.elementId.get(row));
+                Terminal terminal = ((Injection) identifiable).getTerminal();
```
sylvlecl: We should check the type of the identifiable (almost as it was done before):
- if it's an injection, do this
- if it's a bus, it means that it's a configured bus of a bus-breaker voltage level --> in this case we attach the slack to it.
  It's different from the "bus_id" column, because in bus_id we expect only buses from the bus view
- else, throw an exception saying that we only handle injections, busbars, or configured buses as an "element_id"

sylvlecl (changes requested):
We also need to revert changes to pybind11 submodule

sylvlecl (approved):

Issue #224, opened 2021-09-17
Loadflow report


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the expected behavior?**

powsybl loadflow API now allows to produce a report of execution through the `Reporter` API.
It will be interesting to provide access to that report to the user.

But in what form ?

- a first implementation can simply print the logs to a string and return it
- but it would be best to get a structured object/list as in java implementation

* **What is the motivation / use case for changing the behavior?**

Access to functional logs for the user:
 - file import report 
 - loadflow iterations etc. for expert users

CBiasuzzi:
How would we change reasonably the existing APIs to include an execution report (maybe optionally)?
The current API to run an ac loadflow is:

```
def run_ac(network: _Network, parameters: Parameters = None, provider : str = 'OpenLoadFlow') 
   -> _List[ComponentResult]:
```
Example usage,  from demo.py:  
```
results = run_ac(network)
for result in results:
    print(result)
```

Among the several viable approaches:

**Option1:** we could return the report as the second element, in a tuple. 

```
def run_ac(network: _Network, parameters: Parameters = None, provider : str = 'OpenLoadFlow') 
   -> _Tuple[_List[ComponentResult], str]:
```

Example usage: 
```
result, report = run_ac(network)
for result in results:
    print(result)
print(report)
```

This breaks the current APIs' return type. Users must know that the result is a tuple and how to get the wanted 'piece' from the tuple (i.e. results = run_ac(network) does not work as expected, anymore)

**Option2**: we could have a dedicated set of APIs, to separate completely those with and without the report, e.g.

```
def run_ac(network: _Network, parameters: Parameters = None, provider : str = 'OpenLoadFlow') 
   -> _List[ComponentResult]:


def run_ac_with_report(network: _Network, parameters: Parameters = None, provider : str = 'OpenLoadFlow') 
   -> _Tuple[_List[ComponentResult], str]:
```

Option2 does not break the original APIs' return type. Example usage: 
```
result = run_ac(network)
result, report = run_ac_with_report(network)
```


**Option3:** we could return either a  _List[ComponentResult] or the previously mentioned tuple
```
def run_ac(network: _Network, parameters: Parameters = None, provider : str = 'OpenLoadFlow', withReport : bool = False) 
   -> _Union[_Tuple[_List[ComponentResult], str], _List[ComponentResult]]:
```

[...]

PR #416, opened 2022-06-06
adds functional execution reports to loadflows, through the powsybl's Reporter API

CBiasuzzi:

#224

**Other information**:


cpp/src/pypowsybl.cpp
```diff
@@ -461,9 +461,9 @@ std::shared_ptr<load_flow_parameters> createLoadFlowParameters() {
 }
 
 LoadFlowComponentResultArray* runLoadFlow(const JavaHandle& network, bool dc, const std::shared_ptr<load_flow_parameters>& parameters,
-                                          const std::string& provider) {
+                                          const std::string& provider, const JavaHandle& reporter) {
```
sylvlecl: I was thinking: maybe instead of the NULL_HANDLE as we had discussed, we could simply use a nullable `JavaHandle*` here ?
I think this would allow the python binding to naturally translate `None` as a null `JavaHandle*`, removing some burden from the python side.

tests/test_loadflow.py
```diff
@@ -131,3 +132,16 @@ def test_provider_parameters():
     assert LoadFlowComponentStatus.MAX_ITERATION_REACHED == result[0].status
     assert 6 == result[0].iteration_count
 
+def test_run_lf_with_report():
+    n = pp.network.create_ieee14()
+    reporter = rp.Reporter()
+    report1 = str(reporter)
```
sylvlecl, line 138: maybe we can have an assertion on the string representation of a just initialized reporter

pypowsybl/report.py
```diff
@@ -0,0 +1,18 @@
+#
+# Copyright (c) 2022, RTE (http://www.rte-france.com)
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+from pypowsybl._pypowsybl import create_reporter_model, print_report, JavaHandle
```
sylvlecl: we need to take care about namespace pollution:
those objects should not be visible for the user in `pypowsybl.report` module.

In order to achieve it several possibilities:
- use a qualified version in the code, maybe aliasing the module as `_pp` to make it readable, sounds reasonable
- import the names with an added underscore
- or as targetted in future versions, move the code to an inner "implementation" module, and only import here what we want to expose

pypowsybl/report.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Reporter:
+    def __init__(self, task_key: str = '', default_name: str = ''):
+        self.reporter_model_handle = create_reporter_model(task_key, default_name)
+
+    def __repr__(self) -> str:
+        return print_report(self.reporter_model_handle)
+
+    @property
+    def reporter_model(self) -> JavaHandle:
```
sylvlecl: We should not expose to the user the java handle, which is an implementation detail --> let's prefix it with an underscore as we do in other places (see network for ex.)

sylvlecl (changes requested):
Requesting minor changes for visibility matters, + maybe possibility to simplify again null handling?

cpp/src/pypowsybl.cpp
CBiasuzzi: You're right, thanks! Removed the python NULL_HANDLE, added a null-checking in the cpp part

tests/test_loadflow.py
CBiasuzzi: done

pypowsybl/report.py
CBiasuzzi: Done (aliased the referred module as _pp)

pypowsybl/report.py
CBiasuzzi: done

sylvlecl (approved):

PR #425, opened 2022-06-13
upgrade powsybl core to version 4.9

EtienneLt:

update

update version of powsybl core, powsybl open loadflow, powsybl single line diagram and powsybl diagram area

**Breaking change**

Security analysis provider `OpenSecurityAnalysis` is renamed to `OpenLoadFlow`.
Sensitivity analysis provider `OpenSensitivityAnalysis` is renamed to `OpenLoadFlow`.


java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
```diff
@@ -330,8 +330,8 @@ static NetworkDataframeMapper batteries() {
                 .doubles("min_q", ifExistsDouble(NetworkDataframes::getMinMaxReactiveLimits, MinMaxReactiveLimits::getMinQ), setMinQ())
                 .doubles("max_q", ifExistsDouble(NetworkDataframes::getMinMaxReactiveLimits, MinMaxReactiveLimits::getMaxQ), setMaxQ())
                 .strings("reactive_limits_kind", NetworkDataframes::getReactiveLimitsKind)
-                .doubles("p0", Battery::getP0, Battery::setP0)
-                .doubles("q0", Battery::getQ0, Battery::setQ0)
+                .doubles("p0", Battery::getTargetP, Battery::setTargetP)
```
geofjamg, line 333: @sylvlecl what do we do in that case? Duplicate the attribute to have both p0 and targetP in the dataframe?

sylvlecl: It would look ugly for the vast majority of users to completely duplicate the columns.

I am more in favor of the breaking change for this one.

If we want to allow smoother transition between versions, maybe we could implement a global compatibility option to use old namings ?

geofjamg: I'm ok with breaking changes if well explained in release notes. I don't know what will be the feeling of users, when their scripts won't work again when updating version, not always easy with large script to fix all columns name.

sylvlecl (approved):

java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
sylvlecl: Note:
this will be done in #426

Issue #440, opened 2022-06-21
Observability extensions


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the current behavior?**

Observability information is not available.

* **What is the expected behavior?**

Exposing the content of observability extensions:
- InjectionObservability
- BranchObservability

PR #450, opened 2022-07-01
add observability extensions

EtienneLt:

#440


sylvlecl (approved):
There is a pybind11 update which I think is not wanted: we should remove it before merging

java/src/main/java/com/powsybl/dataframe/network/extensions/BranchObservabilityDataframeProvider.java
```diff
@@ -0,0 +1,83 @@
...
+    public String getExtensionName() {
+        return BranchObservability.NAME;
+    }
+
+    @Override
+    public NetworkDataframeMapper createMapper() {
+        return NetworkDataframeMapperBuilder.ofStream(this::itemsStream, this::getOrThrow)
+                .stringsIndex("id", branchObservability -> ((Branch) branchObservability.getExtendable()).getId())
+                .booleans("observable", BranchObservability::isObservable)
+                .doubles("p1_standard_deviation", branchObservability -> branchObservability.getQualityP1().getStandardDeviation(),
```
sylvlecl: As identified on another evolution:
we must take care of the cases where quality will be null.

In that case we can return false/nan for redundant/standard dev.

sylvlecl (changes requested):
Actually, we need to fix a possible NPE

sylvlecl:
> There is a pybind11 update which I think is not wanted: we should remove it before merging

Actually the version was the same as in main branch (version 2.9.2) --> it's OK after merging main.

sylvlecl (approved):

Issue #431, opened 2022-06-16
Report: json serialization


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**

Report can only be dumped as text.

* **What is the expected behavior?**

We should support json serialization of reports. The serialization is provided by the java implementation.

* **What is the motivation / use case for changing the behavior?**

The JSON serialization provides a more structured and complete representation than the text representation.

PR #453, opened 2022-07-12
Add JSON serialization to reports

massimo-ferraro:
Signed-off-by: massimo.ferraro <massimo.ferraro@techrain.eu>

#431 

**Other information**:


pypowsybl/report.py
```diff
@@ -16,3 +16,7 @@ def __repr__(self) -> str:
     @property
     def _reporter_model(self) -> _pp.JavaHandle:
         return self._reporter_model_handle
+
+    @property
+    def json(self) -> str:
```
sylvlecl: I would slightly prefer to name it `to_json()` and make it a method, to make it explicit there is some "work" under the hood (also, if we need it later, this will allow for extra arguments such as pretty printing etc).
This will look familiar to pandas users where there are methods suchs as `to_csv()`, ...

tests/test_loadflow.py
```diff
@@ -163,6 +164,8 @@ def test_run_lf_with_report():
     pp.loadflow.run_ac(n, reporter = reporter)
     report2 = str(reporter)
     assert len(report2) > len(report1)
+    json_report = reporter.json
+    json.loads(json_report)
```
sylvlecl, line 169: is it possible to have some basic assertion on the content ? (I don't know what, maybe just that it's not empty ...)

sylvlecl (changes requested):
Very minor requests

pypowsybl/report.py
CBiasuzzi: done: changed to to_json()

tests/test_loadflow.py
CBiasuzzi: done: added an assertion on json_report'content

sylvlecl (approved):

Issue #435, opened 2022-06-16
Security analysis parameters


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**

When running a security analysis, we can only provide loadflow parameters, not security analysis parameters.
Security analysis parameters include loadflow parameters, but also additional parameters, namely the "increased violations parameters", as of powsybl-core 4.9.0.

* **What is the expected behavior?**

We should be able to provide full security analysis parameters:
```python
params = pp.security.Parameters()
# modifying LF parameters
params.loadflow.transformer_voltage_control_on = True
# changing increased violations params
params.increased_violations.flows_proportional_threshold = 0.2

# Defining implementation specific parameters
params.provider_parameters = { 'key' : 'value'}

security_analysi.run_ac(network, params)
```

* **What is the motivation / use case for changing the behavior?**

Being able to fully configure security analysis execution.

sylvlecl:
To ensure compatibility, we should still allow loadflow parameters to be provided, and wrap them in security analysis parameters.

PR #454, opened 2022-07-18
add security analysis parameters

EtienneLt:

#435


EtienneLt:
I have a mistake i don't know how to solve there is a "com.oracle.svm.core.util.VMError$HostedError: Unsupported size: 72" when I try to free or to create the LoadFlowParameters in securtyAnalysisParameters. If anyone knows ?

pypowsybl/security.py
```diff
@@ -45,6 +45,81 @@ def _limit_violation_repr(self: LimitViolation) -> str:
...
+        c_parameters.low_voltage_absolute_threshold = self.low_voltage_absolute_threshold
+        c_parameters.high_voltage_proportional_threshold = self.high_voltage_proportional_threshold
+        c_parameters.high_voltage_absolute_threshold = self.high_voltage_absolute_threshold
+        c_parameters.provider_parameters_keys = list(self.provider_parameters.keys())
+        c_parameters.provider_parameters_values = list(self.provider_parameters.values())
+        return c_parameters
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}(" \
+               f", load_flow_parameters={self.load_flow_parameters!r}" \
```
EtienneLt, line 149: should it be self.load_flow_parameters.__repr__() ?

pypowsybl/security.py
```diff
@@ -45,6 +45,81 @@ def _limit_violation_repr(self: LimitViolation) -> str:
...
+
+    Please note that security analysis providers may not honor all parameters, according to their capabilities.
+    For example, some providers will not be able to simulate the voltage control of shunt compensators, etc.
+    The exact behaviour of some parameters may also depend on your security analysis provider.
+    Please check the documentation of your provider for that information.
+
+    .. currentmodule:: pypowsybl.security
+
+    Args:
+        provider_parameters: Define parameters linked to the security analysis provider
```
sylvlecl, line 102: we'll need to add some documentation for other parameters, for the user

pypowsybl/security.py
sylvlecl: the !r is the equivalent way to do it for format string, so it's good this way :)

cpp/src/bindings.cpp
```diff
@@ -143,6 +143,20 @@ std::shared_ptr<load_flow_parameters> initLoadFlowParameters() {
...
+    std::shared_ptr<security_analysis_parameters> config_params = pypowsybl::createSecurityAnalysisParameters();
+    parameters->flow_proportional_threshold = config_params->flow_proportional_threshold;
+    parameters->low_voltage_proportional_threshold = config_params->low_voltage_proportional_threshold;
+    parameters->low_voltage_absolute_threshold = config_params->low_voltage_absolute_threshold;
+    parameters->high_voltage_proportional_threshold = config_params->high_voltage_proportional_threshold;
+    parameters->high_voltage_absolute_threshold = config_params->high_voltage_absolute_threshold;
+    parameters->provider_parameters_keys_count = config_params->provider_parameters_keys_count;
+    parameters->provider_parameters_values_count = config_params->provider_parameters_values_count;
+    return std::shared_ptr<security_analysis_parameters>(parameters, [](security_analysis_parameters* ptr){
+    });
```
sylvlecl: Here in the destructor, we must delete ptr and everything in loadflow parameters that requires deletion (which means basically "countries_to_balance", see `initLoadFlowParameters`: I think we should call the same method from both places).

We also need to delete `provider_parameters_keys` and `provider_parameters_values`:
I see it's not done in loadflow parameters, it's a memory leak that we need to fix.

java/src/main/java/com/powsybl/python/security/SecurityAnalysisCFunctions.java
```diff
@@ -244,4 +233,30 @@ public static void freeContingencyResultArrayPointer(IsolateThread thread, PyPow
             freeArrayPointer(contingencyResultArrayPtr);
         });
     }
+
+    @CEntryPoint(name = "freeSecurityAnalysisParameters")
+    public static void freeSecurityAnalysisParameters(IsolateThread thread, PyPowsyblApiHeader.SecurityAnalysisParametersPointer securityAnalysisParametersPointer,
+                                              PyPowsyblApiHeader.ExceptionHandlerPointer exceptionHandlerPtr) {
+        doCatch(exceptionHandlerPtr, () -> {
+            // LoadFlowCFunctions.freeLoadFlowParametersPointer(securityAnalysisParametersPointer.getLoadFlowParameters());
```
sylvlecl: We need to also free memory of data within the loadflow parameters (countries to balance), but not the loadflow parameters memory itself, because it will be freed by the following line.

pypowsybl/security.py
```diff
@@ -45,6 +45,81 @@ def _limit_violation_repr(self: LimitViolation) -> str:
...
+            the names of the existing parameters can be found with method ``get_provider_parameters_names``
+    """
+
+    def __init__(self,
+                 load_flow_parameters: pypowsybl.loadflow.Parameters = None,
+                 flow_proportional_threshold: float = None,
```
sylvlecl: We need to make it apparent that those parameters are linked to "increased violations":
for now it's really not clear what they mean.

Maybe the best option is to have a small dedicated class, line on java side ?
This way we can write:
```python
parameters.increased_violations.low_voltage_proportional_threshold = 0.5
```

sylvlecl (changes requested):
Some minor fixes need on memory management + parameters naming/layout

cpp/src/bindings.cpp
```diff
@@ -396,6 +410,54 @@ PYBIND11_MODULE(_pypowsybl, m) {
                 p.provider_parameters_values_count = providerParametersValues.size();
             });
 
+    py::class_<security_analysis_parameters, std::shared_ptr<security_analysis_parameters>>(m, "SecurityAnalysisParameters")
+            .def(py::init(&initSecurityAnalysisParameters))
+            .def_property("load_flow_parameters", [](const security_analysis_parameters& p) {
+                return (load_flow_parameters_struct) p.as_load_flow_parameters;
+            }, [](security_analysis_parameters& p, const load_flow_parameters loadFlowParameters) {
+                p.as_load_flow_parameters = loadFlowParameters;
```
sylvlecl: For memory management, here we need to completely copy the content of the loadflow parameters (countries to balance and provider parameters), like we do for provider parameters below, otherwise when we free memory of LF parameters we risk to free this twice

pypowsybl/security.py
sylvlecl: still a todo :)

pypowsybl/security.py
sylvlecl: still a todo :)

java/src/main/java/com/powsybl/python/security/SecurityAnalysisCUtils.java
```diff
@@ -0,0 +1,67 @@
...
+            } else {
+                parameters.addExtension((Class) ext.getClass(), ext);
+            }
+        });
+        return parameters;
+    }
+
+    public static SecurityAnalysisParameters createSecurityAnalysisParameters(PyPowsyblApiHeader.SecurityAnalysisParametersPointer securityAnalysisParametersPointer) {
+        return createSecurityAnalysisParameters()
+                .setIncreasedViolationsParameters(new SecurityAnalysisParameters.IncreasedViolationsParameters()
```
sylvlecl, line 64: We need to set the loadflow parameters too, I think they are not taken into account by the computation.

pypowsybl/security.py
```diff
@@ -45,6 +45,81 @@ def _limit_violation_repr(self: LimitViolation) -> str:
...
+        if high_voltage_proportional_threshold is not None:
+            self.high_voltage_proportional_threshold = high_voltage_proportional_threshold
+        if high_voltage_absolute_threshold is not None:
+            self.high_voltage_absolute_threshold = high_voltage_absolute_threshold
+        if provider_parameters is not None:
+            self.provider_parameters = provider_parameters
+
+    def _init_with_default_values(self) -> None:
+        default_parameters = _pypowsybl.SecurityAnalysisParameters()
+        self.load_flow_parameters = pypowsybl.loadflow.Parameters()
```
sylvlecl: I think it would be a little clearer to initialize it with the content of `default_parameters.load_flow_parameters`, it will avoid to read twice the config for LF params (once when we initialize SA params, another one here when we initialize LF params).

sylvlecl (changes requested):
See comments: I think there is an issue with the transmission of LF parameters to java when launching computation

sylvlecl (approved):

java/src/main/java/com/powsybl/python/security/SecurityAnalysisCUtils.java
sylvlecl: Actually it was already working, my bad

PR #455, opened 2022-07-21
Wrap network modifications createNewLineOnLine and connectVoltageLevelOnLine

obrix:
Wrap network modifications attachNewLineOnLine and attachVoltageLevelOnLine as new function of network object.

Feature

2 new methods allow to:
- create a "tee point" on an existing line to connect a voltage level
- insert a voltage level by "cutting" an existing line

Those 2 methods are wrappers of network modifications ConnectLineOnLine and ConnectVoltageLevelOnLine from powsybl-core.


pypowsybl/network.py
```diff
@@ -4087,6 +4087,11 @@ def remove_extensions(self, extension_name: str, ids: _Union[str, _List[str]]) -
             ids = [ids]
         _pp.remove_extensions(self._handle, extension_name, ids)
 
+    def attach_new_line_on_line(self, voltage_level_id: str, bbs_or_bus_id: str, line_id: str, new_line_id: str, percent: float = 50.0) -> None:
```
sylvlecl: Not sure we should add those network modifications inside the network class:
at the end, the risk is to have a huge amount of methods in the network, which will make it difficult to use, because the "main", basic methods will be completely drown among those "utility" methods.

I would rather leave them as free functions. It makes sense to leave them in the network module though.

sylvlecl: Otherwise it looks ok to me:
the only thing I wonder is, should we use dataframes like everywhere else to be able to apply multiple changes with one call ?

obrix: I see what you mean. But does it make sense to chain modify the network in the same call ? The result of a modification possibly impacting the "applicability" of the next one (for example splitting a line in two to attach a new line).
Anyway after reflection I can probably make use of dataframe for a single modification and supporting key words arguments to pass all parameters and maybe add the parameters for the new line instead of defaulting to the parameters of the existing line.

obrix: @sylvlecl I just added the support of a dataframe to set the attribute of new line in attach_new_line_on_line. It seems less relevant to do it for attach_voltage_level_on_line (because no adder is used).

tests/test_network.py
```diff
@@ -1494,7 +1494,43 @@ def test_dump_to_string_with_report():
     report2 = str(reporter)
     assert len(report2) >= len(report1)
 
-
+def test_attach_line_on_line():
+    n = pp.network.create_eurostag_tutorial_example1_network()
```
annetill, line 1543: This test is confusing and must be improved. The voltage_level_id and the bbs_or_bus_id should be created in the test, just after creating the network. This new voltage level will have a new load or a new generator.

obrix:
@EtienneLt Is there a central branch for adaptation to 4.10.0 on which we can rebase ? It seems odd to have to do the adaptation work on each PR, do you want to continue the work here or can I continue it ? (There is still compatibility issues here and there but not related to the work of this PR)

EtienneLt:
> @EtienneLt Is there a central branch for adaptation to 4.10.0 on which we can rebase ? It seems odd to have to do the adaptation work on each PR, do you want to continue the work here or can I continue it ? (There is still compatibility issues here and there but not related to the work of this PR)

Yes this is these pr : https://github.com/powsybl/pypowsybl/pull/482 but I was waiting for the fix on sensi 
I'm going to do it

obrix:
@EtienneLt Ok thanks I missed it. I'll wait and then fix the attach_new_line_on_line test after that. I think that will be the last thing to do.

java/src/main/java/com/powsybl/python/network/NetworkCFunctions.java
```diff
@@ -863,4 +869,60 @@ public static void setMinValidationLevel(IsolateThread thread, ObjectHandle netw
...
+            applyIfPresent(df.getDoubles("g2"), 0, adder::setG2);
+            applyIfPresent(df.getDoubles("r"), 0, adder::setR);
+            applyIfPresent(df.getDoubles("x"), 0, adder::setX);
+            CreateLineOnLine createLineOnLine = new CreateLineOnLineBuilder()
+                    .withBusbarSectionOrBusId(bbsIdBusIdStr)
+                    .withLine(line)
+                    .withPercent(percent)
+                    .withLineAdder(adder)
+                    .build();
+            createLineOnLine.apply(network);
```
EtienneLt: constructor of CreateLineOnLine has a more lot attributes 
```java
public CreateLineOnLine(double percent, String voltageLevelId, String bbsOrBusId, String fictitiousVlId, String fictitiousVlName, boolean createFictSubstation, String fictitiousSubstationId, String fictitiousSubstationName, String line1Id, String line1Name, String line2Id, String line2Name, Line line, LineAdder lineAdder)
```
Do I add them ?

java/src/main/java/com/powsybl/python/network/NetworkCFunctions.java
```diff
@@ -863,4 +869,60 @@ public static void setMinValidationLevel(IsolateThread thread, ObjectHandle netw
...
+            String lineIdStr = CTypeUtil.toString(lineId);
+            Network network = ObjectHandles.getGlobal().get(networkHandle);
+            Line line = network.getLine(lineIdStr);
+            ConnectVoltageLevelOnLine modification = new ConnectVoltageLevelOnLineBuilder()
+                    .withLine(line)
+                    .withBusbarSectionOrBusId(bbsIdBusIdStr)
+                    .withVoltageLevelId(voltageLevelIdStr)
+                    .withPercent(percent)
+                    .build();
+            modification.apply(network);
```
EtienneLt: same here
```java
ConnectVoltageLevelOnLine(double percent, String voltageLevelId, String bbsOrBusId, String line1Id, String line1Name, String line2Id, String line2Name, Line line)
```

java/src/main/java/com/powsybl/python/network/NetworkCFunctions.java
obrix: Not sure if that's relevant. It will probably makes for a heavier API on pypowsybl side. @annetill Do you have an opinion on some of these parameters ?

obrix: I think its related to the fictitious substation creation options and the optionnal renaming of the two new lines created by cutting the existing one.

EtienneLt: Anne told me to add them, I'm doing it

cpp/src/bindings.cpp
```diff
@@ -592,4 +592,11 @@ PYBIND11_MODULE(_pypowsybl, m) {
 
     m.def("get_glsk_factors_end_timestamp", &pypowsybl::getInjectionFactorEndTimestamp, "Get glsk end timestamp", py::arg("importer"));
 
+    m.def("create_line_on_line", &pypowsybl::createLineOnLine, "create a new line connected on line", py::arg("network"), py::arg("voltage_level_id"), py::arg("bbs_or_bus_id"), py::arg("line_id"), py::arg("percent"), py::arg("dataframe"), 
+            py::arg("fictitious_voltage_level_id"), py::arg("fictitious_voltage_level_name"), py::arg("fictitious_substation_id"), py::arg("fictitious_substation_name"), py::arg("line1_id"), py::arg("line1_name"), py::arg("line2_id"), py::arg("line2_name"),
+            py::arg("create_fictitious_substation"));
+
+    m.def("connect_voltage_level_on_line", &pypowsybl::connectVoltageLevelOnLine, "connect a voltage level to a line", py::arg("network"), py::arg("voltage_level_id"), py::arg("bbs_or_bus_id"), py::arg("line_id"),
```
annetill: connect a voltage level to a line -> connect a voltage level on a line

cpp/src/bindings.cpp
```diff
@@ -592,4 +592,11 @@ PYBIND11_MODULE(_pypowsybl, m) {
 
     m.def("get_glsk_factors_end_timestamp", &pypowsybl::getInjectionFactorEndTimestamp, "Get glsk end timestamp", py::arg("importer"));
 
+    m.def("create_line_on_line", &pypowsybl::createLineOnLine, "create a new line between another line and a voltage level", py::arg("network"), py::arg("voltage_level_id"), py::arg("bbs_or_bus_id"), py::arg("line_id"), py::arg("percent"), py::arg("dataframe"),
```
annetill: create a new line between a tee point and an existing voltage level.

pypowsybl/network.py
```diff
@@ -4374,3 +4373,51 @@ def get_extensions_names() -> _List[str]:
         the names of the available extensions
     """
     return _pp.get_extensions_names()
+
+def create_line_on_line(n: Network, voltage_level_id: str, bbs_or_bus_id: str, line_id: str, percent: float = 50.0, df_new_line: _DataFrame = None, fictitious_voltage_level_id = '', fictitious_voltage_level_name = '', fictitious_substation_id = '', fictitious_substation_name = '', line1_id = '', line1_name = '', line2_id = '', line2_name = '', create_fictitious_substation = False, **kwargs: _ArrayLike) -> None:
+    """
+    Connect an existing voltage level (in practice a voltage level where we have some loads or generations) to an existing
```
sylvlecl: We should have a short description as the first line of the docstring.

It's used as a short description in some pages in the online documentation (see for ex. in tables here https://pypowsybl.readthedocs.io/en/stable/reference/loadflow.html)

pypowsybl/network.py
```diff
@@ -4374,3 +4373,51 @@ def get_extensions_names() -> _List[str]:
...
+           - **line2_name** : when the initial line is cut, the line segment at side 2 has a given name (optional).
+    """
+    metadata = _pp.get_network_elements_creation_dataframes_metadata(ElementType.LINE)
+    df_new_line = _adapt_df_or_kwargs(metadata[0], df_new_line, **kwargs)
+    c_df = _create_c_dataframe(df_new_line, metadata[0])
+    _pp.create_line_on_line(n._handle, voltage_level_id, bbs_or_bus_id, line_id, percent, c_df, fictitious_voltage_level_id, fictitious_voltage_level_name, fictitious_substation_id, fictitious_substation_name, line1_id, line1_name, line2_id, line2_name, create_fictitious_substation)
+
+def connect_voltage_level_on_line(n: Network, voltage_level_id: str, bbs_or_bus_id: str, line_id: str, percent: float = 50.0 , line1_id: str = '', line1_name: str = '', line2_id: str = '', line2_name: str = '') -> None:
+    """
+    This method cuts an existing line in two lines and connect an existing voltage level between them. The voltage level should
```
sylvlecl, line 4426: Same here we should have a short description as first line

tests/test_network.py
```diff
@@ -1523,11 +1534,68 @@ def test_dump_to_string_with_report():
...
     xml = bat_path.read_text()
-    n = pp.network.load(str(bat_path), reporter = reporter)
+    n = pp.network.load(str(bat_path), reporter=reporter)
     report2 = str(reporter)
     assert len(report2) >= len(report1)
 
 
+def test_create_line_on_line():
+    n = pp.network.create_eurostag_tutorial_example1_network()
+    n.create_substations(pd.DataFrame.from_records(index='id', data=[{'id': 'P3', 'country': 'BE'}]))
```
sylvlecl: note that for creating single elements, it's probably simpler to use the keyword form instead of the dataframe form:
```python
n.create_substations(id='P3', country='BE')
```

Doesn't change anything to the test, that being said :)

tests/test_network.py
```diff
@@ -1523,11 +1534,68 @@ def test_dump_to_string_with_report():
...
+        data=[('GEN3', 4999, -9999.99, 'VLTEST', True, 100, 150, 300, 'VLTEST_0')],
+        columns=['id', 'max_p', 'min_p', 'voltage_level_id',
+                 'voltage_regulator_on', 'target_p', 'target_q', 'target_v', 'bus_id'],
+        index='id'))
+    print(n.get_generators())
+    df = pd.DataFrame(index=['test_line'], columns=['id', 'r', 'x', 'g1', 'g2', 'b1', 'b2'],
+                      data=[['test_line', 5.0, 50.0, 2.0, 3.0, 4.0, 5.0]])
+
+    pp.network.create_line_on_line(n, "VLTEST", "VLTEST_0", "NHV1_NHV2_1", 75.0, df)
+    retrieved_newline = n.get_lines(id=['test_line'])
```
sylvlecl: Note that you could keep only the row with a `.loc['test_line']` to not repeat it further

pypowsybl/network.py
```diff
@@ -4374,3 +4373,51 @@ def get_extensions_names() -> _List[str]:
         the names of the available extensions
     """
     return _pp.get_extensions_names()
+
+def create_line_on_line(n: Network, voltage_level_id: str, bbs_or_bus_id: str, line_id: str, percent: float = 50.0, df_new_line: _DataFrame = None, fictitious_voltage_level_id = '', fictitious_voltage_level_name = '', fictitious_substation_id = '', fictitious_substation_name = '', line1_id = '', line1_name = '', line2_id = '', line2_name = '', create_fictitious_substation = False, **kwargs: _ArrayLike) -> None:
```
sylvlecl: It feels a bit awkward to define the new line characteristic as a dataframe, since we allow only one line. Plus, we don't really know what is the expected content (since it's not completely the same as the arguments for `create_lines`)`.

I would instead specify the arguments explicitly, like for the other arguments (and remove the kwargs).

tests/test_network.py
```diff
@@ -1523,11 +1534,68 @@ def test_dump_to_string_with_report():
...
+    assert retrieved_newline.loc['test_line', "r"] == 5.0
+    assert retrieved_newline.loc['test_line', "x"] == 50.0
+    assert retrieved_newline.loc['test_line', "g1"] == 2.0
+    assert retrieved_newline.loc['test_line', "g2"] == 3.0
+    assert retrieved_newline.loc['test_line', "b1"] == 4.0
+    assert retrieved_newline.loc['test_line', "b2"] == 5.0
+    assert retrieved_newline.loc['test_line', "connected1"]
+    assert retrieved_newline.loc['test_line', "connected2"]
+
```
sylvlecl, line 1575: We could add some tests to check the created voltage level / substation, and mayb check the various options.

tests/test_network.py
```diff
@@ -1523,11 +1534,68 @@ def test_dump_to_string_with_report():
...
+        index='id'))
+    print(n.get_generators())
+    df = pd.DataFrame(index=['test_line'], columns=['id', 'r', 'x', 'g1', 'g2', 'b1', 'b2'],
+                      data=[['test_line', 5.0, 50.0, 2.0, 3.0, 4.0, 5.0]])
+
+    pp.network.create_line_on_line(n, "VLTEST", "VLTEST_0", "NHV1_NHV2_1", 75.0, df)
+    retrieved_newline = n.get_lines(id=['test_line'])
+    assert retrieved_newline.loc['test_line', "r"] == 5.0
+    assert retrieved_newline.loc['test_line', "x"] == 50.0
+    assert retrieved_newline.loc['test_line', "g1"] == 2.0
```
sylvlecl: really a detail but we should try to be consistent using single quotes everywhere.

pypowsybl/network.py
sylvlecl: About argument naming, "percent" is not very explicit, maybe we could use something like "position"?

pypowsybl/network.py
```diff
@@ -4374,3 +4373,51 @@ def get_extensions_names() -> _List[str]:
...
+           - **line1_name** : when the initial line is cut, the line segment at side 1 has a given name (optional).
+           - **line2_id** : when the initial line is cut, the line segment at side 2 has a given ID (optional).
+           - **line2_name** : when the initial line is cut, the line segment at side 2 has a given name (optional).
+    """
+    metadata = _pp.get_network_elements_creation_dataframes_metadata(ElementType.LINE)
+    df_new_line = _adapt_df_or_kwargs(metadata[0], df_new_line, **kwargs)
+    c_df = _create_c_dataframe(df_new_line, metadata[0])
+    _pp.create_line_on_line(n._handle, voltage_level_id, bbs_or_bus_id, line_id, percent, c_df, fictitious_voltage_level_id, fictitious_voltage_level_name, fictitious_substation_id, fictitious_substation_name, line1_id, line1_name, line2_id, line2_name, create_fictitious_substation)
+
+def connect_voltage_level_on_line(n: Network, voltage_level_id: str, bbs_or_bus_id: str, line_id: str, percent: float = 50.0 , line1_id: str = '', line1_name: str = '', line2_id: str = '', line2_name: str = '') -> None:
```
sylvlecl: Same comment about the "percent" argument

sylvlecl: The voltage level argument seems redundant since the busbar already implicitly defines the voltage level ?
About naming, I wonder if "insert_voltage_level_on_line" would not be better, but maybe it's better to follow the powsybl-core naming.

sylvlecl (changes requested):
I think we should replace the dataframe argument for new line in favor of named arguments.
Other comments, about naming etc, are less important.

pypowsybl/network.py
annetill: The name has already been review once on Java side and sorry if you miss the exchange but I am not going to introduce a second breaking change. As I said, the most important part of this pull request is maybe the documentation and examples in the readthedocs (that will follow). The issue linked to redundancy will be fixed on java side for next release.

sylvlecl: Let's keep in mind we can deviate from powsybl java when considered relevant, in pypowsybl: in particular we can avoid the redundancy starting now.
About the naming the question was more "do we deviate from java?", but let's keep the java naming (for the method, not for "percent").

pypowsybl/network.py
EtienneLt: with https://github.com/powsybl/powsybl-core/pull/2292#discussion_r974971126 it will be position_percent

pypowsybl/network.py
```diff
@@ -4374,3 +4373,67 @@ def get_extensions_names() -> _List[str]:
...
+                        fictitious_substation_id: str = '', fictitious_substation_name: str = '') -> None:
+    """
+    Connect an existing voltage level to an existing line through a tee point.
+
+    Connect an existing voltage level (in practice a voltage level where we have some loads or generations) to an existing
+    line through a tee point. This method cuts an existing line in two, creating a fictitious voltage level between them (an a fictitious substation if asked).
+    Then it links an existing voltage level to this fictitious voltage level by creating a new line described in the given dataframe.
+
+    Args:
+           - **n** : the network
```
sylvlecl: We need to follow the google docstring format (removing bullet points and stars)

pypowsybl/network.py
```diff
@@ -4374,3 +4373,67 @@ def get_extensions_names() -> _List[str]:
...
+           - **new_line_b1** :
+           - **new_line_b2** :
+           - **new_line_g1** :
+           - **new_line_g2** :
+           - **line_id** :  the id on of the line on which we want to create a tee point.
+           - **line1_id** : when the initial line is cut, the line segment at side 1 has a given ID (optional).
+           - **line1_name** : when the initial line is cut, the line segment at side 1 has a given name (optional).
+           - **line2_id** : when the initial line is cut, the line segment at side 2 has a given ID (optional).
+           - **line2_name** : when the initial line is cut, the line segment at side 2 has a given name (optional).
+           - **percent** : when the existing line is cut in two lines, percent is equal to the ratio between the parameters of the first line
```
sylvlecl: name to be updated

pypowsybl/network.py
```diff
@@ -4374,3 +4373,67 @@ def get_extensions_names() -> _List[str]:
...
+
+def create_line_on_line(n: Network, bbs_or_bus_id: str, new_line_id: str, new_line_r: float, new_line_x: float, new_line_b1: float,
+                        new_line_b2: float, new_line_g1: float, new_line_g2: float, line_id: str, line1_id: str = '', line1_name: str = '', line2_id: str = '', line2_name: str = '',
+                        position_percent: float = 50.0, create_fictitious_substation: bool = False, fictitious_voltage_level_id: str = '', fictitious_voltage_level_name: str = '',
+                        fictitious_substation_id: str = '', fictitious_substation_name: str = '') -> None:
+    """
+    Connect an existing voltage level to an existing line through a tee point.
+
+    Connect an existing voltage level (in practice a voltage level where we have some loads or generations) to an existing
+    line through a tee point. This method cuts an existing line in two, creating a fictitious voltage level between them (an a fictitious substation if asked).
```
sylvlecl: let's try to keep reasonable line length in docstrings so that it's readable in code

java/src/main/java/com/powsybl/python/network/NetworkCFunctions.java
```diff
@@ -864,4 +868,120 @@ public static void setMinValidationLevel(IsolateThread thread, ObjectHandle netw
...
+                                                 CCharPointer line1Id,
+                                                 CCharPointer line1Name,
+                                                 CCharPointer line2Id,
+                                                 CCharPointer line2Name,
+                                                 double positionPercent,
+                                                 ExceptionHandlerPointer exceptionHandlerPtr) {
+        doCatch(exceptionHandlerPtr, () -> {
+            String bbsIdBusIdStr = CTypeUtil.toString(bbsIdBusId);
+            String lineIdStr = CTypeUtil.toString(lineId);
+            String line1IdStr = CTypeUtil.toString(line1Id).equals("") ? null : CTypeUtil.toString(line1Id);
```
sylvlecl: We could have an util method for this very common case of transforming empty to null

sylvlecl (approved):

PR #462, opened 2022-07-27
Build on macos 11 (upgrade from 10.15)

sylvlecl:

Technical update

Support for macos 10.15 in github actions will be removed on 30/08/2022:
https://github.com/actions/virtual-environments/issues/5583

In consequence, we now use macos-11. We drop support for macos 10.15.

**Technical note:**
`pip` does not seem to support well the fact that macos 10.16 and 11.0 equivalent, therefore we target 10.16 in the build process in order to ensure maximum compatibility.

### DIFF
--- a/.github/workflows/dev-ci.yml
+++ b/.github/workflows/dev-ci.yml
@@ -95,7 +95,7 @@ jobs:
         config:
           - {
             name: darwin,
-            os: macos-10.15,
+            os: macos-11,
           }
           - {
             name: windows,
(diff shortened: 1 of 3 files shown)


geofjamg (approved):

sylvlecl:
Outcome of some tests:
- Build on macos 11 with MACOSX_DEPLOYMENT_TARGET=10.15 produces a wheel which indeed works on macos 10.15.
  It does not matter if we drop support of 10.15, of course, but good to know
- Building with MACOSX_DEPLOYMENT_TARGET=10.16 (which is a synonymous of 11.0) allows to install and run on macos 11.0 independently of the value of SYSTEM_VERSION_COMPAT (env variable which controls the OS version "seen" by scripting languages)

So finally, to support 11.0 and higher, the best option seems to use MACOSX_DEPLOYMENT_TARGET=10.16.

Issue #463, opened 2022-07-27
Make powsybl Network more informative


AnkurArohi:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the current behavior?**
Currently the info recieved is meta and does not contain info about network elements

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

* **What is the expected behavior?**
The network should be more informative letting the user know how many elements are present in the network.

* **What is the motivation / use case for changing the behavior?**

Make the user understand the network in a summarized manner

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #464, opened 2022-07-27
Make powsybl network informative

AnkurArohi:
- [x ] The commit message follows our guidelines

https://github.com/powsybl/pypowsybl/issues/463

Feature

https://github.com/powsybl/pypowsybl/issues/463

Network creation is more informative with the description of number of elements of a particular type

**Other information**:

### DIFF
--- a/pypowsybl/network.py
+++ b/pypowsybl/network.py
@@ -186,6 +186,12 @@ class Network:  # pylint: disable=too-many-public-methods
         self._source_format = att.source_format
         self._forecast_distance = _datetime.timedelta(minutes=att.forecast_distance)
         self._case_date = _datetime.datetime.utcfromtimestamp(att.case_date)
+        self.elements = {"Bus":self.get_buses().shape[0],"Busbarsections":self.get_busbar_sections().shape[0],
+                         "Gens" : self.get_generators().shape[0],"Load":self.get_loads().shape[0],
+                         "Lines":self.get_lines().shape[0],
+                         "Trafo2w":self.get_2_windings_transformers().shape[0],
+                         "Trafo3w":self.get_3_windings_transformers().shape[0],
+                         "Dangling-Lines":self.get_dangling_lines().shape[0]}
 
     @property
     def id(self) -> str:
@@ -200,6 +206,13 @@ class Network:  # pylint: disable=too-many-public-methods
         Name of this network
         """
         return self._name
+
+    @property
+    def objects(self) -> str:
+        """
+        Elements of this Network
+        """
+        return f"This powsybl network contains :{self.elements}"
 
     @property
     def source_format(self) -> str:
@@ -224,7 +237,7 @@ class Network:  # pylint: disable=too-many-public-methods
 
     def __str__(self) -> str:
         return f'Network(id={self.id}, name={self.name}, case_date={self.case_date}, ' \
-               f'forecast_distance={self.forecast_distance}, source_format={self.source_format})'
+               f'forecast_distance={self.forecast_distance}, source_format={self.source_format}, elements={self.objects})'
 
     def __repr__(self) -> str:
         return str(self)
--- a/tests/test_network.py
+++ b/tests/test_network.py
@@ -1,3 +1,4 @@
+# pylint: : disable-all
 #
 # Copyright (c) 2020-2022, RTE (http://www.rte-france.com)
 # This Source Code Form is subject to the terms of the Mozilla Public
--- a/tests/test_network.py
+++ b/tests/test_network.py
@@ -5,30 +5,27 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 #
 import copy
-import re
-import unittest
 import datetime
+import pathlib
+import re
+import tempfile
+import unittest
 from os.path import exists
 
-import pandas as pd
-import pytest
-from numpy import NaN
-import numpy as np
-
-import pypowsybl as pp
-import pathlib
 import matplotlib.pyplot as plt
 import networkx as nx
-
+import numpy as np
+import pandas as pd
+import pypowsybl as pp
+import pypowsybl.report as rp
+import pytest
+import util
+from numpy import NaN
 from pypowsybl.network import ValidationLevel
 
-import util
-import tempfile
-
-import pypowsybl.report as rp
 
 TEST_DIR = pathlib.Path(__file__).parent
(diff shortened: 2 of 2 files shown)


pypowsybl/network.py
```diff
@@ -201,6 +207,13 @@ def name(self) -> str:
         """
         return self._name
 
+    @property
+    def objects(self) -> str:
```
sylvlecl, line 211: I don't think adding this property is useful, it's almost a duplicate of the string representation

tests/test_network.py
```diff
@@ -1,34 +1,32 @@
...
 from pypowsybl.network import ValidationLevel
 
-import util
-import tempfile
-
-import pypowsybl.report as rp
 
 TEST_DIR = pathlib.Path(__file__).parent
-DATA_DIR = TEST_DIR.parent / 'data'
+DATA_DIR = TEST_DIR.parent / "data"
```
sylvlecl, line 29: The PR should only contain changes related to your feature, not changes in the whole file.
Also, we need to take care to respect the existing conventions in new code (single quotes instead of double quotes for example here).

pypowsybl/network.py
```diff
@@ -186,6 +186,12 @@ def __init__(self, handle: _pp.JavaHandle):
         self._source_format = att.source_format
         self._forecast_distance = _datetime.timedelta(minutes=att.forecast_distance)
         self._case_date = _datetime.datetime.utcfromtimestamp(att.case_date)
+        self.elements = {"Bus":self.get_buses().shape[0],"Busbarsections":self.get_busbar_sections().shape[0],
```
sylvlecl, line 189: That information will become wrong as soon as the user changes something in the network:
it's not the right place to do it

pypowsybl/network.py
```diff
@@ -224,7 +237,7 @@ def forecast_distance(self) -> _datetime.timedelta:
 
     def __str__(self) -> str:
         return f'Network(id={self.id}, name={self.name}, case_date={self.case_date}, ' \
-               f'forecast_distance={self.forecast_distance}, source_format={self.source_format})'
+               f'forecast_distance={self.forecast_distance}, source_format={self.source_format}, elements={self.objects})'
```
sylvlecl, line 240: If you want to display up-to-date information, you need to retrieve the number of elements from here:
if we want to do this, we should implement a specific call to the java backend, otherwise it will greatly hurt the performance of `str` to gather all the dataframes (may grow to tenths of seconds on a large network).

pypowsybl/network.py
```diff
@@ -186,6 +186,12 @@ def __init__(self, handle: _pp.JavaHandle):
         self._source_format = att.source_format
         self._forecast_distance = _datetime.timedelta(minutes=att.forecast_distance)
         self._case_date = _datetime.datetime.utcfromtimestamp(att.case_date)
+        self.elements = {"Bus":self.get_buses().shape[0],"Busbarsections":self.get_busbar_sections().shape[0],
+                         "Gens" : self.get_generators().shape[0],"Load":self.get_loads().shape[0],
+                         "Lines":self.get_lines().shape[0],
+                         "Trafo2w":self.get_2_windings_transformers().shape[0],
```
sylvlecl, line 192: We need to be as much as possible consistent in the naming of things accross the lib:
- 2 windings transformers instead of trafo2w
- same for 3 windings
- generators instead of "gens"

sylvlecl (changes requested):
Thanks for your PR @AnkurArohi :
see the review, if we want to add such summary information in the string representation of the network, we will need to fix the implementation to have up to date information, and also change the retrieval of that information with a specific call to java, for performance reasons.

pypowsybl/network.py
AnkurArohi: @sylvlecl 
Thank you for the review, I do understand your point but the motive here is to inform the user about the network elements.
The old string representation had just the meta info.
The issue of not updated elements is right , that has to be dynamically done
How do you suggest that?

pypowsybl/network.py
AnkurArohi: Yes i want to dynamically update this info.
How do you suggest we calll java backend

pypowsybl/network.py
AnkurArohi: yes i will correct this

tests/test_network.py
AnkurArohi: I tried to make the file consistent to the pylint norm

pypowsybl/network.py
AnkurArohi: can you help me a bit in understanding where should i call java backend

pypowsybl/network.py
AnkurArohi: @sylvlecl I am trying to figure out the right place to do it , can you help me pls

AnkurArohi (commented):
@sylvlecl 
Thanks for your guidance i appreciate your points, actually the dynamic updated network was not thought to be displayed with elements , the initial thought was to only display the elements from the CGMES Import as i am currently working with that , obviously it makes more sense to dynamically display the elements once the user updates / add new / drops old elements i am facing issues to understand where should i call java backend for getting updated values, perhaps you could help me.

The other issue with the consistency will be corrected and the pylint changes to the test were made to satisfy the pylint norm.
Thanks

pypowsybl/network.py
sylvlecl: I think we could retrieve the information where we actually use it, which would mean in \_\_str__.

That's why it needs to be a pretty "light" operation, we don't want a call to `str(network)` to take long.
For this to be efficient, it must not retrieve the whole dataframes (this implies some quite heavy data transfer between the java, object, representation and the C structure underlying the dataframe), but it should instead only get the size information from the java network. So this means adding methods in the C layer and java layer (NetworkCFunctions.java) for this, for example to get the count of elements for a given element type.

tests/test_network.py
sylvlecl: Understood but such a large change should go into another PR to separate the concerns.

About the particular question of quotes, pylint does not dictate anything between the 2, but consistency matters:
in pypowsybl and in many codebases, we have chosen single quote for simple strings (and 3xdouble quotes for docstrings and multi-line strings).

AnkurArohi: ok, i have noted my todos:

- Maintain consistency for names of elements (Easy)
- Look to add java method  to make number of elements easily and dynamically updateable (Difficult)
- Use single quotes (ok)

Issue #478, opened 2022-08-17
Short circuits related extensions


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature

* **What is the expected behavior?**

We should expose the following IIDM extensions:

- IdentifiableShortCircuit (provides short circuit current limits)
- GeneratorShortCircuit (provides data for short circuit simulation)

PR #479, opened 2022-08-18
add extensions identifiableShortCircuit and generatorShortCircuit

EtienneLt:

#478


java/src/main/java/com/powsybl/dataframe/network/extensions/IdentifiableShortCircuitDataframeAdder.java
```diff
@@ -0,0 +1,72 @@
...
+
+        IdentifiableShortCircuitSeries(UpdatingDataframe dataframe) {
+            this.id = dataframe.getStrings("id");
+            this.ipMin = dataframe.getDoubles("ip_min");
+            this.ipMax = dataframe.getDoubles("ip_max");
+        }
+
+        void create(Network network, int row) {
+            String id = this.id.get(row);
+            Generator g = network.getGenerator(id);
```
sylvlecl: Here we should support any identifiable :
actually, our usual use case is to have limits on voltage levels.

java/src/main/java/com/powsybl/dataframe/network/extensions/IdentifiableShortCircuitDataframeAdder.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.List;
+
+/**
+ * @author Etienne Lesot <etienne.lesot@rte-france.com>
+ */
+public class IdentifiableShortCircuitDataframeAdder extends AbstractSimpleAdder {
+
+    private static final List<SeriesMetadata> METADATA = List.of(
+            SeriesMetadata.stringIndex("id"),
+            SeriesMetadata.strings("element_type"),
```
sylvlecl, line 30: Not necessary for creation

tests/test_network_extensions.py
```diff
@@ -292,3 +292,27 @@ def test_branch_observability():
...
     n.remove_extensions(extension_name, [element_id])
     assert n.get_extensions(extension_name).empty
+
+def test_identifiable_short_circuit():
+    n = pn.create_four_substations_node_breaker_network()
+    extension_name = 'identifiableShortCircuit'
+    extensions = n.get_extensions(extension_name)
+    assert extensions.empty
+    element_id = 'GH1'
+    n.create_extensions(extension_name, id=element_id, ip_min=3.2, ip_max=5.1)
```
sylvlecl, line 369: would be better to test on a voltage level, which is the typical use case

sylvlecl (changes requested):
We need to support creating limits on all identifiables, not only generators.

sylvlecl (approved):

PR #484, opened 2022-08-30
Network modification: implements createFeederBay

colineplqt:

No

Feature

It is possible to create a new injection, connect it, and create the associated topology via CreateFeederBay


java/src/main/java/com/powsybl/python/network/NetworkModificationCFunctions.java
```diff
@@ -0,0 +1,108 @@
...
+            if (reporter == null) {
+                modification.apply(network, throwException, Reporter.NO_OP);
+            } else {
+                modification.apply(network, throwException, reporter);
+            }
+        });
+    }
+
+    public static InjectionAdder<?> createAdder(PyPowsyblApiHeader.ElementType elementType, VoltageLevel voltageLevel, UpdatingDataframe df) {
+        if (elementType == PyPowsyblApiHeader.ElementType.LOAD) {
```
colineplqt: I am really not sure about this piece of code. The code is already present in the different DataframeAdder create method, but contains the call to the add() function that I don't need. Maybe it would be possible to enhance that.

EtienneLt: yes I think it would be better to add a method to create the adder in the DataframeAdder and the create method will just call these one and call create()

cpp/src/bindings.cpp
```diff
@@ -592,4 +592,6 @@ PYBIND11_MODULE(_pypowsybl, m) {
 
     m.def("get_glsk_factors_end_timestamp", &pypowsybl::getInjectionFactorEndTimestamp, "Get glsk end timestamp", py::arg("importer"));
 
+    m.def("create_feeder_bay", &pypowsybl::createFeederBay, "Create feeder bay", py::arg("network"), py::arg("busbarSectionId"), py::arg("injectionPositionOrder"), py::arg("injectionDirection"), py::arg("throwException"), py::arg("reporter"), py::arg("dataframe"), py::arg("elementType"));
```
EtienneLt: for python conventions busbar_section_id, injection_position_order ... would be better

pypowsybl/network_modification.py
```diff
@@ -0,0 +1,23 @@
...
+import pypowsybl._pypowsybl as _pp
+from numpy.typing import ArrayLike as _ArrayLike
+from pandas import DataFrame as _DataFrame
+
+from pypowsybl.network import Network
+from pypowsybl.report import Reporter as _Reporter
+from pypowsybl.utils.dataframes import _adapt_df_or_kwargs, _create_c_dataframe
+
+
+def create_new_feeder_bay(n: Network, busbar_section_id: str, injection_position_order: int,
```
EtienneLt: would it be better in network module ?

pypowsybl/network_modification.py
```diff
@@ -0,0 +1,23 @@
...
+                          df: _DataFrame, element_type: _pp.ElementType, injection_direction: str = "BOTTOM", throw_exception: bool = False,
+                          reporter: _Reporter = None, **kwargs: _ArrayLike) -> None:
+    metadata = _pp.get_network_elements_creation_dataframes_metadata(element_type)
+    df = _adapt_df_or_kwargs(metadata[0], df, **kwargs)
+    c_df = _create_c_dataframe(df, metadata[0])
```
EtienneLt: can network methods be reused there ?

pypowsybl/network_modification.py
```diff
@@ -0,0 +1,23 @@
...
+from pandas import DataFrame as _DataFrame
+
+from pypowsybl.network import Network
+from pypowsybl.report import Reporter as _Reporter
+from pypowsybl.utils.dataframes import _adapt_df_or_kwargs, _create_c_dataframe
+
+
+def create_new_feeder_bay(n: Network, busbar_section_id: str, injection_position_order: int,
+                          df: _DataFrame, element_type: _pp.ElementType, injection_direction: str = "BOTTOM", throw_exception: bool = False,
+                          reporter: _Reporter = None, **kwargs: _ArrayLike) -> None:
```
EtienneLt: if you could add some docstring it would be great like on the network methods (a getter for example)

pypowsybl/network_modification.py
colineplqt: I thought about putting it in this module because the network module is pretty big and there will be a lot of functions added for modifications. But I can move it.

java/src/main/java/com/powsybl/dataframe/network/adders/AbstractSimpleAdder.java
```diff
@@ -30,4 +32,8 @@ public void addElements(Network network, List<UpdatingDataframe> dataframes) {
     }
 
     public abstract void addElements(Network network, UpdatingDataframe dataframe);
+
+    public void addElementsWithBay(Network network, UpdatingDataframe dataframe, String busbarSectionId, int injectionPositionOrder,
+                                   ConnectablePosition.Direction direction, boolean throwException, Reporter reporter) {
+    }
```
colineplqt, line 92: I added this method to make it easier, but actually it doesn't apply to every elements. I don't know how to make it look better...

EtienneLt: Maybe you can add the attributes busbarSectionId, injectionPositionOrder and direction in the dataframe, what do you think ?

pypowsybl/network.py
```diff
@@ -4124,6 +4124,228 @@ def remove_extensions(self, extension_name: str, ids: _Union[str, _List[str]]) -
...
+            reporter: the reporter to be used to create an execution report, default is None (no report).
+            kwargs: attributes of the injection as keyword arguments.
+
+        """
+        return self.create_feeder_bay(busbar_section_id, generator_position_order, df, ElementType.GENERATOR, generator_direction, throw_exception, None if reporter is None else reporter._reporter_model, kwargs)
+
+    def create_dangling_line_bay(self, busbar_section_id: str, dangling_line_position_order: int, df: _DataFrame,
+                                 dangling_line_direction: str = "BOTTOM", throw_exception: bool = False,
+                                 reporter: _Reporter = None, **kwargs: _ArrayLike) -> None:
+        """
```
colineplqt: @annetill, here, it would be possible to define a different default direction per injection. Do you think it would be worth it?

pypowsybl/network.py
```diff
@@ -4124,6 +4124,228 @@ def remove_extensions(self, extension_name: str, ids: _Union[str, _List[str]]) -
...
+        to every parallel busbar section with an open disconnector.
+
+        Args:
+            busbar_section_id: id of the busbar section to which the injection will be connected with a closed
+            disconnector.
+            load_position_order: the order of the load, will fill the ConnectablePosition extension
+            df: the dataframe containing the attributes of the load.
+            load_direction: the direction of the load, will fill the ConnectablePosition extension, default
+            is BOTTOM.
+            throw_exception: whether the calculation should throw exceptions. In any case, errors will be logged.
```
sylvlecl: Not convinced that this option is really useful: the user can catch the exception if he wants anyway ?

java/src/main/java/com/powsybl/python/network/NetworkModificationCFunctions.java
sylvlecl: I guess that method can be removed now, it seems it has been refactored?

pypowsybl/network.py
```diff
@@ -4124,6 +4124,228 @@ def remove_extensions(self, extension_name: str, ids: _Union[str, _List[str]]) -
...
+        """
+        Creates a load, connects it to the network and creates the topology. The voltage level containing the
+        busbar section should be described in node/breaker topology. The load is connected to the busbar with a
+        breaker and a closed disconnector. If the network has position extensions, the load will also be connected
+        to every parallel busbar section with an open disconnector.
+
+        Args:
+            busbar_section_id: id of the busbar section to which the injection will be connected with a closed
+            disconnector.
+            load_position_order: the order of the load, will fill the ConnectablePosition extension
```
sylvlecl: I think this should be optional, since on simple networks there is no position, and therefore the extension will not be created.

docs/user_guide/network.rst
```diff
@@ -330,4 +330,32 @@ You can now run a loadflow to check our network actually works !
...
+
+.. testcode::
+
+    df = pd.DataFrame(index=["new_load"], columns=["id", "p0", "q0"], data=[["new_load", 100.0, 50.0]])
+
+You can then create the load and connect it to a busbar section:
+
+.. testcode::
+
+    network.create_feeder_bay(busbar_section_id="S1VL2_BBS1", injection_position_order=0, df=df, element_type=ElementType.LOAD)
```
sylvlecl: Since we have dedicated methods for each type of element, I would rather use them in the documentation

pypowsybl/network.py
```diff
@@ -4124,6 +4124,228 @@ def remove_extensions(self, extension_name: str, ids: _Union[str, _List[str]]) -
             ids = [ids]
         _pp.remove_extensions(self._handle, extension_name, ids)
 
+    def create_load_bay(self, busbar_section_id: str, load_position_order: int, df: _DataFrame,
```
sylvlecl: We should be homogeneous: either allow the creation of only one element and use explicit named arguments, or allow to create multiple elements and use dataframe arguments.

I would rather go for the second solution, to allow the creation of multiple elements, as is already possible when we create them without feeder bay. In that case, the busbar section etc. would be columns of the dataframe.

java/src/main/java/com/powsybl/dataframe/network/adders/BatteryDataframeAdder.java
```diff
@@ -84,4 +93,24 @@ public void addElements(Network network, UpdatingDataframe dataframe) {
...
+        for (int row = 0; row < dataframe.getRowCount(); row++) {
+            BatteryAdder adder = series.createAdder(network, row);
+            CreateFeederBay modification = new CreateFeederBayBuilder()
+                    .withInjectionAdder(adder)
+                    .withBbsId(busbarSectionId)
+                    .withInjectionPositionOrder(injectionPositionOrder)
+                    .withInjectionDirection(direction)
+                    .build();
+            if (reporter == null) {
+                modification.apply(network, throwException, Reporter.NO_OP);
```
sylvlecl: Note that it's shorter to have only one method call :)
```java
modification.apply(network, throwException, reporter == null ? Reporter.NO_OP : reporter);
```

java/src/main/java/com/powsybl/dataframe/network/adders/AbstractSimpleAdder.java
sylvlecl: In addition of the dataframe stuff, we can probably share more code since the only difference between adding with bays or not is the "addition" mechanism: we could have something like
```java
@FunctionalInterface
interface AdditionStrategy {
    void add(UpdatingDataframe dataframe, InjectionAdder<?> adder);  //adds with or without bay
}

//This is still implemented in subclasses
public abstract void addElements(Network network, UpdatingDataframe dataframe, AdditionStrategy addition);

public void addElementsWithBay(Network network, UpdatingDataframe dataframe) {
    addElements(network, dataframe, // method which adds with bays);
}

public void addElements(Network network, UpdatingDataframe dataframe) {
    addElements(network, dataframe, (dataframe, adder) -> adder.add()); // (well actually there is no .add, you need to do like in CreateFeederBay....)
}
```

pypowsybl/network.py
```diff
@@ -4124,6 +4124,228 @@ def remove_extensions(self, extension_name: str, ids: _Union[str, _List[str]]) -
...
+                        load_direction: str = "BOTTOM", throw_exception: bool = False,
+                        reporter: _Reporter = None, **kwargs: _ArrayLike) -> None:
+        """
+        Creates a load, connects it to the network and creates the topology. The voltage level containing the
```
sylvlecl: In all docstrings:
we must have a short description as the first line, because it's used in the HTML doc as a summary where there is not much space (in table cells).

pypowsybl/network.py
sylvlecl: To be homogeneous with other network modifications, I would move those methods outside the `Network` class.

Alternatively, we could imagine giving the possibility of creating bays in the existing creation methods (`create_loads` etc).

@annetill @geofjamg : any opinion about this last proposition ?

sylvlecl (changes requested):
Mainly:
- Proposition to use dataframes to be able to create multiple elements with bays at once.
- Some design proposition to re-use more code.

pypowsybl/network.py
colineplqt: For now it is not possible to use CreateFeederBay without the order. But it will be done in a future version and then it will be possible to change it here too.

java/src/main/java/com/powsybl/dataframe/network/adders/AbstractSimpleAdder.java
colineplqt: Actually, the dataframe with the injection attributes are supposed to represent the InjectionAdder. So I think it's better like that. To be discussed is if we want to put the attributes necessary to computation like the position, direction ect. in another dataframe.

pypowsybl/network.py
colineplqt: The goal of this boolean is to allow the use of CreateFeederBay with no failure. Is there another option in Python that would replace it?

sylvlecl: I mean that the user could simply do:
```python
try:
    pp.create_feeder_bay(...)
except pp.PyPowsyblError:
    # log or something else
```

But actually the option can be useful if we allow to create multiple feeders with one call (with a dataframe):
this way it will not stop on first failure.

Let's keep it and rename it raise_exception ? (in python they say raise instead of throw :) )

pypowsybl/network.py
sylvlecl: Ok

java/src/main/java/com/powsybl/dataframe/network/adders/AbstractSimpleAdder.java
colineplqt: About the functional interface, actually it's not that easy because of the boolean throwException and the reporter....

java/src/test/java/com/powsybl/python/network/SingleLineDiagramUtilTest.java
```diff
@@ -42,8 +37,8 @@ void test() throws IOException {
         Network network = FourSubstationsNodeBreakerFactory.create();
         try (StringWriter writer = new StringWriter()) {
             SingleLineDiagramUtil.writeSvg(network, "S1VL1", writer);
-            assertEquals(TestUtil.normalizeLineSeparator(new String(ByteStreams.toByteArray(Objects.requireNonNull(SingleLineDiagramUtil.class.getResourceAsStream("/sld.xml"))), StandardCharsets.UTF_8)),
-                         fixSvg(TestUtil.normalizeLineSeparator(writer.toString())));
+//            assertEquals(TestUtil.normalizeLineSeparator(new String(ByteStreams.toByteArray(Objects.requireNonNull(SingleLineDiagramUtil.class.getResourceAsStream("/sld.xml"))), StandardCharsets.UTF_8)),
```
annetill: These changes must maybe be reverted?

pypowsybl/network.py
```diff
@@ -4438,3 +4430,402 @@ def connect_voltage_level_on_line(network: Network, bbs_or_bus_id: str, line_id:
...
+        ConnectablePosition extension, default is BOTTOM.
+
+
+    """
+    if df is not None:
+        df['voltage_level_id'] = df.apply(lambda row: network.get_busbar_sections(attributes=['voltage_level_id']).loc[row['busbar_section_id']].get(0), axis=1)
+    return create_feeder_bay(network, [df], ElementType.VSC_CONVERTER_STATION, raise_exception, reporter)
+
+
+def create_feeder_bay(network: Network, dfs: _List[_Optional[_DataFrame]], element_type: _pp.ElementType, raise_exception: bool, reporter: _Optional[_Reporter]) -> None:
```
sylvlecl: We are already not consistent today between all those methods, but I would propose to make this one protected (leading underscore)

tests/test_network.py
```diff
@@ -1600,3 +1600,16 @@ def test_connect_voltage_level_on_line():
     assert retrieved_splittedline2.loc['NHV1_NHV2_1_2', "voltage_level1_id"] == "N_VL"
     assert retrieved_splittedline2.loc['NHV1_NHV2_1_2', "voltage_level2_id"] == "VLHV2"
     assert retrieved_splittedline2.loc['NHV1_NHV2_1_2', "r"] == 0.75
+
+
+def test_add_load_bay():
```
sylvlecl, line 1606: unfortunately I think sonar is right that we should add some tests for other types of elements as well ^^

tests/test_network.py
```diff
@@ -1600,3 +1600,16 @@ def test_connect_voltage_level_on_line():
...
+    n = pp.network.create_four_substations_node_breaker_network()
+    df = pd.DataFrame(index=["new_load"], columns=["id", "p0", "q0", "busbar_section_id", "position_order"], data=[["new_load", 10.0, 3.0, "S1VL1_BBS", 0]])
+    pp.network.create_load_bay(network=n, df=df, raise_exception=True)
```
sylvlecl, line 1609: It would be nice to have a test with the keyword approach too (not dataframe)

docs/user_guide/network.rst
```diff
@@ -330,4 +330,49 @@ You can now run a loadflow to check our network actually works !
...
+.. image:: ../_static/images/four_substation_node_breaker_s1vl2.svg
+
+Then you need to create a dataframe with the attributes of the injection that you want to add to the network. Let's say
+that we want to add a load, with the id "new_load" on the voltage level S1VL2. The load will have a p0 of 100 and a
+q0 of 50. It is also necessary to put in the dataframe the id of the busbar section where we want to put the load and
+the order position.
+
+.. testcode::
+
+    df = pd.DataFrame(index=["new_load"], columns=["id", "p0", "q0", "voltage_level_id", "busbar_section_id", "position_order"], data=[["new_load", 100.0, 50.0, "S1VL2", "S1VL2_BBS1", 10]])
```
sylvlecl: I would suggest to use the keyword approach for the user guide (more user friendly).
In any case we can also refine the doc later.

java/src/main/java/com/powsybl/dataframe/network/adders/NetworkElementAdder.java
```diff
@@ -27,5 +28,15 @@ public interface NetworkElementAdder {
      * The first dataframe is considered the "primary" dataframe, other dataframes
      * can provide additional data (think steps for the tap changers).
      */
-    void addElements(Network network, List<UpdatingDataframe> dataframes);
+    default void addElements(Network network, List<UpdatingDataframe> dataframes) {
```
sylvlecl: I don't think this needs a default implementation

java/src/main/java/com/powsybl/dataframe/network/adders/NetworkElementAdder.java
```diff
@@ -27,5 +28,15 @@ public interface NetworkElementAdder {
...
+    }
+
+    default void addElementsWithBay(Network network, List<UpdatingDataframe> dataframe, boolean throwException, Reporter reporter) {
+    }
+
+    default void addElements(Network network, UpdatingDataframe dataframe) {
+    }
+
+    default void addElements(Network network, UpdatingDataframe dataframe, AbstractSimpleAdder.AdditionStrategy addition, boolean throwException, Reporter reporter) {
```
sylvlecl: This is an implementation detail: it does not need to go in the interface

java/src/main/java/com/powsybl/dataframe/network/adders/NetworkElementAdder.java
```diff
@@ -27,5 +28,15 @@ public interface NetworkElementAdder {
...
      * can provide additional data (think steps for the tap changers).
      */
-    void addElements(Network network, List<UpdatingDataframe> dataframes);
+    default void addElements(Network network, List<UpdatingDataframe> dataframes) {
+    }
+
+    default void addElementsWithBay(Network network, List<UpdatingDataframe> dataframe, boolean throwException, Reporter reporter) {
+    }
+
+    default void addElements(Network network, UpdatingDataframe dataframe) {
```
sylvlecl: I think this is not needed in the interface

java/src/main/java/com/powsybl/dataframe/network/adders/NetworkElementAdder.java
```diff
@@ -27,5 +28,15 @@ public interface NetworkElementAdder {
      * The first dataframe is considered the "primary" dataframe, other dataframes
      * can provide additional data (think steps for the tap changers).
      */
-    void addElements(Network network, List<UpdatingDataframe> dataframes);
+    default void addElements(Network network, List<UpdatingDataframe> dataframes) {
+    }
+
+    default void addElementsWithBay(Network network, List<UpdatingDataframe> dataframe, boolean throwException, Reporter reporter) {
+    }
```
sylvlecl: Would be better to throw an unsupported operation exception by default

java/src/main/java/com/powsybl/dataframe/network/adders/ShuntDataframeAdder.java
```diff
@@ -41,7 +45,10 @@ public class ShuntDataframeAdder implements NetworkElementAdder {
             SeriesMetadata.ints("section_count"),
             SeriesMetadata.doubles("target_deadband"),
             SeriesMetadata.doubles("target_v"),
-            SeriesMetadata.strings("model_type")
+            SeriesMetadata.strings("model_type"),
+            SeriesMetadata.strings("busbar_section_id"),
```
sylvlecl, line 49: Note for later: it's not ideal to have the new columns in the metadata of existing dataframes, but let's keep it like that for now.

docs/user_guide/network.rst
colineplqt: I think the documentation needs to evolve a lot in the future. Either when other hypothesis are available, to create the network using all of them or at least when the extensions will be available, to show really all the possibilities offered. I did not put too much for now because I really think there will be a lot to evolve.

tests/test_network.py
colineplqt: Do you think I need to add one with the keywords for each element?

sylvlecl: Well it would be nice, but maybe we could kind of "automate" it to avoid having to do it by hand, by using the same input dictionary to build a dataframe or keywords, in all tests.

I propose to postpone this, it could be a more general improvement of tests (also for existing methods).

docs/user_guide/network.rst
sylvlecl: Ok!

sylvlecl (approved):

PR #492, opened 2022-09-08
Network modifications: Add utility functions for creating feeder bays.

colineplqt:

No

Feature

Adds utility functions to get the positions of connectables and the available ones in a voltage level


java/src/main/java/com/powsybl/python/network/NetworkModificationCFunctions.java
```diff
@@ -0,0 +1,89 @@
...
+        Network network = ObjectHandles.getGlobal().get(networkHandle);
+        String busbarSectionIdStr = CTypeUtil.toString(busbarSectionId);
+        BusbarSection busbarSection = network.getBusbarSection(busbarSectionIdStr);
+        Optional<Range<Integer>> positionsOrders = getUnusedOrderPositionsBefore(busbarSection);
+        if (positionsOrders.isPresent()) {
+            int max = positionsOrders.get().getMaximum();
+            int min = positionsOrders.get().getMinimum();
+            return createIntegerArray(Arrays.asList(min, max));
+        } else {
+            //TODO: It actually doesn't work
```
colineplqt: In this function and the one below, I am not sure what should be returned in the case where there are no extensions. The called method will raise an exception that is not caught in python.

java/src/main/java/com/powsybl/python/network/NetworkModificationCFunctions.java
```diff
@@ -0,0 +1,100 @@
...
+    public static PyPowsyblApiHeader.ArrayPointer<PyPowsyblApiHeader.SeriesPointer> getConnectablesOrderPositions(IsolateThread thread, ObjectHandle networkHandle,
+                                                     CCharPointer voltageLevelId, PyPowsyblApiHeader.ExceptionHandlerPointer exceptionHandlerPtr) {
+        return doCatch(exceptionHandlerPtr, () -> {
+            String voltageLevelIdStr = CTypeUtil.toString(voltageLevelId);
+            Network network = ObjectHandles.getGlobal().get(networkHandle);
+
+            VoltageLevel voltageLevel = network.getVoltageLevel(voltageLevelIdStr);
+            Map<String, List<Integer>> feederPositionsOrders = getFeederPositionsByConnectable(voltageLevel);
+            //TODO: Make it better. For now there's nothing to create a dataframe with a list as a column
+            Map<String, String> feederPositionsOrdersString = new HashMap<>();
```
colineplqt: For now, there's no mapper that would allow a list in a column of the dataframe. So for a quick fix, I put converted it to String. To be discussed.

java/src/main/java/com/powsybl/python/network/NetworkModificationCFunctions.java
```diff
@@ -0,0 +1,100 @@
...
+            if (positionsOrders.isPresent()) {
+                int max = positionsOrders.get().getMaximum();
+                int min = positionsOrders.get().getMinimum();
+                return createIntegerArray(Arrays.asList(min, max));
+            } else {
+                return createIntegerArray(Collections.EMPTY_LIST);
+            }
+        } catch (PowsyblException e) {
+            throw new PowsyblException(e.getMessage());
+        }
```
colineplqt: Here and on the method afterwards, it seems like the exceptions are not caught...

java/src/main/java/com/powsybl/python/network/NetworkModificationCFunctions.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @CEntryPoint(name = "getConnectablesOrderPositions")
+    public static PyPowsyblApiHeader.ArrayPointer<PyPowsyblApiHeader.SeriesPointer> getConnectablesOrderPositions(IsolateThread thread, ObjectHandle networkHandle,
+                                                     CCharPointer voltageLevelId, PyPowsyblApiHeader.ExceptionHandlerPointer exceptionHandlerPtr) {
+        return doCatch(exceptionHandlerPtr, () -> {
+            String voltageLevelIdStr = CTypeUtil.toString(voltageLevelId);
+            Network network = ObjectHandles.getGlobal().get(networkHandle);
+
+            VoltageLevel voltageLevel = network.getVoltageLevel(voltageLevelIdStr);
+            Map<String, List<Integer>> feederPositionsOrders = getFeederPositionsByConnectable(voltageLevel);
```
EtienneLt, line 48: there is a problem here, getFeederPositionsByConnectable return a Map<String, Integer>

java/src/main/java/com/powsybl/python/network/Dataframes.java
```diff
@@ -348,6 +355,21 @@ private static DataframeMapper<VoltageLevel> createBusBreakerViewBuses() {
                 .build();
     }
 
+    private static List<Map.Entry<String, String>> getPositions(Map<String, String> map) {
+        Set<Map.Entry<String, String>> entriesSet = map.entrySet();
+        List<Map.Entry<String, String>> entriesList = new ArrayList<>(entriesSet.size());
+        entriesList.addAll(entriesSet);
+        return entriesList;
```
EtienneLt, line 368: I'm not sure but something like 
```java
return new ArrayList<>(map.entrySet());
```
or
```java
return map.entrySet().stream().collect(Collectors.toList());
``` 
seems to work

java/src/main/java/com/powsybl/python/network/NetworkModificationCFunctions.java
EtienneLt: is it still the case ?

EtienneLt: just maybe both methods can be merged into one they seemed to be very similar

java/src/main/java/com/powsybl/python/network/Dataframes.java
```diff
@@ -348,6 +357,22 @@ private static DataframeMapper<VoltageLevel> createBusBreakerViewBuses() {
...
+        List<Pair<String, ConnectablePosition.Feeder>> entriesList = new ArrayList<>(entriesSet.size());
+        entriesSet.forEach(e -> e.getValue().forEach(feeder -> entriesList.add(Pair.of(e.getKey(), feeder))));
+        return entriesList;
+    }
+
+    private static DataframeMapper<Map<String, List<ConnectablePosition.Feeder>>> createFeederMapDataframe() {
+        return new DataframeMapperBuilder<Map<String, List<ConnectablePosition.Feeder>>, Pair<String, ConnectablePosition.Feeder>>()
+                .itemsProvider(Dataframes::getPositions)
+                .stringsIndex("connectable_id", Pair::getLeft)
+                .ints("order_position", pair -> pair.getRight().getOrder().orElse(null))
```
colineplqt: Here actually the null in case there's no order is wrong, it raises an exception when the dataframe is created. I don't know yet how to fix it.

pypowsybl/network.py
```diff
@@ -4124,6 +4124,39 @@ def remove_extensions(self, extension_name: str, ids: _Union[str, _List[str]]) -
...
+    def get_connectables_order_positions(self, voltage_level_id: str) -> _DataFrame:
+        """
+        Gets the order positions of every connectable of a given voltage level in a dataframe.
+        """
+        series_array = _pp.get_connectables_order_positions(self._handle, voltage_level_id)
+        return _create_data_frame_from_series_array(series_array).sort_values(by=['order_position'])
+
+    def get_unused_order_positions_before(self, busbar_section_id: str) -> _Optional[pd.Interval]:
+        """
+        Gets all the available positions between the lowest used position of a given busbar section and the
```
sylvlecl: We need to have a short description as first line (in particular for HTML doc).

pypowsybl/network.py
```diff
@@ -4124,6 +4124,39 @@ def remove_extensions(self, extension_name: str, ids: _Union[str, _List[str]]) -
...
+        """
+        positions = _pp.get_unused_order_positions_before(self._handle, busbar_section_id)
+        if len(positions) == 0:
+            return None
+        else:
+            return pd.Interval(left=positions[0], right=positions[1], closed='both')
+
+    def get_unused_order_positions_after(self, busbar_section_id: str) -> _Optional[pd.Interval]:
+        """
+        Gets all the available positions between the highest used position of a given busbar section and the
```
sylvlecl: We need to have a short description as first line (in particular for HTML doc).

pypowsybl/network.py
```diff
@@ -4124,6 +4124,39 @@ def remove_extensions(self, extension_name: str, ids: _Union[str, _List[str]]) -
...
+        Gets the order positions of every connectable of a given voltage level in a dataframe.
+        """
+        series_array = _pp.get_connectables_order_positions(self._handle, voltage_level_id)
+        return _create_data_frame_from_series_array(series_array).sort_values(by=['order_position'])
+
+    def get_unused_order_positions_before(self, busbar_section_id: str) -> _Optional[pd.Interval]:
+        """
+        Gets all the available positions between the lowest used position of a given busbar section and the
+        highest used position of the busbar section with the section index equal to the section index of the given
+        busbar section minus one. The result is a list with the lowest available value as the first integer and the
```
sylvlecl: I find it weird to have only those 2 extreme values and not the available values that we could have in-between.
The fact that we use and interval as return type can let the user believe that all values in-between are available.

Actually, I am really not convinced that the 2 methods "get_unused_positions" are useful, because the "get_connectables_order_positions" already provides us all the occupied slots, and therefore also the free slots, or am I missing something ?

java/src/main/java/com/powsybl/python/network/NetworkModificationCFunctions.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @CEntryPoint(name = "getUnusedConnectableOrderPositionsBefore")
+    public static PyPowsyblApiHeader.ArrayPointer<CIntPointer> getUnusedConnectableOrderPositionsBefore(IsolateThread thread, ObjectHandle networkHandle,
+                                                                                             CCharPointer busbarSectionId,
+                                                                                             PyPowsyblApiHeader.ExceptionHandlerPointer exceptionHandlerPtr) {
+        Network network = ObjectHandles.getGlobal().get(networkHandle);
+        String busbarSectionIdStr = CTypeUtil.toString(busbarSectionId);
+        BusbarSection busbarSection = network.getBusbarSection(busbarSectionIdStr);
+        try {
```
sylvlecl: (if we keep the method)
The try catch is not useful, but we need to wrap the method in a `doCatch` to propagate exceptions to C code.

java/src/main/java/com/powsybl/python/network/NetworkModificationCFunctions.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @CEntryPoint(name = "getUnusedConnectableOrderPositionsAfter")
+    public static PyPowsyblApiHeader.ArrayPointer<CIntPointer> getUnusedConnectableOrderPositionsAfter(IsolateThread thread, ObjectHandle networkHandle,
+                                                                                                          CCharPointer busbarSectionId,
+                                                                                                          PyPowsyblApiHeader.ExceptionHandlerPointer exceptionHandlerPtr) {
+        Network network = ObjectHandles.getGlobal().get(networkHandle);
+        String busbarSectionIdStr = CTypeUtil.toString(busbarSectionId);
+        BusbarSection busbarSection = network.getBusbarSection(busbarSectionIdStr);
+        try {
```
sylvlecl: (if we keep the method)
The try catch is not useful, but we need to wrap the method in a `doCatch` to propagate exceptions to C code.

pypowsybl/network.py
```diff
@@ -4124,6 +4124,39 @@ def remove_extensions(self, extension_name: str, ids: _Union[str, _List[str]]) -
             ids = [ids]
         _pp.remove_extensions(self._handle, extension_name, ids)
 
+    def get_connectables_order_positions(self, voltage_level_id: str) -> _DataFrame:
```
sylvlecl: **A general comment about the feature**
there is actually already a generic mechanism to expose extensions data, through the `network.get_extensions` method.
However, it's indeed not very user-friendly because you need to know the name of the extension and so on.
So exposing their data through more targetted user-friendly methods makes sense.

But still, we should probably ALSO expose the `ConnectablePosition` extension through the generic mechanism, to be able to view the directions, the feeder name, and to be able to change them.
This should be done in another branch.

**About the place of the method**
I wonder if we should leave it in the Network class, because it's very, very, specific.

pypowsybl/network.py
```diff
@@ -4124,6 +4124,39 @@ def remove_extensions(self, extension_name: str, ids: _Union[str, _List[str]]) -
...
+        """
+        Gets the order positions of every connectable of a given voltage level in a dataframe.
```
sylvlecl: I think we should go into more details about what is the "order position" (in the long description, so after the first line): 
what it is, where it used.

pypowsybl/network.py
sylvlecl: Tracked it in #495

pypowsybl/network.py
colineplqt: Actually, these methods do give all the available positions on the "right" or "left" of the busbar section. Everything in the interval is available.
Let's say you have two busbar sections separated by a disconnector. One has feeders at positions 5, 6, 7 and the other has feeders at positions 10,11.
Then get_unused_order_positions_before(bbs1) would return [-infinity, 4]
get_unused_order_positions_after(bbs1) and get_unused_order_positions_before(bbs2) would return [8,9] and finally
get_unused_order_positions_after(bbs2) would return [12, +infinity].
It is quite practical if one wants to automate the computation of the position : one can then just get the available positions depending on whether one wants to put the equipment on the right or on the left and then for example take the value in the middle on the interval. 
The other method get_connectables_order_positions is easier to use to put a feeder between two feeders or with displaying the taken positions. 
Is it that way that you understood it?

sylvlecl: I misunderstood the behaviour indeed, sorry !

colineplqt: I will add some documentation, because it is not very clear or easy to explain.

colineplqt:
More documentation will be added with the improved documentation from create feeder bay.

cpp/src/pypowsybl.cpp
```diff
@@ -1093,6 +1093,13 @@ std::vector<SeriesMetadata> getTwtFeederBaysMetadata() {
     return res;
 }
 
+SeriesArray* getConnectablesOrderPositions(const JavaHandle& network, const std::string voltage_level_id) {
+    return new SeriesArray(callJava<array*>(::getConnectablesOrderPositions, network, (char*) voltage_level_id.c_str()));
+}
 
+std::vector<int> getUnusedConnectableOrderPositions(const pypowsybl::JavaHandle network, const std::string busbarSectionId, const std::string beforeOrAfter) {
+    auto positionsArrayPtr = callJava<array*>(::getUnusedConnectableOrderPositions, network, (char*) busbarSectionId.c_str(), (char*) beforeOrAfter.c_str());
+    return toVector<int>(positionsArrayPtr);
```
sylvlecl: I think there is a memory leak: we need to free the `array*` returned from java here
(I see there is another one in the existing code!)

java/src/main/java/com/powsybl/python/network/Dataframes.java
```diff
@@ -351,6 +361,22 @@ private static DataframeMapper<VoltageLevel> createBusBreakerViewBuses() {
                 .build();
     }
 
+    private static List<Pair<String, ConnectablePosition.Feeder>> getPositions(Map<String, List<ConnectablePosition.Feeder>> map) {
+        Set<Map.Entry<String, List<ConnectablePosition.Feeder>>> entriesSet = map.entrySet();
```
sylvlecl, line 365: Very minor detail: note that we could probably make it a more readable 1-liner with streams

java/src/main/java/com/powsybl/python/network/NetworkModificationsCFunctions.java
```diff
@@ -199,4 +202,95 @@ public static void createBranchFeederBaysTwt(IsolateThread thread, ObjectHandle
...
+                return createIntegerArray(Arrays.asList(min, max));
+            } else {
+                return createIntegerArray(Collections.EMPTY_LIST);
+            }
+        });
+
+    }
+
+    static Map<String, List<ConnectablePosition.Feeder>> getFeedersByConnectable(VoltageLevel voltageLevel) {
+        Map<String, List<ConnectablePosition.Feeder>> feedersByConnectable = new HashMap<>();
```
sylvlecl, line 246: Very minor detail too: note that the hash map will introduce its own sort order:
it could be more natural to keep the order of the connectables stream?

sylvlecl (commented):
minor memory leak to fix

PR #494, opened 2022-09-13
Wrap create branch feeder bays network modification

obrix:
This pull request provide a python API for the create branch feeder bays implementation. The branch is based on the 4.10.0 adaptation branch.

**Other information**:


obrix:
@annetill  This lacks doc and tests but feel free to comment regarding API and provided parameters.

pypowsybl/network.py
```diff
@@ -4374,3 +4374,9 @@ def get_extensions_names() -> _List[str]:
         the names of the available extensions
     """
     return _pp.get_extensions_names()
+
+def create_branch_feeder_bays(n: Network, bus_bar_section_id_1: str, bus_bar_section_id_2: str, position_order_1: int, position_order_2: int, direction_1: str, direction_2, df_new_line: _DataFrame = None, **kwargs: _ArrayLike) -> None:
+    metadata = _pp.get_network_elements_creation_dataframes_metadata(ElementType.LINE)
+    df_new_line = _adapt_df_or_kwargs(metadata[0], df_new_line, **kwargs)
+    c_df = _create_c_dataframe(df_new_line, metadata[0])
+    _pp.create_branch_feeder_bays(n._handle, bus_bar_section_id_1, bus_bar_section_id_2, position_order_1, position_order_2, direction_1, direction_2, c_df)
```
sylvlecl: let's restore the newline before merge

pypowsybl/network.py
```diff
@@ -4374,3 +4374,9 @@ def get_extensions_names() -> _List[str]:
         the names of the available extensions
     """
     return _pp.get_extensions_names()
+
+def create_branch_feeder_bays(n: Network, bus_bar_section_id_1: str, bus_bar_section_id_2: str, position_order_1: int, position_order_2: int, direction_1: str, direction_2, df_new_line: _DataFrame = None, **kwargs: _ArrayLike) -> None:
+    metadata = _pp.get_network_elements_creation_dataframes_metadata(ElementType.LINE)
```
sylvlecl: Let's not forget the docstring before merge

sylvlecl: We probably need to have 2 methods, on for lines and one for 2 windings transformers

pypowsybl/network.py
```diff
@@ -4374,3 +4374,9 @@ def get_extensions_names() -> _List[str]:
         the names of the available extensions
     """
     return _pp.get_extensions_names()
+
+def create_branch_feeder_bays(n: Network, bus_bar_section_id_1: str, bus_bar_section_id_2: str, position_order_1: int, position_order_2: int, direction_1: str, direction_2, df_new_line: _DataFrame = None, **kwargs: _ArrayLike) -> None:
```
sylvlecl: Like for injections, I would rather use only dataframes as arguments, to be able to create multiple elements at once.

Also, I think position should be optional since it does not exist in all networks (should be natural if we use dataframes).

java/src/main/java/com/powsybl/python/network/NetworkCFunctions.java
```diff
@@ -863,4 +868,43 @@ public static void setMinValidationLevel(IsolateThread thread, ObjectHandle netw
...
+                                        ExceptionHandlerPointer exceptionHandlerPtr) {
+        doCatch(exceptionHandlerPtr, () -> {
+            Network network = ObjectHandles.getGlobal().get(networkHandle);
+            String bbsId1Str = CTypeUtil.toString(bbsId1);
+            String bbsId2Str = CTypeUtil.toString(bbsId2);
+            String direction1Str = CTypeUtil.toString(direction1);
+            String direction2Str = CTypeUtil.toString(direction2);
+
+            UpdatingDataframe df = createDataframe(cDataframeLine);
+            LineAdder adder = network.newLine();
```
sylvlecl: We should try to re-use more the code of line creation that already exists, like we're trying to do for injections.

sylvlecl (changes requested):
Comments similar to injection bays PR: let's try to finish that one first, before going on with this one.

pypowsybl/network.py
obrix: The two methods have been implemented. I will try to see how to better factorized what is duplicated.

pypowsybl/network.py
obrix: All arguments has been moved into the dataframe. There is only the substation for twt creation that remain separated.

sylvlecl: Actually, I think the substation argument is not useful (it's redundant): 
we should be able to retrieve it from the bus or busbars that are defined. Is it possible to implement this ?

java/src/main/java/com/powsybl/dataframe/network/adders/FeederBaysTwtSeries.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.powsybl.dataframe.SeriesMetadata;
+import com.powsybl.dataframe.update.UpdatingDataframe;
+
+import java.util.List;
+
+public class FeederBaysTwtSeries extends AbstractFeederBaysSeries {
+    private static final List<SeriesMetadata> METADATA = List.of(
+            SeriesMetadata.stringIndex("id"),
+            SeriesMetadata.strings("bus_bar_section_id_1"),
```
colineplqt: I think here it should be busbar_section_id_1 (same for bus_bar_section_id_2) or we change it on create_feeder_bay to be coherent. What do you think?

java/src/main/java/com/powsybl/dataframe/network/adders/FeederBaysTwtSeries.java
```diff
@@ -0,0 +1,33 @@
...
+            SeriesMetadata.strings("bus_bar_section_id_2"),
+            SeriesMetadata.ints("position_order_1"),
+            SeriesMetadata.ints("position_order_2"),
+            SeriesMetadata.strings("direction_1"),
+            SeriesMetadata.strings("direction_2"),
```
colineplqt, line 16: Also the direction could be optional.

colineplqt (commented):
A few comments related to coherence with create_feeder_bay. Maybe the voltage level ids should also not be in the dataframe.

java/src/main/java/com/powsybl/dataframe/network/adders/FeederBaysTwtSeries.java
obrix: Yes it seems busbar and not bus_bar is used everywhere on python side. I will change it here.

java/src/main/java/com/powsybl/dataframe/network/adders/FeederBaysTwtSeries.java
obrix: I think applyIfPresent in AbstractFeederBaysSeries handles the fact it is optional ?

pypowsybl/network.py
```diff
@@ -4751,3 +4749,44 @@ def _get_c_dataframes_and_add_voltage_level_id(network: Network, dfs: _List[_Opt
         else:
             c_dfs.append(_create_c_dataframe(df, metadata[i]))
     return c_dfs
+
+def get_branch_feeder_bays_metadata(element_type: ElementType) -> _List[_List[_pp.SeriesMetadata]]:
+    metadata = _pp.get_network_elements_creation_dataframes_metadata(element_type)
+    metadata[0].append(_pp.SeriesMetadata('busbar_section_id_1', 0, True, True, True))
```
sylvlecl: Maybe it would be more consistent with the existing code to get them from java, after all.
It seems that you already have defined the METADATA fields in the java classes :)

pypowsybl/network.py
```diff
@@ -4751,3 +4749,44 @@ def _get_c_dataframes_and_add_voltage_level_id(network: Network, dfs: _List[_Opt
...
+    metadata[0].append(_pp.SeriesMetadata('busbar_section_id_2', 0, True, True, True))
+    metadata[0].append(_pp.SeriesMetadata('position_order_1', 2, True, True, True))
+    metadata[0].append(_pp.SeriesMetadata('position_order_2', 2, True, True, True))
+    metadata[0].append(_pp.SeriesMetadata('direction_1', 0, True, True, True))
+    metadata[0].append(_pp.SeriesMetadata('direction_2', 0, True, True, True))
+    return metadata
+
+def create_branch_feeder_bays_line(network: Network, df_new_line: _DataFrame = None, **kwargs: _ArrayLike) -> None:
```
sylvlecl: For consistency with injections: `create_line_bays`
I would rename df_new_line to simply `df`, also, for consistency.

pypowsybl/network.py
```diff
@@ -4751,3 +4749,44 @@ def _get_c_dataframes_and_add_voltage_level_id(network: Network, dfs: _List[_Opt
...
+        df_new_line: Attributes as a dataframe.
+    Notes:
+        The voltage level containing the busbar section should be described in node/breaker topology.
+    """
+    metadata = get_branch_feeder_bays_metadata(ElementType.LINE)
+    df_new_line = _adapt_df_or_kwargs(metadata[0], df_new_line, **kwargs)
+    c_df = _create_c_dataframe(df_new_line, metadata[0])
+    _pp.create_branch_feeder_bays_line(network._handle, c_df)
+
+def create_branch_feeder_bays_twt(network: Network, df_new_twt: _DataFrame = None, **kwargs: _ArrayLike) -> None:
```
sylvlecl: For consistency with injections: `create_2_windings_transformer_bays`
I would rename df_new_twt to simply `df`, also, for consistency.

sylvlecl (approved):

PR #511, opened 2022-10-06
Wrap replaceTeePointByVoltageLevelOnLine

obrix:
Signed-off-by: Bertrand Rix <bertrand.rix@artelys.com>

**Other information**:


obrix:
For the documentation of the function I used the one from powsybl core. Also the parameters name are the same as in the core API but I don't really get logic (line1c_id/ linec2_id), in doubt I kept the same names.

pypowsybl/network.py
```diff
@@ -4751,3 +4751,38 @@ def _get_c_dataframes_and_add_voltage_level_id(network: Network, dfs: _List[_Opt
...
+        linezp_id : The ID of the existing line connecting the tee point to the attached voltage level
+        voltage_level_id : The ID of the existing attached voltage level
+        bbs_or_bus_id : The ID of the existing bus or bus bar section in the attached voltage level voltageLevelId,
+          where we want to connect the new lines line1C and lineC2
+        line1c_id : The ID of the new line connecting the first voltage level to the attached voltage level
+        linec2_id : The ID of the new line connecting the second voltage level to the attached voltage level
+        line1c_name : The optional name of the new line connecting the first voltage level to the attached voltage level
+        linec2_name : The optional name of the new line connecting the second voltage level to the attached voltage level
+
+        VL1 ---------- tee point ---------- VL2                            VL1 ---------- attached voltage level ---------- VL2
```
sylvlecl: We need to check how this renders in the HTML doc, I have a bad feeling :)

pypowsybl/network.py
```diff
@@ -4751,3 +4751,38 @@ def _get_c_dataframes_and_add_voltage_level_id(network: Network, dfs: _List[_Opt
         else:
             c_dfs.append(_create_c_dataframe(df, metadata[i]))
     return c_dfs
+
+def replace_tee_point_by_voltage_level_on_line(network: Network, line1z_id: str, linez2_Id: str, linezp_id: str,
+                                  voltage_level_id: str, bbs_or_bus_id: str, line1c_id: str, linec2_id: str,
+                                  line1c_name: str = None, linec2_name: str = None) -> None:
+    """
+    This method transform the action done in the create_line_on_line function into the action done in the connect_voltage_level_on_line class :
```
sylvlecl: We need to have a 1-line short description as the first line, otherwise the doc will render ugly things.

sylvlecl (commented):
We need to check the rendering in HTML doc

pypowsybl/network.py
obrix: I understand the bad feeling, I'll remove it as suggested by Anne we will put that in the readthedoc.

pypowsybl/network.py
obrix: Done

sylvlecl:
As discussed today, let's put this on hold until java API is reviewed, in particular:
- using voltage level ID (tee point ID) instead of lines IDs
- parameters naming

obrix:
@sylvlecl I updated the pull request with latest changes in powsybl-core. Can you take a look and see if there is still work to do ?

EtienneLt (approved):

obrix:
@EtienneLt I am fixing the small conflict with main.

java/src/main/java/com/powsybl/python/network/NetworkModificationsCFunctions.java
```diff
@@ -327,4 +327,37 @@ private static List<ConnectablePosition.Feeder> get3wtFeeders(ConnectablePositio
...
+                                                           CCharPointer bbsOrBusId,
+                                                           CCharPointer newLine1Id,
+                                                           CCharPointer newLine1Name,
+                                                           CCharPointer newLine2Id,
+                                                           CCharPointer newLine2Name,
+                                                           PyPowsyblApiHeader.ExceptionHandlerPointer exceptionHandlerPtr) {
+
+        String teePointLine1Str = CTypeUtil.toString(teePointLine1);
+        String teePointLineStr = CTypeUtil.toString(teePointLine2);
+        String teePointLineToRemoveStr = CTypeUtil.toStringOrNull(teePointLineToRemove);
```
EtienneLt, line 345: @obrix @sylvlecl the builder has requireNonNull on several arguments on which we use CTypeUtil.toStringOrNull(), do we let the builder make all the checks or we use CTypeUtil.toString() ?

obrix: No strong opinion here, making the builder do the real check will probably give more explicit and easy to spot errors.

EtienneLt: @geofjamg @sylvlecl ?

sylvlecl: It's often better to try catch errors as early as possible, but it's OK for me this way:
the code of the builder is already in charge of checking arguments validity, let's not duplicate this role.

For ex. the builder could provide a meaningful error message, in that case it's better to let this through instead of having a plain null check before (but in the current code, it does not provide any such message, though :) ).

sylvlecl (approved):

java/src/main/java/com/powsybl/python/network/NetworkModificationsCFunctions.java
EtienneLt: ok

EtienneLt (approved):

PR #520, opened 2022-10-19
Dynawaltz integration

### LINKED ISSUE #506: Dynawaltz integration
Feature

We should be able to run DynaWaltz in pypowsybl. The Python API could be something like:
```python
import pypowsybl as pp

network = pp.network.load('xxx')

# create dynamic data mapping
model_mapping = pp.dynamic.create_model_mapping()
# or alternative to support
model_mapping = pp.dynamic.ModelMapping()
for l in network.loads:
    if load == "LOAD2":
        model_mapping.map_one_transformer_load(static_id=l.id, parameter_set_id='LOT')
    else:
        model_mapping.map_alpha_beta_load(static_id=l.id, parameter_set_id='LAB')

# etc, ...

# create timeseries config
timeseries_config = pp.dynamic.create_timeseries_config()
# or alternative to support
timeseries_config = pp.dynamic.TimeSeriesConfig()
for l in network.loads:
    timeseries_config.add(dynamic_model_id=l.id, variables=['load_PPu', 'load_QPu'])


# create simulation events and curves 
simu = pp.dynamic.create_simulation()
# or alternative to support
simu = pp.dynamic.Simulation()
simu.add_event(10, pp.dynamic.EventType.BRANCH_DISCONNECTION, 'l1')
simu.add_event(15, pp.dynamic.EventType.SWITCH_CLOSE, 's1')

# run simulation
result = simu.run(network, model_mapping, timeseries_config, start=0, stop=60);
print(result.status)
# get a Pandas series
ts = result.get_timeseries('c1')
```

Alternative to create model mapping from data frames to allow bulk creations

```python 
load_one_transformer_data = {}
load_alpha_beta_data = {}
for l in network.loads:
    if load == "LOAD2":
        load_one_transformer_data[l.id] = ['LOT']
    else:
        load_alpha_beta_data[l.id] = ['LAB']

model_mapping.map_one_transformer_loads(pd.DataFrame.from_dict(load_one_transformer_data, orient='index',  columns=['parameter_set_id']))
model_mapping.map_alpha_beta_loads(pd.DataFrame.from_dict(load_alpha_beta_data, orient='index',  columns=['parameter_set_id']))
```

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

nicolas-pierr:
- [x] Licence headers

**Does this PR already have an issue describing the problem ?**
Fixes #506


nicolas-pierr:
Hello @geofjamg and @flo-dup 
I'm basing the dynawaltz integration on the powsybl-dynawo api.
In the given part of the example, you want to add dynamically events with their timestep : 
```
simu = pp.dynamic.Simulation()
simu.add_event(10, pp.dynamic.EventType.BRANCH_DISCONNECTION, 'l1')
simu.add_event(15, pp.dynamic.EventType.SWITCH_CLOSE, 's1')
```
However, powsybl-dynawo doesn't allow it, it is only able to load via a parameter file (model.par).

Do you want this feature to be added or should it work like the current powsybl-dynawo way and load configurations and parameters from files ?

Also does this conversation should be in the issue or in this PR ?

Thank you

flo-dup:
Indeed the timestep is set through the `.par` file. Therefore you need to modify that file to put the timestep defined by the user. I think this needs to be done in powsybl-dynawo though, because it makes sense for any user to define the timestep in the event and not among some values in the `.par` file. So far the `.par` file is seen as an input or a config file in powsybl-dynawo, we need to change this...

flo-dup:
We could change the API in the meantime (I mean before powsybl-dynawo is modified to enable time step modifications);
```python
simu = pp.dynamic.Simulation()
simu.add_event(parametersSetId1, pp.dynamic.EventType.BRANCH_DISCONNECTION, 'l1')
simu.add_event(parametersSetId2, pp.dynamic.EventType.SWITCH_CLOSE, 's1')
```

java/src/main/java/com/powsybl/python/dynamic/DynamicModelMapper.java
```diff
@@ -0,0 +1,71 @@
+package com.powsybl.python.dynamic;
```
sylvlecl, line 8: Don't forget to add license headers at some point ;)

EtienneLt: and SPDX in the copyright.
```java
 /**
 * Copyright (c) 2022, RTE (http://www.rte-france.com)
 * This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/.
 *  SPDX-License-Identifier: MPL-2.0
 */
```

cpp/src/bindings.cpp
```diff
@@ -640,4 +640,43 @@ PYBIND11_MODULE(_pypowsybl, m) {
     m.def("remove_aliases", &pypowsybl::removeAliases, "remove specified aliases on a network", py::arg("network"), py::arg("dataframe"));
 
     m.def("close", &pypowsybl::closePypowsybl, "Closes pypowsybl module.");
+    /************************DYNAMIC SIMULATION WITH DYNAWALTZ*******************************/
+    //TODO descriptions of methods
```
sylvlecl: Actually docstrings here are not important, they are not exposed to the user.

About the separation comment above : I would propose to move everything to a dedicated function instead. We should probably have done it before to better organize the bindings.

cpp/src/bindings.cpp
```diff
@@ -640,4 +640,43 @@ PYBIND11_MODULE(_pypowsybl, m) {
...
+    m.def("create_dynamic_model_mapping", &pypowsybl::createDynamicModelMapping, "TODO desc");
+    m.def("create_timeseries_mapping", &pypowsybl::createTimeseriesMapping, "TODO desc");
+    m.def("create_event_mapping", &pypowsybl::createEventMapping, "TODO desc");
+
+    //running simulations
+    m.def("run_dynamic_model", &pypowsybl::runDynamicModel, "Run a dynamic model with Dynawaltz", py::call_guard<py::gil_scoped_release>(),
+        py::arg("dynamic_model"), py::arg("network"), py::arg("dynamic_mapping"), py::arg("event_mapping"), py::arg("timeseries_mapping"), py::arg("start"), py::arg("stop"));
+
+    //model mapping
+    m.def("add_alphabeta_load", &pypowsybl::addAlphaBetaLoad, "TODO desc", py::arg("dynamic_mapping_handle"), py::arg("static_id"), py::arg("dynamic_param"));
```
sylvlecl: I think instead of having multiple methods here, we could have only one for all models, which would receive a dataframe input exactly as we have done for all updates methods of the network. 
This will allow to:
- minimize the code boiler plate necessary for each kind of event
- map multiple models in only one call. This is very important for performances, since calling 100s of methods in a for loop will have very poor performance in python.

sylvlecl: (but of course we still need to have multiple methods on the python side, for the user api

pypowsybl/dynamic.py
```diff
@@ -35,46 +36,57 @@ def __init__(self) -> None:
...
+        self.add_all_dynamic_mappings(static_id=static_id,
+                                      parameter_set_id=dynamic_param,
+                                      mapping_type=MappingType.GENERATOR_SYNCHRONOUS_FOUR_WINDINGS_PROPORTIONAL_REGULATIONS)
 
     def add_current_limit_automaton(self, static_id: str, dynamic_param: str, branch_side: BranchSide) -> None:
         _pp.add_current_limit_automaton(
             self._handle, static_id, dynamic_param, branch_side.value)
 
-    def add_all_dynamic_mappings(self, df: _pd.DataFrame) -> None:
+    def add_all_dynamic_mappings(self, df: _pd.DataFrame = None, **kwargs: _ArrayLike) -> None:
```
NicolasPIERRE-Artelys, line 76: I have issues to type kwargs here.
Using the same typing as in network.py does not work since I allow enum values like : 
```
mapping.add_all_dynamic_mappings(static_id="LOAD1", parameter_set_id="LAB", mapping_type=MappingType.LOAD_ALPHA_BETA)
```

cpp/src/bindings.cpp
nicolas-pierr: The change has been done here : f90ea09cb29b308e0b84299496bc1948c0bb7bff .
However some mappings like CurrentLimitAutomaton (and maybe other mapping in the future) need more infos.

The actual behavior is to have some default values (possible because CurrentLimitAutomaton needs an enum argument) to be able to map them _via_ the dataframe call.
But they still need a special python/java binding to be able to specify all the arguments.

pypowsybl/dynamic.py
```diff
@@ -0,0 +1,190 @@
...
+    def add_generator_synchronous_four_windings_proportional_regulations(self, static_id: str, dynamic_param: str) -> None:
+        self.add_all_dynamic_mappings(static_id=static_id,
+                                      parameter_set_id=dynamic_param,
+                                      mapping_type=DynamicMappingType.GENERATOR_SYNCHRONOUS_FOUR_WINDINGS_PROPORTIONAL_REGULATIONS)
+
+    def add_current_limit_automaton(self, static_id: str, dynamic_param: str, branch_side: BranchSide) -> None:
+        _pp.add_current_limit_automaton(
+            self._handle, static_id, dynamic_param, branch_side.value)
+
+    def add_all_dynamic_mappings(self, df: _pd.DataFrame = None, **kwargs: _Union[str, DynamicMappingType]) -> None:
```
sylvlecl: This is not exactly the design I had in mind: we should have the same approach as in the network class:
- all methods above should accept a dataframe or keywords as arguments (`add_current_limit_automaton` etc), to be able to add at once multiple instances of the model
- this method should be removed: we don't want to support mixing different models in the same dataframe, because as you underlined, they may have different arguments, hence different columns, which this approach forbids.

java/src/main/java/com/powsybl/python/dynamic/DynamicModelMapper.java
```diff
@@ -0,0 +1,136 @@
...
+     */
+    public void addAllMappings(UpdatingDataframe mappingDf) {
+        StringSeries staticIdSeries = mappingDf.getStrings("static_id");
+        StringSeries parameterSetIdSeries = mappingDf.getStrings("parameter_set_id");
+        IntSeries dynamicModelSeries = mappingDf.getInts("mapping_type");
+        for (int row = 0; row < mappingDf.getRowCount(); ++row) {
+            DynamicMappingType type = DynamicMappingType.fromCValue(dynamicModelSeries.get(row));
+            String staticId = staticIdSeries.get(row);
+            String parameterSetId = parameterSetIdSeries.get(row);
+            switch (type) {
```
sylvlecl: Instead of this switch/case, we should imitate the facilities already in place for the network, with the `NetworkElementAdder` interface.

This should allow to have different columns for the current limit automaton, in particular.

pypowsybl/dynamic.py
```diff
@@ -0,0 +1,190 @@
...
+    def add_curve(self, dynamic_id: str, variable: str) -> None:
+        _pp.add_curve(self._handle, dynamic_id, variable)
+
+    def add_curves(self, dynamic_id: str, variables: _List[str]) -> None:
+        for var in variables:
+            _pp.add_curve(self._handle, dynamic_id, var)
+
+
+class EventType(_Enum):
+    QUADRIPOLE_DISCONNECT = 'QUADRIPOLE_DISCONNECT'
```
sylvlecl: We really need to re-work the naming of the events:
- I don't think "quadripole" is an english word ? In any case, in powsybl, we talk about "branch" instead
- I am not sure what "set point boolean" means: does it change the regulation status of a generator ?
- BRANCH_DISCONNECTION : is it a duplicate of QUADRIPOLE_DISCONNECT?  :)

pypowsybl/dynamic.py
```diff
@@ -0,0 +1,190 @@
...
+    @staticmethod
+    def get_possible_events() -> _List[EventType]:
+        return list(EventType)
+
+
+class SimulationResult:
+    def __init__(self, handle: _pp.JavaHandle) -> None:
+        self._handle = handle
+
+    def state(self) -> str:
```
sylvlecl: let's keep only `status`

pypowsybl/dynamic.py
```diff
@@ -0,0 +1,190 @@
...
+            f"Pypowsybl-DynamicSimulationError: Unknown event {event}")
+
+    @staticmethod
+    def get_possible_events() -> _List[EventType]:
+        return list(EventType)
+
+
+class SimulationResult:
+    def __init__(self, handle: _pp.JavaHandle) -> None:
+        self._handle = handle
```
sylvlecl, line 197: I propose to initialize all python data from the java handle (status, curves) in the constructor, this will avoid to call the java part each time we want to get the curves, returning a new python object.

Note the difference with the network `get_` methods, because the network is mutable. Here the result is immutable, so it makes more sense not to have `get_` methods.

pypowsybl/dynamic.py
```diff
@@ -0,0 +1,190 @@
...
+                self._handle,
+                network._handle,
+                model_mapping._handle,
+                event_mapping._handle,
+                timeseries_mapping._handle,
+                start, stop)
+        )
+
+    @staticmethod
+    def set_config(path: str, config_name: str) -> None:
```
sylvlecl: Should be removed until we support changing the config at runtime.

sylvlecl (commented):
A set of comments, did not finish the review yet: thanks for the work so far :)

pypowsybl/dynamic.py
nicolas-pierr: From what I understood in powsybl-dynawaltz package, "set point boolean" events can only be used with a generator id. It allows to connect/disconnect the generator.
I'll rename quadripole to branch in pypowsybl, but powsybl-dynawaltz still use "quadripole".
And yes now that the naming is clearer, BRANCH_DISCONNECTION and QUADRIPOLE_DISCONNECT are the same event (and refer to the QUADRIPOLE_DISCONNECT  event in powsybl-dynawaltz)

java/src/main/java/com/powsybl/python/dynamic/DynamicModelMapper.java
nicolas-pierr: I have rewritten the whole part about dynamic mapping to work with adders like in networks.

pypowsybl/dynamic.py
nicolas-pierr: I fixed it. I could not find a good way to make only one java call to get all curves. In Python it is easy to concatenate dataframes, however it is not in java. And there was no easy way to return an array of dataframes without getting into memory allocations troubles. I don't know if I missed an example of this type (passing an array of a generic object from java to python).

pypowsybl/dynamic.py
nicolas-pierr: You are right, it was not meant ot be pushed

nicolas-pierr:
I rebased on main, I saw that the omega_ref is automaticaly mapped to every synchronized generator. Do you confirm that the user will never have to map omega_ref ?

nicolas-pierr:
Also I made a branch on pypowsyb-notebooks with an example to run.
All data asssociated (config, network and .par files) is provided.
Specific branch is [here](https://github.com/powsybl/pypowsybl-notebooks/tree/dynawaltz)

java/src/main/java/com/powsybl/dataframe/dynamic/adders/GeneratorSynchronousFourWindingsAdder.java
```diff
@@ -0,0 +1,11 @@
+package com.powsybl.dataframe.dynamic.adders;
+
```
EtienneLt, line 9: headers and author missing

java/src/main/java/com/powsybl/dataframe/dynamic/adders/GeneratorSynchronousFourWindingsProportionalRegulationsAdder.java
```diff
@@ -0,0 +1,11 @@
+package com.powsybl.dataframe.dynamic.adders;
+
```
EtienneLt, line 9: headers and author missing

java/src/main/java/com/powsybl/dataframe/dynamic/adders/GeneratorSynchronousThreeWindingsAdder.java
```diff
@@ -0,0 +1,11 @@
+package com.powsybl.dataframe.dynamic.adders;
+
```
EtienneLt, line 9: same

java/src/main/java/com/powsybl/dataframe/dynamic/adders/GeneratorSynchronousThreeWindingsProportionalRegulationsAdder.java
```diff
@@ -0,0 +1,11 @@
+package com.powsybl.dataframe.dynamic.adders;
+
```
EtienneLt, line 9: same

java/src/main/java/com/powsybl/dataframe/dynamic/adders/OneTransformerLoadAdder.java
```diff
@@ -0,0 +1,10 @@
+package com.powsybl.dataframe.dynamic.adders;
+
```
EtienneLt, line 9: same

java/src/main/java/com/powsybl/python/dynamic/CurveMappingSupplier.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.LinkedList;
+import java.util.List;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+import com.powsybl.dynamicsimulation.Curve;
+import com.powsybl.dynamicsimulation.CurvesSupplier;
+import com.powsybl.dynawaltz.DynaWaltzCurve;
+import com.powsybl.iidm.network.Network;
+
```
EtienneLt, line 20: author missing

java/src/main/java/com/powsybl/python/dynamic/DynamicModelMapper.java
```diff
@@ -0,0 +1,79 @@
...
+import com.powsybl.dynawaltz.automatons.CurrentLimitAutomaton;
+import com.powsybl.dynawaltz.dynamicmodels.GeneratorSynchronousFourWindings;
+import com.powsybl.dynawaltz.dynamicmodels.GeneratorSynchronousFourWindingsProportionalRegulations;
+import com.powsybl.dynawaltz.dynamicmodels.GeneratorSynchronousThreeWindings;
+import com.powsybl.dynawaltz.dynamicmodels.GeneratorSynchronousThreeWindingsProportionalRegulations;
+import com.powsybl.dynawaltz.dynamicmodels.LoadAlphaBeta;
+import com.powsybl.dynawaltz.dynamicmodels.LoadOneTransformer;
+import com.powsybl.iidm.network.Branch;
+import com.powsybl.iidm.network.Network;
+
```
EtienneLt, line 23: author missing

java/src/main/java/com/powsybl/python/dynamic/DynamicSimulationCFunctions.java
```diff
@@ -0,0 +1,209 @@
...
+import com.powsybl.python.commons.PyPowsyblApiHeader.DataframeMetadataPointer;
+import com.powsybl.python.commons.PyPowsyblApiHeader.DataframePointer;
+import com.powsybl.python.commons.PyPowsyblApiHeader.DynamicMappingType;
+import com.powsybl.python.commons.PyPowsyblApiHeader.SeriesPointer;
+import com.powsybl.python.commons.Util;
+import com.powsybl.python.network.Dataframes;
+import com.powsybl.python.network.NetworkCFunctions;
+import com.powsybl.timeseries.DoublePoint;
+import com.powsybl.timeseries.TimeSeries;
+
```
EtienneLt, line 45: author missing

java/src/main/java/com/powsybl/python/dynamic/DynamicSimulationContext.java
```diff
@@ -0,0 +1,58 @@
...
+import com.powsybl.commons.PowsyblException;
+import com.powsybl.dynamicsimulation.CurvesSupplier;
+import com.powsybl.dynamicsimulation.DynamicModelsSupplier;
+import com.powsybl.dynamicsimulation.DynamicSimulation;
+import com.powsybl.dynamicsimulation.DynamicSimulationParameters;
+import com.powsybl.dynamicsimulation.DynamicSimulationProvider;
+import com.powsybl.dynamicsimulation.DynamicSimulationResult;
+import com.powsybl.dynamicsimulation.EventModelsSupplier;
+import com.powsybl.iidm.network.Network;
+
```
EtienneLt, line 24: same

java/src/main/java/com/powsybl/python/dynamic/EventSupplier.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.List;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+import com.powsybl.dynamicsimulation.EventModel;
+import com.powsybl.dynamicsimulation.EventModelsSupplier;
+import com.powsybl.dynawaltz.events.EventQuadripoleDisconnection;
+import com.powsybl.dynawaltz.events.EventSetPointBoolean;
+import com.powsybl.iidm.network.Network;
+
```
EtienneLt, line 20: same

EtienneLt (approved):

nicolas-pierr:
Hello, some changes from this morning broke my compilation. I see that the generators models have changed completly. Can you confirm that GeneratorSynchronous and GeneratorFictious are the new mapping for generators that must be done ? And that GeneratorSynchronousThreeWindings and its friends were deleted from powsybl-dynawaltz ?

flo-dup:
Indeed, all synchronous generators have been merged into one `GeneratorSynchronous` class, which takes the library name as new parameter. Therefore you have to change [these mappings](https://github.com/powsybl/pypowsybl/blob/f4a98d1d0bdb9e93be535474713fa719e9db4e58/java/src/main/java/com/powsybl/dataframe/dynamic/adders/DynamicMappingAdderFactory.java#L22-L29) so that they each give a `GeneratorSynchronousAdder` which will retain the correct corresponding library name.

nicolas-pierr:
I updated the mappings to `GeneratorSynchronous` variants already implemented. They now use the `GeneratorSynchronous` class. I also created a `DynamicMappingAdder` for `GeneratorSynchronous`, it allows to add via dataframe different types of `GeneratorSynchronous` with a single call of python.

nicolas-pierr:
I also updated the events mapping with the new eventTime and their arguments.
There are compilations errors due to commits on main.

nicolas-pierr:
Some information about some changes under the hood:
- I moved a Dataframe function `createSeriesMetadata` from NetworkCUtil to CTypeUtil, as it is not specific to Network and i needed it to add dynawaltz mapping with dataframes.
- I also modified some typings in `dataframes.py` as I called them with enum and not standard python types. And added `_add_index_to_kwargs` that creates an explicit index for a kwargs that will be used as Dataframe.

geofjamg (approved):

PR #527, opened 2022-10-27
add identifiable, injection, branch and terminal dataframes in network

### LINKED ISSUE #522: Dataframes of generic types of network elements (identifiables, injections, branches, terminals)

Feature

When we want to get for example a list of all injections, we need to call multiple methods and concatenate the resulting dataframes.
For example, it's pretty hard to get the list of all elements connected to a specific bus.

Instead the API could provide native methods to get common information of a generic type:
- `get_identifiables`: all identifiables, with only their id
- `get_injections`: all injections with voltage_level_id, bus_id
- `get_branches`: all branches with voltage_level1_id, bus1_id, voltage_level2_id, bus2_id
- `get_terminals`: the list of all terminals, described by their element_id, element_side, voltage_level_id, bus_id

In particular, the `get_terminals` method provides a practical view to the users of those objects which are very important in the java implementation.

EtienneLt:

#522


java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
```diff
@@ -860,6 +864,50 @@ static String getPhaseTapChangerRegulatedSide(TwoWindingsTransformer transformer
...
+                NetworkDataframes::getT2OrThrow)
+                .stringsIndex("id", Connectable::getId)
+                .strings("type", connectable -> connectable.getType().toString())
+                .strings("voltage_level_id", connectable -> ((Injection) connectable).getTerminal().getVoltageLevel().getId())
+                .strings("bus_id", connectable -> ((Injection) connectable).getTerminal().getBusView().getBus() == null ? "" :
+                        ((Injection) connectable).getTerminal().getBusView().getBus().getId())
+                .build();
+    }
+
+    static NetworkDataframeMapper branches() {
```
EtienneLt, line 888: Maybe we should add hvdc line and the legs of the 3 windings transformers ? @sylvlecl

sylvlecl: Not really in favor of it because in all other places in powsybl branches are only lines + 2 windings transformers (elements that connect 2 terminals). Or we need to give it a different name ?

The 3 windings transformers will not fit well since there will be 3 voltages levels.

EtienneLt: Do we create another dataframe with 2 windings, 3 windings legs, hvdc and lines ? @annetill @sylvlecl

annetill: I understand for three windings transformers even if my experience is that sometimes I really miss something really simple that gives me the buses and the edges, as the simplest model for calculations. For HVDC lines, I think it is questionnable not to have it. I know that we never put them in branches in Java. But we are in pypowsybl and we have to ease network analysis.

EtienneLt: to conclude what do I do ? I let a dataframe with the branches of lines and 2 windings transformers ? and no other dataframes ? @annetill @sylvlecl

geofjamg: yes

geofjamg (approved):

PR #530, opened 2022-10-30
Avoid releasing null_ptr Java handle

geofjamg:
Signed-off-by: Geoffroy Jamgotchian <geoffroy.jamgotchian@gmail.com>

No

Optimization

Even if it seems to be correctly supported on Java side by GraalVM SDK, we  should avoid releasing `JavaHandle` built from a `null_ptr`. We have this case many times because of reporter java handle built from `null_ptr`.

**Other information**:

### DIFF
--- a/cpp/src/pypowsybl.cpp
+++ b/cpp/src/pypowsybl.cpp
@@ -109,7 +109,9 @@ T callJava(F f, ARGS... args) {
 //Destruction of java object when the shared_ptr has no more references
 JavaHandle::JavaHandle(void* handle):
     handle_(handle, [](void* to_be_deleted) {
-        callJava<>(::destroyObjectHandle, to_be_deleted);
+        if (to_be_deleted) {
+            callJava<>(::destroyObjectHandle, to_be_deleted);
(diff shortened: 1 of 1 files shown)


sylvlecl:
I think that most of the time (always ?), the `JavaHandle` of the reporter itself is null, not the `void*` inside it, no?

geofjamg:
> I think that most of the time (always ?), the `JavaHandle` of the reporter itself is null, not the `void*` inside it, no?

I did some more tests and I confirm thart on Linux and MacOS `(reporter == nullptr) ? nullptr : *reporter` always end up to a creation of `JavaHandle` with a null_ptr inside and an extra not needed call to Java object release.
Maybe we can also try to improve `(reporter == nullptr) ? nullptr : *reporter` to avoid this, but in anycase as it is often hard to figureout how objects are built using implicit contructor, it is safer to check for null pointer.

sylvlecl:
Indeed we should remove the implicit conversions after all.

sylvlecl (approved):

PR #531, opened 2022-11-02
expose aliases

### LINKED ISSUE #526: Expose network elements aliases

Feature

- Aliases of network identifiables are not available in the python API.
- We cannot either add new aliases
- However, they can naturally be used to define elements to be updated, since getter methods of the `Network` allows the use of ID or aliases seemlessly.
- Filtering rows by aliases should also already work naturally for the same reason

We should be able to query existing aliases, and add new ones.

_Queries:_
I propose to have a `get_aliases()` method returning a dataframe for all identifiables with following columns:
- id (index)
- alias
- alias_type

_Creation:_
Method `create_aliases` taking as input the same dataframe format as above.

_Removal:_
Method `remove_aliases` taking as input a dataframe with 2 columns: the id and the alias.
(or only the alias ?)

EtienneLt:

#526


EtienneLt:
doc is missing

java/src/main/java/com/powsybl/dataframe/network/adders/AliasDataframeAdder.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        void create(Network network, int row) {
+            Identifiable identifiable = network.getIdentifiable(ids.get(row));
+            if (identifiable == null) {
+                throw new PowsyblException("identifiable " + ids.get(row) + " does not exist");
+            }
+            if (aliasTypes != null) {
+                String type = aliasTypes.get(row);
+                if (type != null) {
```
sylvlecl: Here it will never be null in practice, I think we should check for empty instead

pypowsybl/network.py
```diff
@@ -3984,6 +3997,36 @@ def create_curve_reactive_limits(self, df: _DataFrame, **kwargs: _ArrayLike) ->
...
+
+            Valid attributes are:
+
+            - **id**:    the identifier of the network element associated to the alias
+            - **alias**:     name of the alias
+            - **alias_type**:     type of the alias (optional)
+
+        Examples:
+
+            .. code-block:: python
```
sylvlecl, line 4024: I think it will not render correctly without a blank line

sylvlecl: We need to add the methods to the API reference in sphinx doc

sylvlecl (approved):

PR #541, opened 2022-11-22
Get sld method takes some layout params; returns also sld metadata 

CBiasuzzi:
Signed-off-by: Christian Biasuzzi <christian.biasuzzi@soft.it>

feature

adds four layout parameters to the get_single_line_diagram (use_name, center_name, diagonal_label and the choice to use a nominal voltage style provider, instead of the default topological style provider).
The method adds to the returned Svg object also the sld metatadata, to be possibly used by some visualization tools developed in python

**Other information**:


pypowsybl/network.py
```diff
@@ -318,17 +323,22 @@ def write_single_line_diagram_svg(self, container_id: str, svg_file: PathOrStr)
...
         _pp.write_single_line_diagram_svg(self._handle, container_id, svg_file)
 
-    def get_single_line_diagram(self, container_id: str) -> Svg:
+    def get_single_line_diagram(self, container_id: str, use_name: bool = False, center_name: bool = False, diagonal_label: bool = False, topological_coloring: bool = True) -> Svg:
         """
         Create a single line diagram from a voltage level or a substation.
 
         Args:
             container_id: a voltage level id or a substation id
+            use_name: (layout parameter) use names instead of ids in labels
```
sylvlecl: Your documentation makes me think: maybe we could group those parameters in a small class `LayoutParameters` ?
(only on the python side, not all the way to java)

Maybe this would scale better if we add other parameters in the future, or other arguments to the method (like custom labels).

pypowsybl/network.py
```diff
@@ -318,17 +323,22 @@ def write_single_line_diagram_svg(self, container_id: str, svg_file: PathOrStr)
...
+            center_name: (layout parameter) center labels
+            diagonal_label: (layout parameter) display diagonal labels
+            topological_coloring: (layout parameter) when False, use a nominal voltage style provider (instead of the default topological style provider)
```
sylvlecl: Note: we could formulate it more naturally for the python user, like "when False, coloring is based only on nominal voltage"?

pypowsybl/network.py
CBiasuzzi: that's a great suggestion!  done.

pypowsybl/network.py
CBiasuzzi: Thx for the better wording!, done.

sylvlecl:
@CBiasuzzi : no agreement, after all, about the deletion of the method to write SVG to file:
let's do as you proposed and add layout parameters and an optional path for the metadata to that method !

CBiasuzzi:
> @CBiasuzzi : no agreement, after all, about the deletion of the method to write SVG to file: let's do as you proposed and add layout parameters and an optional path for the metadata to that method !

I added the parameters and the optional metadata file to the write_svg method; I also tried to wrap the layout parameters in a structure, in the c->c++->java layers, to make adding additional layout options easier, in the future

sylvlecl (approved):

PR #545, opened 2022-11-29
[Network modifications] Create voltage level topology from scratch.

colineplqt:
Signed-off-by: Coline PILOQUET <coline.piloquet@rte-france.com>

No

Wraps CreateVoltageLevelTopology from powsybl-core. It allows to create a symmetrical voltage level from scratch, by specifying the number of busbar sections, the number of sections and the switch kind between sections.

**Other information**:
For now, I did not write the documentation. I will create a dedicated PR with fixes, and documentation about all the network modification as it will be clearer to explain. (branch creation, coupling device creation + this PR)


pypowsybl/_pypowsybl.pyi
```diff
@@ -499,3 +499,5 @@ def get_connectables_order_positions(network: JavaHandle, voltage_level_id: str)
 def get_unused_order_positions(network: JavaHandle, busbar_section_id: str, before_or_after: str) -> List[int]: ...
 def remove_aliases(network: JavaHandle, dataframe: Dataframe) -> None: ...
 def close() -> None: ...
+def get_voltage_level_topology_creation_metadata() -> List[SeriesMetadata]: ...
+def create_voltage_level_topology(network: JavaHandle, dataframe: Dataframe, switch_kind: List[str], raise_exception: bool, reporter: Optional[JavaHandle]) -> None: ...
```
colineplqt: raise_exception should be optional

java/src/main/java/com/powsybl/dataframe/network/adders/VoltageLevelTopologyCreationSeries.java
```diff
@@ -0,0 +1,35 @@
+/**
+ * Copyright (c) 2022, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
EtienneLt: could you add "SPDX-License-Identifier: MPL-2.0"

java/src/main/java/com/powsybl/python/network/NetworkModificationsCFunctions.java
```diff
@@ -226,6 +230,38 @@ public static PyPowsyblApiHeader.ArrayPointer<CIntPointer> getUnusedConnectableO
 
     }
 
+    @CEntryPoint(name = "createVoltageLevelTopology")
+    public static void createVoltageLevelTopology(IsolateThread thread, ObjectHandle networkHandle,
```
EtienneLt: we should not create a new method to get metadata each time we add a new method. it would be great if it can be refactored

java/src/main/java/com/powsybl/dataframe/network/adders/VoltageLevelTopologyCreationSeries.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.List;
+
+/**
+ * @author Coline PILOQUET <coline.piloquet at rte-france.com>
+ */
+public final class VoltageLevelTopologyCreationSeries {
+
+    private static final List<SeriesMetadata> METADATA = List.of(
+            SeriesMetadata.stringIndex("voltage_level_id"),
```
EtienneLt: maybe just id ? what do you think ?

tests/test_network.py
```diff
@@ -2012,5 +2012,28 @@ def test_aliases():
     assert n.get_aliases().empty
 
 
+def test_voltage_level_topology_creation():
+    network = pp.network.create_four_substations_node_breaker_network()
+    voltage_levels = pd.DataFrame.from_records(index='id', data=[
+        {'substation_id': 'S1', 'id': 'VL1', 'topology_kind': 'NODE_BREAKER', 'nominal_v': 225}
+    ])
+    network.create_voltage_levels(voltage_levels)
```
EtienneLt: I prefer the test to be with
```python
network.create_voltage_levels(id='VL1', substation_id='S1', topology_kind= 'NODE_BREAKER', ...)
```
because I think users would prefer to do this way

tests/test_network.py
```diff
@@ -2012,5 +2012,28 @@ def test_aliases():
...
+    switch = ['BREAKER']
+    pp.network.create_voltage_level_topology(network, switch, df)
+
+
+def test_voltage_level_topology_creation_from_kwargs():
+    network = pp.network.create_four_substations_node_breaker_network()
+    voltage_levels = pd.DataFrame.from_records(index='id', data=[
+        {'substation_id': 'S1', 'id': 'VL1', 'topology_kind': 'NODE_BREAKER', 'nominal_v': 225}
+    ])
+    network.create_voltage_levels(voltage_levels)
```
EtienneLt: same here

EtienneLt:
Could you add new methods in network.rst ? I think it is missing

java/src/main/java/com/powsybl/python/network/NetworkModificationsCFunctions.java
```diff
@@ -226,6 +230,38 @@ public static PyPowsyblApiHeader.ArrayPointer<CIntPointer> getUnusedConnectableO
...
+            builder.withSwitchKinds(switchKind);
+            NetworkModification modification = builder.build();
+            modification.apply(network, throwException, reporter == null ? Reporter.NO_OP : reporter);
+        });
+    }
+
+    public static CreateVoltageLevelTopologyBuilder createBuilder(UpdatingDataframe dataframe) {
+        CreateVoltageLevelTopologyBuilder builder = new CreateVoltageLevelTopologyBuilder();
+        for (int row = 0; row < dataframe.getRowCount(); row++) {
+            applyIfPresent(dataframe.getStrings("voltage_level_id"), row, builder::withVoltageLevelId);
```
colineplqt: TODO: fix

colineplqt:
@EtienneLt, new design for all the network modifications. What do you think? For now I only put the creation of a voltage level but it should apply to all the modifications.

EtienneLt:
> @EtienneLt, new design for all the network modifications. What do you think? For now I only put the creation of a voltage level but it should apply to all the modifications.

I think it is quite good, just I think we should add subfolders in java/src/main/java/com/powsybl/dataframe/network/adders/ something like modification and creation (if you have a better idea)
There minor details to change too

java/src/main/java/com/powsybl/dataframe/DataframeNetworkModificationType.java
```diff
@@ -0,0 +1,15 @@
+/**
+ * Copyright (c) 2022, RTE (http://www.rte-france.com)
```
EtienneLt: I dont know if it is 2023 or if we keep the opening's year of the pr

EtienneLt:
There should not be any change in cpp/lib/pybind11

java/src/main/java/com/powsybl/dataframe/DataframeNetworkModificationType.java
EtienneLt: ok that is not a problem

java/src/main/java/com/powsybl/dataframe/network/modifications/VoltageLevelTopologyCreation.java
```diff
@@ -0,0 +1,76 @@
...
+        List<SwitchKind> switchKinds = null;
+        for (int row = 0; row < dataframe.getRowCount(); row++) {
+            applyIfPresent(dataframe.getStrings("id"), row, builder::withVoltageLevelId);
+            applyIfPresent(dataframe.getInts("low_busbar_index"), row, builder::withLowBusbarIndex);
+            applyIfPresent(dataframe.getInts("busbar_count"), row, builder::withBusbarCount);
+            applyIfPresent(dataframe.getInts("low_section_index"), row, builder::withLowSectionIndex);
+            applyIfPresent(dataframe.getInts("section_count"), row, builder::withSectionCount);
+            applyIfPresent(dataframe.getStrings("busbar_section_prefix_id"), row, builder::withBusbarSectionPrefixId);
+            applyIfPresent(dataframe.getStrings("switch_prefix_id"), row, builder::withSwitchPrefixId);
+            switchKinds = getSwitchKinds(dataframe.getStringValue("switch_kinds", row).orElse(""));
```
EtienneLt, line 57: there is no problem when "switch_kinds" is null ?

cpp/src/bindings.cpp
```diff
@@ -204,6 +204,9 @@ PYBIND11_MODULE(_pypowsybl, m) {
             .value("TWTS", validation_type::TWTS)
             .value("TWTS3W", validation_type::TWTS3W);
 
+    py::enum_<network_modification_type>(m, "NetworkModificationType")
+            .value("VOLTAGE_LEVEL_TOPOLOGY_CREATION", network_modification_type::VOLTAGE_LEVEL_TOPOLOGY_CREATION);
```
EtienneLt, line 208: this enum will have all the other hypothesis ?

colineplqt: Yes!

java/src/main/java/com/powsybl/python/network/NetworkCFunctions.java
```diff
@@ -972,4 +967,14 @@ public static void removeFeederBays(IsolateThread thread, ObjectHandle networkHa
         });
     }
 
+    @CEntryPoint(name = "getModificationMetadata")
```
sylvlecl: I guess it should go into the other file `NetworkModificationsCFunctions`

java/src/main/java/com/powsybl/dataframe/DataframeNetworkModificationType.java
```diff
@@ -0,0 +1,15 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.dataframe;
+
+/**
+ * @author Coline Piloquet <coline.piloquet at rte-france.com>
+ */
+public enum DataframeNetworkModificationType {
```
sylvlecl: probably better to move it to the network modifications subfolder

java/src/main/java/com/powsybl/dataframe/network/modifications/VoltageLevelTopologyCreation.java
```diff
@@ -0,0 +1,72 @@
...
+    );
+
+    @Override
+    public List<SeriesMetadata> getMetadata() {
+        return METADATA;
+    }
+
+    private CreateVoltageLevelTopologyBuilder createBuilder(UpdatingDataframe dataframe, int row) {
+        CreateVoltageLevelTopologyBuilder builder = new CreateVoltageLevelTopologyBuilder();
+        String switchKindStr = dataframe.getStringValue("switch_kinds", row).isEmpty() ||
```
sylvlecl: we should avoid to call `getStringValue` 3 times, only one is necessary (there can be a small performed for each call)

pypowsybl/network.py
```diff
@@ -5327,5 +5342,53 @@ def replace_tee_point_by_voltage_level_on_line(network: Network, tee_point_line1
...
+
+    Examples:
+
+    .. code-block:: python
+        pp.network.create_voltage_level_topology(network=network, raise_exception=True, id='VL', busbar_count=3,
+                                                 section_count=3, switch_kinds='BREAKER, DISCONNECTOR')
+    """
+    metadata = _pp.get_network_modification_metadata(NetworkModificationType.VOLTAGE_LEVEL_TOPOLOGY_CREATION)
+    df = _adapt_df_or_kwargs(metadata, df, **kwargs)
+    for i in range(len(df)):
```
sylvlecl: This kind of for loop can be VERY slow in python, we should try to avoid it as much as possible.
Unfortunately I am not sure you can apply a python method on each element in a really more efficient way.
It might be slightly better to use [Series.transform](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.transform.html) at least, but maybe not a big difference.

Since this network modification will never be a performance bottleneck, I think, let's keep it this way if there is no other solution.

sylvlecl (commented):
Minor suggestions on my side.
It's nice to have implemented it in a way that will allow to easily add more modifications!

sylvlecl (approved):
(approving since previous suggestions are not must-do, actually)

pypowsybl/network.py
EtienneLt: maybe with map method I can do something

PR #553, opened 2023-01-05
Allowing to modify load flow validation parameters

### LINKED ISSUE #286: [Loadflow validation] Add parameters to the API

Feature.

In its first version, the LF validation integration will not allow to define validation config in python.
```python
pp.loadflow.run_validation(network, ValidationType.ALL)
```

We should be able to override the validation configuration with a python object.
```python
parameters = pp.loadflow.ValidationParameters()
parameters.threshold = 1.0
pp.loadflow.run_validation(network, ValidationType.ALL, parameters=parameters)
```

HugoKulesza:

**Does this PR already have an issue describing the problem ?** 

Issue #286


cpp/src/bindings.cpp
```diff
@@ -315,6 +315,20 @@ PYBIND11_MODULE(_pypowsybl, m) {
             .def_readwrite("provider_parameters_keys", &pypowsybl::LoadFlowParameters::provider_parameters_keys)
             .def_readwrite("provider_parameters_values", &pypowsybl::LoadFlowParameters::provider_parameters_values);
 
+    py::class_<pypowsybl::LoadFlowValidationParameters>(m, "LoadFlowValidationParameters")
+            .def(py::init(&pypowsybl::createValidationConfig))
+            .def_readwrite("threshold", &pypowsybl::LoadFlowValidationParameters::threshold)
+            .def_readwrite("verbose", &pypowsybl::LoadFlowValidationParameters::verbose)
+            .def_readwrite("load_flow_name", &pypowsybl::LoadFlowValidationParameters::load_flow_name)
```
EtienneLt: not important but to harmonise with all the code it should be loadflow_name

cpp/src/bindings.cpp
```diff
@@ -315,6 +315,20 @@ PYBIND11_MODULE(_pypowsybl, m) {
...
+            .def_readwrite("epsilon_x", &pypowsybl::LoadFlowValidationParameters::epsilon_x)
+            .def_readwrite("apply_reactance_correction", &pypowsybl::LoadFlowValidationParameters::apply_reactance_correction)
+            .def_readwrite("load_flow_parameters", &pypowsybl::LoadFlowValidationParameters::load_flow_parameters)
```
EtienneLt: loadflow_parameters

cpp/src/bindings.cpp
```diff
@@ -335,7 +349,8 @@ PYBIND11_MODULE(_pypowsybl, m) {
     m.def("run_load_flow", &pypowsybl::runLoadFlow, "Run a load flow", py::call_guard<py::gil_scoped_release>(),
           py::arg("network"), py::arg("dc"), py::arg("parameters"), py::arg("provider"), py::arg("reporter"));
 
-    m.def("run_load_flow_validation", &pypowsybl::runLoadFlowValidation, "Run a load flow validation", py::arg("network"), py::arg("validation_type"));
+    m.def("run_load_flow_validation", &pypowsybl::runLoadFlowValidation, "Run a load flow validation", py::arg("network"),
```
EtienneLt: run_loadflow_validation

cpp/src/pypowsybl-api.h
```diff
@@ -56,6 +56,21 @@ typedef struct load_flow_parameters_struct {
     int provider_parameters_values_count;
 } load_flow_parameters;
 
+typedef struct load_flow_validation_parameters_struct {
```
EtienneLt: here too

EtienneLt: and same for after

pypowsybl/sensitivity.py
```diff
@@ -145,35 +145,35 @@ class Parameters:  # pylint: disable=too-few-public-methods
     .. currentmodule:: pypowsybl.sensitivity
 
     Args:
-        load_flow_parameters: parameters that are common to loadflow and sensitivity analysis
+        loadflow_parameters: parameters that are common to loadflow and sensitivity analysis
```
sylvlecl: We need to be careful with such renamings in the python API:
it will break existing user code.

So if we really want to change it, we need to identify it as a breaking change in the release note.

I would prefer not to mix it with this PR which is only about adding a new feature, well separated from this part of the code.

tests/test_loadflow.py
```diff
@@ -120,6 +120,40 @@ def test_validation_all():
...
+    for attribute, values in attributes.items():
+        for value in values:
+            parameters = lf.ValidationParameters(**dict([(attribute, value)]))
+            assert value == getattr(parameters, attribute)
+
+            parameters = lf.ValidationParameters()
+            setattr(parameters, attribute, value)
+            assert value == getattr(parameters, attribute)
+
+    n = pp.network.create_four_substations_node_breaker_network()
```
sylvlecl, line 165: I would advise to separate in 2 tests the getters/setters and the test of the behaviour below

sylvlecl (changes requested):
Minor comments.
In addition, we need to update the documentation (loadflow.rst in "reference"), so that the ValidationConfig class appears in the documentation on dedicated page (https://pypowsybl.readthedocs.io/en/stable/reference/loadflow.html).

pypowsybl/sensitivity.py
HugoKulesza: Should I then undo all the renaming I did that does not concern the validation parameters (c++ and python), or only the ones in python that change the way the user can call parameters ?

EtienneLt: only undo those that were before

sylvlecl: We can keep the changes that are not part of the python API, yes :)

EtienneLt: I'm not sure but those in sensitivity api are visible by the user aren't they ?

sylvlecl (approved):

Issue #558, opened 2023-01-13
Expose secondary voltage control extensions


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the expected behavior?**

We should be able to manipulate secondary voltage control extension introduced in https://github.com/powsybl/powsybl-core/pull/2436 

* **What is the motivation / use case for changing the behavior?**

Being able to simulate secondary voltage control with open loadflow.

* **Implementation notes**

Things that may need some design work:
- it's an extension of `Network`, is it an issue with the existing mechanism ?
- the extension has nested objects: the natural representation would be multiple tables.
  Should we implement multiple "DataframeProvider", instead of juste one, for this extension ?

EtienneLt:
done with #564

PR #564, opened 2023-02-03
Allowing multiple dataframes extensions + exposing secondary voltage level extension  

HugoKulesza:

Issue #558


java/src/main/java/com/powsybl/dataframe/network/extensions/AbstractSingleDataframeNetworkExtension.java
```diff
@@ -0,0 +1,31 @@
+/**
+ * Copyright (c) 2021-2022, RTE (http://www.rte-france.com)
```
EtienneLt: 2023

java/src/main/java/com/powsybl/dataframe/network/extensions/AbstractSingleDataframeNetworkExtension.java
```diff
@@ -0,0 +1,31 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
```
EtienneLt, line 7: * SPDX-License-Identifier: MPL-2.0

java/src/main/java/com/powsybl/dataframe/network/extensions/DataframeKey.java
```diff
@@ -0,0 +1,48 @@
+/**
+ * Copyright (c) 2021-2022, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
EtienneLt: same

java/src/main/java/com/powsybl/dataframe/network/extensions/DataframeKey.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package com.powsybl.dataframe.network.extensions;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * @author Hugo Kulesza <hugo.kulesza@rte-france.com>
+ */
+public class DataframeKey {
```
EtienneLt: it is only used for extension ? maybe it should have a more explicit name ?

java/src/main/java/com/powsybl/dataframe/network/extensions/SecondaryVoltageControlDataframeAdder.java
```diff
@@ -0,0 +1,105 @@
+/**
+ * Copyright (c) 2021-2022, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
```
EtienneLt, line 7: same for all new headers

java/src/main/java/com/powsybl/dataframe/network/extensions/SecondaryVoltageControlDataframeAdder.java
```diff
@@ -0,0 +1,105 @@
...
+                    }
+                }
+
+                SecondaryVoltageControl.ControlZone newZone = new SecondaryVoltageControl.ControlZone(
+                        name,
+                        pilotPoint,
+                        controlUnits
+                );
+                adder.addControlZone(newZone);
+            }
```
EtienneLt, line 89: should we check if zoneCount is equal to 0 ?

java/src/main/java/com/powsybl/dataframe/network/extensions/SecondaryVoltageControlDataframeProvider.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public String getExtensionName() {
+        return SecondaryVoltageControl.NAME;
+    }
+
+    @Override
+    public ExtensionInformation getExtensionInformation() {
+        return new ExtensionInformation(SecondaryVoltageControl.NAME,
+                "Provides information about the secondary voltage control zones and units",
+                "index : ");
```
EtienneLt: I think it would be great to add that 2 dataframes are needed and their attributes definitions

java/src/main/java/com/powsybl/dataframe/network/extensions/AbstractSingleDataframeNetworkExtension.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Optional;
+import java.util.List;
+
+/**
+ * @author Hugo Kulesza <hugo.kulesza@rte-france.com>
+ */
+public abstract class AbstractSingleDataframeNetworkExtension {
+
+    public List<Optional<String>> getExtensionTableNames() {
+        return new ArrayList<>();
```
EtienneLt, line 22: maybe Collections.emptyList() is better ? @sylvlecl ?

java/src/main/java/com/powsybl/dataframe/network/extensions/ExtensionDataframeKey.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * @author Hugo Kulesza <hugo.kulesza@rte-france.com>
+ */
+public class ExtensionDataframeKey {
+    private String extensionName;
+    private Optional<String> tableName;
```
sylvlecl: The common practice is to use `Optional` only as a return type, and simply use "nullable" values in function argument and field classes. (see sonar rule about it). Basically wrapping in an optional is useful mostly for chaining methods on the return value.
--> let's keep with this practice by removing the optional here and in the constructor

java/src/main/java/com/powsybl/dataframe/network/extensions/ExtensionDataframeKey.java
```diff
@@ -0,0 +1,49 @@
...
+package com.powsybl.dataframe.network.extensions;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * @author Hugo Kulesza <hugo.kulesza@rte-france.com>
+ */
+public class ExtensionDataframeKey {
+    private String extensionName;
```
sylvlecl: all fields can be `final`

java/src/main/java/com/powsybl/dataframe/network/extensions/ExtensionDataframeKey.java
```diff
@@ -0,0 +1,49 @@
...
+    private Optional<String> tableName;
+
+    private int hashCode;
+
+    public ExtensionDataframeKey(String extensionName, Optional<String> tableName) {
+        this.extensionName = extensionName;
```
sylvlecl: we can add a non-null check here (`Objects.requireNonNull`)

java/src/main/java/com/powsybl/dataframe/network/extensions/ExtensionDataframeKey.java
```diff
@@ -0,0 +1,49 @@
...
+        this.tableName = tableName;
+        this.hashCode = Objects.hash(extensionName, tableName);
```
sylvlecl, line 24: it's not usual to cache it in a field of the class, but why not! (I guess there is a kind of tradeoff between memory used by the class and performance, and actually this class is really meant to be used as a key in hashmap, so it makes sense)

java/src/main/java/com/powsybl/dataframe/network/extensions/AbstractSingleDataframeNetworkExtension.java
```diff
@@ -0,0 +1,29 @@
...
+import com.powsybl.dataframe.network.NetworkDataframeMapper;
+
+import java.util.*;
+
+/**
+ * @author Hugo Kulesza <hugo.kulesza@rte-france.com>
+ */
+public abstract class AbstractSingleDataframeNetworkExtension implements NetworkExtensionDataframeProvider {
+
+    public List<Optional<String>> getExtensionTableNames() {
```
sylvlecl: In collection classes too it's a little unusual to use optionals, I would prefer plain strings instead. Actually in this list in particular, I think we should not have null values at all.
(Note that it's possible that some map implementations don't support null key, but I think for ex. `HashMap` does)

java/src/main/java/com/powsybl/dataframe/network/extensions/SecondaryVoltageControlDataframeProvider.java
```diff
@@ -0,0 +1,115 @@
...
+        network.removeExtension(SecondaryVoltageControl.class);
+    }
+
+    @Override
+    public NetworkElementAdder createAdder() {
+        return new SecondaryVoltageControlDataframeAdder();
+    }
+
+    private class ControlUnitWithZone {
+        private SecondaryVoltageControl.ControlUnit unit;
```
sylvlecl: fields can be final

java/src/main/java/com/powsybl/dataframe/network/extensions/SecondaryVoltageControlDataframeProvider.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public Map<Optional<String>, NetworkDataframeMapper> createMappers() {
+        Map<Optional<String>, NetworkDataframeMapper> mappers = new HashMap<>();
+        mappers.put(Optional.of("zones"),
+                NetworkDataframeMapperBuilder.ofStream(this::zonesStream)
+                        .stringsIndex("name", ControlZone::getName)
+                        .doubles("target_v", zone -> zone.getPilotPoint().getTargetV(), (zone, v) -> zone.getPilotPoint().setTargetV(v))
+                        .strings("bus_ids", zone -> zone.getPilotPoint().getBusbarSectionsOrBusesIds().get(0)) // TODO : parser liste ids
```
sylvlecl: indeed there is a small todo here :)
just joining the strings using a "," should be OK

java/src/main/java/com/powsybl/dataframe/network/extensions/SecondaryVoltageControlDataframeAdder.java
```diff
@@ -0,0 +1,106 @@
...
+            this.unitId = unitsDf.getStrings("unit_id");
+            this.participate = unitsDf.getInts("participate");
+            this.unitZoneName = unitsDf.getStrings("zone_name");
+        }
+
+        void create(Network network) {
+            var adder = network.newExtension(SecondaryVoltageControlAdder.class);
+            for (int zone = 0; zone < zoneCount; zone++) {
+                String name = zoneName.get(zone);
+                SecondaryVoltageControl.PilotPoint pilotPoint = new SecondaryVoltageControl.PilotPoint(List.of(busIds.get(zone)),
```
sylvlecl: I think we can split on "," here to get a list of IDs

tests/test_network_extensions.py
```diff
@@ -452,6 +452,26 @@ def test_standby_automaton():
...
+                data=[('GEN', True, 'zone_test')])
+    n.create_extensions(extension_name, [zones_df, units_df])
+
+    e1 = n.get_extensions(extension_name, "zones").loc['zone_test']
+    assert e1.target_v == 400
+    assert e1.bus_ids == "NHV1"
+    e2 = n.get_extensions(extension_name, "units").loc['GEN']
+    assert e2.participate
+    assert e2.zone_name == 'zone_test'
+
```
sylvlecl, line 474: we could add a test to update the values of `target_v` and `participate` with `update_extensions`

java/src/main/java/com/powsybl/python/network/NetworkCFunctions.java
```diff
@@ -677,7 +680,7 @@ public static void updateNetworkElementsExtensionsWithSeries(IsolateThread threa
                                                                  PyPowsyblApiHeader.ExceptionHandlerPointer exceptionHandlerPtr) {
         doCatch(exceptionHandlerPtr, () -> {
             String name = CTypeUtil.toString(namePtr);
-            NetworkDataframeMapper mapper = NetworkDataframes.getExtensionDataframeMapper(name);
+            NetworkDataframeMapper mapper = NetworkDataframes.getExtensionDataframeMapper(name, Optional.empty());
```
sylvlecl: Actually we should also adapt this "update" part so that "update_extensions" also works with the additional table argument

sylvlecl (changes requested):
Nice work overall :)

2 kinds of comments:
- "cosmetic" suggestions
- additional feature to be implemented: the "update" part, and less importantly the parsing of comma separated bus IDs. Actually those 2 things could also be done in a further pull request, but this should not be too much code

java/src/main/java/com/powsybl/dataframe/network/extensions/ExtensionDataframeKey.java
```diff
@@ -0,0 +1,52 @@
...
+        this.extensionName = Objects.requireNonNull(extensionName);
+        this.tableName = tableName;
+        this.hashCode = Objects.hash(extensionName, tableName);
+    }
+
+    public String getExtensionName() {
+        return extensionName;
+    }
+
+    public String getTableName() {
```
EtienneLt: that's here where it should return an Optional I'll fix it

java/src/main/java/com/powsybl/dataframe/network/extensions/SecondaryVoltageControlDataframeProvider.java
```diff
@@ -0,0 +1,158 @@
...
+        return new SecondaryVoltageControlDataframeAdder();
+    }
+
+    private static class ControlUnitWithZone {
+        private final SecondaryVoltageControl.ControlUnit unit;
+        private final String zoneName;
+
+        public ControlUnitWithZone(SecondaryVoltageControl.ControlUnit unit, String zoneName) {
+            this.unit = unit;
+            this.zoneName = zoneName;
```
EtienneLt: maybe a check non null can be added ?

java/src/main/java/com/powsybl/dataframe/network/extensions/ExtensionDataframeKey.java
EtienneLt: indeed the method is not used I delete

sylvlecl (approved):

EtienneLt (approved):

PR #583, opened 2023-02-24
[Network modifications] Refactoring of modifications 

colineplqt:

No

Refactor all the network modifications to match the new design.
Breaking change on some as it now takes a dataframe as input


java/src/main/java/com/powsybl/dataframe/network/adders/VscStationDataframeAdder.java
```diff
@@ -39,10 +39,7 @@ public class VscStationDataframeAdder extends AbstractSimpleAdder {
             SeriesMetadata.doubles("target_v"),
             SeriesMetadata.doubles("target_q"),
             SeriesMetadata.doubles("loss_factor"),
-            SeriesMetadata.booleans("voltage_regulator_on"),
-            SeriesMetadata.strings("bus_or_busbar_section_id"),
-            SeriesMetadata.ints("position_order"),
-            SeriesMetadata.strings("direction")
```
EtienneLt, line 45: just to be sure these metadata are in the new dataframes ?

java/src/main/java/com/powsybl/dataframe/network/modifications/CreateLineOnLine.java
```diff
@@ -0,0 +1,101 @@
...
+        applyIfPresent(dataframe.getStrings("fictitious_substation_id"), row, builder::withFictitiousSubstationId);
+        applyIfPresent(dataframe.getStrings("fictitious_substation_name"), row, builder::withFictitiousSubstationName);
+        applyIfPresent(dataframe.getStrings("line1_id"), row, builder::withLine1Id);
+        applyIfPresent(dataframe.getStrings("line2_id"), row, builder::withLine2Id);
+        applyIfPresent(dataframe.getStrings("line1_name"), row, builder::withLine1Name);
+        applyIfPresent(dataframe.getStrings("line2_name"), row, builder::withLine2Name);
+        Optional<String> lineIdOptional = dataframe.getStringValue("line_id", row);
+        if (lineIdOptional.isEmpty()) {
+            throw new PowsyblException("LineId cannot be empty.");
+        }
```
EtienneLt: just for a tip you can do something like 
```java
dataframe.getStringValue("line_id", row).orElseThrow(() -> new PowsyblException("LineId cannot be empty."))
```

java/src/main/java/com/powsybl/dataframe/network/modifications/NetworkModifications.java
```diff
@@ -27,14 +27,22 @@ private NetworkModifications() {
...
+            Map.entry(REVERT_CONNECT_VOLTAGE_LEVEL_ON_LINE, new RevertConnectVoltageLevelOnLine()),
+            Map.entry(REPLACE_TEE_POINT_BY_VOLTAGE_LEVEL_ON_LINE, new ReplaceTeePointByVoltageLevelOnLine())
     );
 
     public static NetworkModification getModification(DataframeNetworkModificationType type) {
         return MODIFICATION.get(type);
     }
 
-    public static void applyModification(DataframeNetworkModificationType type, Network network, UpdatingDataframe dataframe, boolean throwException, ReporterModel reporter) {
+    public static void applyModification(DataframeNetworkModificationType type, Network network, List<UpdatingDataframe> dataframe, boolean throwException, ReporterModel reporter) {
```
EtienneLt, line 45: maybe dataframes now ?

pypowsybl/network.py
```diff
@@ -5329,37 +5366,37 @@ def get_unused_order_positions_after(network: Network, busbar_section_id: str) -
     return pd.Interval(left=positions[0], right=positions[1], closed='both')
 
 
-def replace_tee_point_by_voltage_level_on_line(network: Network, tee_point_line1: str, tee_point_line2: str,
-                                               tee_point_line_to_remove: str,
-                                               bbs_or_bus_id: str, new_line1_id: str, new_line2_id: str,
-                                               new_line1_name: str = None, new_line2_name: str = None) -> None:
+def replace_tee_point_by_voltage_level_on_line(network: Network,  df: _DataFrame = None, raise_exception: bool = False,
+                                               reporter: _Reporter = None, **kwargs: _ArrayLike) -> None:
```
EtienneLt: I see with the tests it will be a breaking change, won't it ?

java/src/main/java/com/powsybl/dataframe/network/adders/VscStationDataframeAdder.java
colineplqt: They are in ADDITIONNAL_METADATA in createFeederBay. It is only used for network modifications.

pypowsybl/network.py
colineplqt: Yes it does, I made it so that all the functions would take a dataframe as input and not arguments straight. So it is a breaking change, which is why I am not sure if it is a good idea...

colineplqt: The refactoring can be reused by creating the dataframes in the different functions but I felt like it made sense to have all the modifications taking dataframes as input but I am not sure.

EtienneLt: I agree and I think it is better this way but I'm not sure about the breaking change

colineplqt: Is there something like deprecated in Python?

EtienneLt: there seems to be something https://pypi.org/project/Python-Deprecated/

setup.py
```diff
@@ -108,6 +108,7 @@ def build_extension(self, ext):
         'numpy>=1.20.0',
         'pandas>=1.4.4; sys_platform == "darwin" and platform_machine == "arm64"',
         'pandas>=1.3.5; sys_platform != "darwin" or platform_machine != "arm64"',
-        'networkx'
+        'networkx',
+        'Deprecated'
```
EtienneLt: not sure but it seems to be called Python-Deprecated

EtienneLt (approved):

PR #584, opened 2023-03-03
enable to dis/connect terminals and branches

EtienneLt:

feature

enable to dis/connect terminals and branches with update methods in network

### DIFF
--- a/java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
+++ b/java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
@@ -297,7 +297,7 @@ public final class NetworkDataframes {
 
     static NetworkDataframeMapper buses() {
         return NetworkDataframeMapperBuilder.ofStream(n -> n.getBusView().getBusStream(),
-                getOrThrow((b, id) -> b.getBusView().getBus(id), "Bus"))
+                        getOrThrow((b, id) -> b.getBusView().getBus(id), "Bus"))
                 .stringsIndex("id", Bus::getId)
                 .strings("name", b -> b.getOptionalName().orElse(""))
                 .doubles("v_mag", Bus::getV, Bus::setV)
@@ -814,7 +814,7 @@ public final class NetworkDataframes {
 
     private static NetworkDataframeMapper rtcs() {
         return NetworkDataframeMapperBuilder.ofStream(network -> network.getTwoWindingsTransformerStream()
-                .filter(t -> t.getRatioTapChanger() != null), NetworkDataframes::getT2OrThrow)
+                        .filter(t -> t.getRatioTapChanger() != null), NetworkDataframes::getT2OrThrow)
                 .stringsIndex("id", TwoWindingsTransformer::getId)
                 .ints("tap", t -> t.getRatioTapChanger().getTapPosition(), (t, p) -> t.getRatioTapChanger().setTapPosition(p))
                 .ints("low_tap", t -> t.getRatioTapChanger().getLowTapPosition())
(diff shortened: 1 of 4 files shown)


java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
```diff
@@ -869,49 +869,105 @@ static String getPhaseTapChangerRegulatedSide(TwoWindingsTransformer transformer
...
 
     static NetworkDataframeMapper branches() {
-        return NetworkDataframeMapperBuilder.ofStream(Network::getBranchStream, NetworkDataframes::getT2OrThrow)
+        return NetworkDataframeMapperBuilder.ofStream(Network::getBranchStream, getOrThrow(Network::getBranch, "Branch"))
                 .stringsIndex("id", Branch::getId)
                 .strings("type", branch -> branch.getType().toString())
                 .strings("voltage_level1_id", branch -> branch.getTerminal1().getVoltageLevel().getId())
                 .strings("bus1_id", branch -> branch.getTerminal1().getBusView().getBus() == null ? "" :
                         branch.getTerminal1().getBusView().getBus().getId())
+                .booleans("connected_1", branch -> branch.getTerminal1().isConnected(),
```
sylvlecl: we should remove the underscore for consistency (`connected1`), same for side 2

java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
```diff
@@ -869,49 +869,105 @@ static String getPhaseTapChangerRegulatedSide(TwoWindingsTransformer transformer
...
+        if (connectable == null) {
+            throw new PowsyblException("connectable " + id + " not found");
+        }
+        String side = dataframe.getStringValue("element_side", index).orElse(null);
+        if (side == null) {
+            if (connectable instanceof Branch || connectable instanceof ThreeWindingsTransformer) {
+                throw new PowsyblException("side must be provided for this element");
+            }
+            return connectable.getTerminals().get(0);
+        } else if (side.equals("ONE")) {
```
sylvlecl: It would be better with an enum than with strings

java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
```diff
@@ -869,49 +869,105 @@ static String getPhaseTapChangerRegulatedSide(TwoWindingsTransformer transformer
...
+        String id = dataframe.getStringValue("element_id", index)
+                .orElseThrow(() -> new IllegalArgumentException("element_id column is missing"));
+        Connectable<?> connectable = network.getConnectable(id);
+        if (connectable == null) {
+            throw new PowsyblException("connectable " + id + " not found");
+        }
+        String side = dataframe.getStringValue("element_side", index).orElse(null);
+        if (side == null) {
+            if (connectable instanceof Branch || connectable instanceof ThreeWindingsTransformer) {
+                throw new PowsyblException("side must be provided for this element");
```
sylvlecl: We could add the id in the message so that the user knows which is the element

java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
```diff
@@ -869,49 +869,105 @@ static String getPhaseTapChangerRegulatedSide(TwoWindingsTransformer transformer
...
-                        ((Injection) connectable).getTerminal().getBusView().getBus().getId())
+                .strings("voltage_level_id", connectable -> ((Injection<?>) connectable).getTerminal().getVoltageLevel().getId())
+                .strings("bus_id", connectable -> ((Injection<?>) connectable).getTerminal().getBusView().getBus() == null ? "" :
+                        ((Injection<?>) connectable).getTerminal().getBusView().getBus().getId())
                 .build();
     }
 
     static NetworkDataframeMapper branches() {
-        return NetworkDataframeMapperBuilder.ofStream(Network::getBranchStream, NetworkDataframes::getT2OrThrow)
+        return NetworkDataframeMapperBuilder.ofStream(Network::getBranchStream, getOrThrow(Network::getBranch, "Branch"))
```
sylvlecl, line 893: Not part of this PR but it would be better with an enum here too

java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
EtienneLt: I'm not sure because there are enums for branch and enums for three windings transformers and if we add an enum here we will have more code

java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
EtienneLt: why do we need an enum here for ?

sylvlecl: Mainly so that people don't have to go and read the method implementation to know what are the possible values

sylvlecl (approved):

PR #599, opened 2023-04-19
Add support of hvdc sensi

flo-dup:

No

Bug fix

Sensitivity on HVDC line not supported

Sensitivity on HVDC line supported


obrix:
Seems okay, having a small test would be good and understand why the documentation was describing a working example with hvdc line would also be good. Maybe it was working in a different way (variable set) ? Or maybe there was a regression ?

flo-dup:
> Seems okay, having a small test would be good and understand why the documentation was describing a working example with hvdc line would also be good. Maybe it was working in a different way (variable set) ? Or maybe there was a regression ?

A test was added, thanks @EtienneLt! The result seems ok to me.
About the documentation, I only found this [here](https://pypowsybl.readthedocs.io/en/latest/user_guide/sensitivity.html#other-kind-of-sensitivities), not a working example. Was there another one?
I don't think it could have been through the variable set as in the code it means that the variable type is then `SensitivityVariableType.INJECTION_ACTIVE_POWER`.

obrix (approved):

EtienneLt (approved):

PR #604, opened 2023-05-11
add voltage interval to filter network area diagram

EtienneLt:

feature

allow to filter network area diagram according to nominal voltage


java/src/main/java/com/powsybl/python/network/NetworkAreaDiagramUtil.java
```diff
@@ -30,32 +31,136 @@ public final class NetworkAreaDiagramUtil {
...
+                         Double highNominalVoltageBound, Double lowNominalVoltageBound) {
         try (Writer writer = Files.newBufferedWriter(Paths.get(svgFile), StandardCharsets.UTF_8)) {
-            writeSvg(network, voltageLevelIds, depth, writer);
+            writeSvg(network, voltageLevelIds, depth, writer, highNominalVoltageBound, lowNominalVoltageBound);
         } catch (IOException e) {
             throw new UncheckedIOException(e);
         }
     }
+
+    static VoltageLevelFilter getNominalVoltageFilter(Network network, List<String> voltageLevelIds,
```
EtienneLt, line 70: @flo-dup @So-Fras here the code is taken and modified from powsybl diagram

java/src/main/java/com/powsybl/python/network/NetworkAreaDiagramUtil.java
```diff
@@ -30,32 +31,136 @@ public final class NetworkAreaDiagramUtil {
...
+        Set<VoltageLevel> startingSet = new HashSet<>();
+
+        for (String voltageLevelId : voltageLevelIds) {
+            VoltageLevel vl = network.getVoltageLevel(voltageLevelId);
+            if (vl == null) {
+                throw new PowsyblException("Unknown voltage level id '" + voltageLevelId + "'");
+            }
+            if (highNominalVoltageBound > 0) {
+                if (vl.getNominalV() < lowNominalVoltageBound || vl.getNominalV() > highNominalVoltageBound) {
+                    throw new PowsyblException("vl '" + voltageLevelId +
```
So-Fras: Do we really want to throw an exception here? Couldn't a log warn be enough?

java/src/main/java/com/powsybl/python/network/NetworkAreaDiagramUtil.java
```diff
@@ -30,32 +31,136 @@ public final class NetworkAreaDiagramUtil {
...
+            Set<VoltageLevel> nextDepthVoltageLevels = new HashSet<>();
+            for (VoltageLevel vl : voltageLevelsDepth) {
+                if (!visitedVoltageLevels.contains(vl)) {
+                    if (highNominalVoltageBound > 0) {
+                        if (vl.getNominalV() >= lowNominalVoltageBound
+                                && vl.getNominalV() <= highNominalVoltageBound) {
+                            visitedVoltageLevels.add(vl);
+                            vl.visitEquipments(new VlVisitor(nextDepthVoltageLevels, visitedVoltageLevels));
+                        }
+                    } else {
```
So-Fras, line 114: With these conditions, it seems to me that we do not allow users to give a lowNominalVoltageBound but no highNominalVoltageBound (by setting that variable to -1) whereas the opposite is possible. Am I right?

So-Fras (changes requested):
Thank you for your work! I left but a few comments.
I also have a general remark: at the moment, we will not see voltage levels that are within the depth and within the voltage bounds if they are beyond (depth-wise) a voltage level outside voltage bounds. Is it what users expect?

java/src/main/java/com/powsybl/python/network/NetworkAreaDiagramUtil.java
EtienneLt: well not sure if we select a initial voltage level with his nominal voltage out of the voltage bounds it won't work anyway so I think if we don't throw an error there will be an empty diagram

java/src/main/java/com/powsybl/python/network/NetworkAreaDiagramUtil.java
EtienneLt: Yes I am not sure what behaviour we want there, do we want to have a filter on only low or high bound ? we consider a bound not usuable if it is < 0 ?

java/src/main/java/com/powsybl/python/network/NetworkAreaDiagramUtil.java
So-Fras: I agree: if all your input voltage levels are out of bounds, you will end up with an empty diagram (and some warnings). But it may be harsch for users to get an exception for giving only one voltage level out of bounds.

java/src/main/java/com/powsybl/python/network/NetworkAreaDiagramUtil.java
So-Fras: I think that it may be useful to be able to filter only on low bound or only on high bound.

So-Fras (approved):

PR #609, opened 2023-05-30
BytesIO support for network loading

obrix:
Add bytesIO support for network loading. Only zipped network file are supported for now. Inside the zip the same formats as the load from file function are supported. Support of non zipped memory blob will be supported later, there is some modifications on powsybl-core to do to make it work.

**Other information**:


tests/test_network.py
```diff
@@ -45,6 +49,32 @@ def test_load_network_from_string():
     n = pp.network.load_from_string('simple-eu.uct', file_content)
     assert 1 == len(n.get_substations())
 
+def test_load_compressed():
+  with open(DATA_DIR.joinpath('simple-eu.uct'), "rb") as fh:
+      network_data = io.BytesIO(fh.read()).read()
+
+      bz2data = bz2.compress(network_data)
+      n = pp.network.load_from_binary_buffer('simple-eu.uct', io.BytesIO(bz2data))
```
geofjamg: @obrix do we really need to specify a name?

obrix: Not sure, It depends on format read. Still have to figure out the proper parameter to expose because :

* When importing CGMES from one or multiple zip, filename is not important and can be anything : import succeed. So not needed.
* But for data format such as single UCTE file a check is done by the importer to find a element with ".uct" extension in the ReadOnlyMemDataSource data hashmap. So it is not trivial to have a completely generic approach.

Maybe the solution is just to dedicate the multiple bytesIO api to CGMES zipped file import, with no mandatory filename. And then have a single bytesIO import API that has a mandatory file name parameter.

obrix: Or for the single byteIO api we could add a mandatory format parameter as an enum (UCTE, MAT, etc...). This would avoid the mandatory filename when only the extension is needed

So we could have : 

```
load_from_binary_buffer(byteBuffer, Format.UCTE)

load_cgmes_from_binary_buffer([byteBufferZip1, byteBufferZip2...])
```

obrix:
After discussing with @geofjamg it seems the issue of having to provide a filename for loading network needs to be handled correctly on powsybl-core side. The idea would be to base the file format detection on a partial file header reading instead of relying on finding a specific extension in the filename (like in UCTEimporter or MatPowerImporter).

This issue does not impact the loading of cgmes from zipped one or multiple zipfile as ByteIO.

So it has been decided to limit this development to zipped cgmes reading from ByteIOs (which was the main use case) and add the ByteIOs reading for other compression or uncompressed file when a proper format guessing will be introduced in powsybl core (probably next release).

java/src/main/resources/META-INF/native-image/com.github.luben/zstd-jni/index.json
```diff
@@ -0,0 +1,10 @@
+[
```
geofjamg, line 1: Why isn't it same kind of metadata description as the other?

java/src/main/java/com/powsybl/python/network/NetworkCFunctions.java
```diff
@@ -166,6 +163,46 @@ public static ObjectHandle loadNetworkFromString(IsolateThread thread, CCharPoin
...
+                                                           ExceptionHandlerPointer exceptionHandlerPtr) {
+        return doCatch(exceptionHandlerPtr, () -> {
+            Properties parameters = createParameters(parameterNamesPtrPtr, parameterNamesCount, parameterValuesPtrPtr, parameterValuesCount);
+            Reporter reporter = ObjectHandles.getGlobal().get(reporterHandle);
+            List<Integer> bufferSizes = CTypeUtil.toIntegerList(dataSizes, bufferCount);
+            List<ReadOnlyDataSource> dataSourceList = new ArrayList<>();
+            for (int i = 0; i < bufferCount; ++i) {
+                ByteBuffer buffer = CTypeConversion.asByteBuffer(data.read(i), bufferSizes.get(i));
+                Optional<CompressionFormat> format = detectCompressionFormat(buffer);
+                if (format.isPresent() && CompressionFormat.ZIP.equals(format.get())) {
```
geofjamg, line 178: We could also support the no compression case ?

java/src/main/java/com/powsybl/python/commons/Util.java
```diff
@@ -358,4 +368,62 @@ public static DataframeNetworkModificationType convert(PyPowsyblApiHeader.Networ
...
+        } else if (compareSignature(buffer, BZIP2_SIGNATURE)) {
+            return Optional.of(CompressionFormat.BZIP2);
+        } else if (compareSignature(buffer, ZSTD_SIGNATURE)) {
+            return Optional.of(CompressionFormat.ZSTD);
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    public static InputStream deCompressedInputStream(InputStream is, Optional<CompressionFormat> format) throws IOException {
```
geofjamg: To remove because not yet used

geofjamg (changes requested):
Maybe we can add a section in the documentatiobn with a small exemple to how read a case file from a a bytesIO ?

java/src/main/resources/META-INF/native-image/com.github.luben/zstd-jni/index.json
obrix: Without it I had some issues at native image building phase.

java/src/main/java/com/powsybl/python/network/NetworkCFunctions.java
obrix: I don't think so, you are going to run again on the initial problem of not having an extension to try to guess the format. With a zip format you can try to access the files inside with a basename but it is still the extension that is used to select an importer.

geofjamg: yes you are right

java/src/main/resources/META-INF/native-image/com.github.luben/zstd-jni/index.json
geofjamg: ok so this is a new way to define metadata

java/src/main/java/com/powsybl/python/commons/Util.java
obrix: done

geofjamg (approved):

EtienneLt:
@geofjamg @obrix it works only with a zipped network but not with a file from an archive ?

EtienneLt:
we should be able to load a network from bytes (io.Bytes ...) what do you think ?

obrix:
@EtienneLt Yes the issue is we cannot do this without also providing the format explicitly and setting a "filename" with a matching extension in the datasource. We talked with @geofjamg about reworking the way we guess the format in the core to be able to provide a byte array and let the importer guess the format.

PR #627, opened 2023-07-10
Voltage initializer interface with open reac impl

nicolas-pierr:

Add an interface to OpenReac.


pypowsybl/openreac.py
```diff
@@ -0,0 +1,67 @@
...
+            raise TypeError(f"error of length of {keys}, {values}")
+
+    def set_objective(self, objective: _pp.OpenReacObjective) -> None:
+        _pp.open_reac_set_objective(self._handle, objective)
+
+    def set_objective_distance(self, distance: float) -> None:
+        _pp.open_reac_set_objective_distance(self._handle, distance)
+
+
+class OpenReacResults:
```
obrix: Indicators and status are missing.

pypowsybl/openreac.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    """
+
+    def __init__(self, result_handle: _pp.JavaHandle) -> None:
+        self._handle = result_handle
+
+    def apply_all_modification(self, network: _Network) -> None:
```
obrix: I am wondering if we can be a bit more explicit with this API, like be able to list generated modifications explicitly. But I think this can be done later and probably need an evolution of modification wrapping.

java/src/main/java/com/powsybl/python/openreac/OpenReacCFunctions.java
```diff
@@ -0,0 +1,187 @@
...
+        OpenReacParameters params = ObjectHandles.getGlobal().get(paramsHandle);
+
+        logger().info("Launching an OpenReac run");
+        OpenReacResult result = OpenReacRunner.run(network, network.getVariantManager().getWorkingVariantId(), params,
+                new OpenReacConfig(debug), LocalComputationManager.getDefault());
+        logger().info("OpenReac run done.");
+        return ObjectHandles.getGlobal().create(result);
+    }
+
+    @CEnum("OpenReacObjective")
```
obrix: Can you put the Enums in PyPowsyblApiHeader.java with the others enums ?

java/src/main/java/com/powsybl/python/openreac/OpenReacCFunctions.java
```diff
@@ -0,0 +1,187 @@
...
+        @CEnumLookup
+        public static native OpenReacObjective fromCValue(int value);
+    }
+
+    @CEnum("OpenReacStatus")
+    public enum OpenReacStatus {
+        OK,
+        NOT_OK;
+
+        public static OpenReacStatus fromJavaEnum(com.powsybl.openreac.parameters.output.OpenReacStatus obj) {
```
obrix: Enum object are kept simple in other cases, can you write a static conversion function separated like its done in Util.java ?

tests/test_openreac.py
```diff
@@ -0,0 +1,28 @@
+#
```
obrix: You are a bit light regarding tests, can you add test for the open reac API itself and not just for parameters ?

EtienneLt:
Could you add some documentation in rst file in docs/user_guide and in docs/reference ?

pypowsybl/openreac.py
```diff
@@ -0,0 +1,72 @@
...
+        _pp.open_reac_apply_all_modifications(self._handle, network._handle)
+
+    def get_status(self) -> _pp.OpenReacStatus:
+        return _pp.open_reac_get_status(self._handle)
+
+    def get_indicators(self) -> Dict[str, str]:
+        return _pp.open_reac_get_indicators(self._handle)
+
+
+def run_open_reac(network: _Network, params: OpenReacParameters, debug: bool) -> OpenReacResults:
```
EtienneLt: I think "run" is enough for the name of the method because when used it will be pp.openreac.run()

pypowsybl/openreac.py
```diff
@@ -0,0 +1,72 @@
...
+        '''
+        for id in transformer_id:
+            _pp.open_reac_add_variable_two_windings_transformers(
+                self._handle, id)
+
+    def add_specific_voltage_limits(self, limits: Dict[str, Tuple[float, float]]) -> None:
+        '''
+        Indicate to OpenReac to override the network voltages limits. Use this if OpenReac cannot converge because of infeasibility.
+        keys are voltage ids, values are (lower limit, upper limit)
+        '''
```
EtienneLt: could you use the same template as in loadflow or other module for example 
 """
  Get list of parameters for the specified loadflow provider.

  Args:
     provider (str): the provider, if not specified the provider will be the default one.

  Returns:
      the list of provider's parameters
    """

tests/test_openreac.py
```diff
@@ -0,0 +1,28 @@
...
+    params.add_constant_q_generators(["gen1", "gen2"])
+    params.add_variable_two_windings_transformers(["twt1", "twt2"])
+
+    params.add_algorithm_param([("foo", "bar"), ("bar", "bar2")])
+    params.add_specific_voltage_limits({"vl_id": (0.5,1.2)})
+
+    params.set_objective(OpenReacObjective.SPECIFIC_VOLTAGE_PROFILE)
+    params.set_objective_distance(1.3)
+
+# can't test runner because it need ampl and knitro.
```
EtienneLt: maybe you can add integration test that does not run on pipelines

geofjamg (changes requested):
As a general comment, we should as for other analysis function use in the api, the function name and not the implementation (do we call the lf openloadflow or dynaflow). Here the function is "voltage initialization" and not openreac. Everywhere in the api we should use voltage initialization. These comment is also true on java side but as python api is user oriented and could not have any breaking change we could do it first on python api.

java/pom.xml
```diff
@@ -357,6 +357,11 @@
             <groupId>com.powsybl</groupId>
             <artifactId>powsybl-ampl-converter</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.powsybl</groupId>
+            <artifactId>powsybl-open-reac</artifactId>
+            <version>0.2.0</version>
```
EtienneLt: version should be in properties 
```xml
<openreac.version>0.2.0</openreac.version>
```

pypowsybl/voltage_initializer.py
```diff
@@ -0,0 +1,137 @@
+from typing import Dict, List, Tuple
+from pypowsybl import _pypowsybl as _pp
+from pypowsybl.network import Network as _Network
+
+
+class VoltageInitializerParameters:
+    """
+    """
```
EtienneLt: can you add a main comment on what does these class

pypowsybl/voltage_initializer.py
```diff
@@ -0,0 +1,137 @@
+from typing import Dict, List, Tuple
```
EtienneLt: it would be better for now (before refactoring) to import with a _ to hide this import for users
I mean 
```python
from typing import Dict as _Dict, List as _List, Tuple as _Tuple
```

tests/test_voltage_initializer.py
```diff
@@ -0,0 +1,50 @@
+#
+# Copyright (c) 2020-2022, RTE (http://www.rte-france.com)
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
```
EtienneLt: can you add 
SPDX-License-Identifier: MPL-2.0
and 2023

pypowsybl/voltage_initializer.py
EtienneLt: header is missing

EtienneLt: now that the refactoring is merged it should be good to do the same

java/pom.xml
geofjamg: Done

pypowsybl/voltage_initializer.py
geofjamg: Done

pypowsybl/voltage_initializer.py
geofjamg: Done

tests/test_voltage_initializer.py
geofjamg: Done

pypowsybl/voltage_initializer.py
geofjamg: Done

geofjamg (approved):

EtienneLt (approved):

PR #630, opened 2023-07-17
Add dev Dockerfile

marjanbozorg:

**Other information**:

### DIFF
--- a/pypowsybl.dev.dockerfile
+++ b/pypowsybl.dev.dockerfile
@@ -1,0 +1,27 @@
+FROM python:3.7-bookworm
+
+RUN apt-get update && apt-get install -y \
+    cmake \
+    gcc \
+ && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


pypowsybl.dev.dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM python:3.7-bookworm
```
EtienneLt, line 1: I think it would be better to use a more recent version of python.

marjanbozorg: To build pypowsybl from the sources mentioned that Python >= 3.7 for Linux, Windows and MacOS amd64 is required. That is the reason I put python 3.7

EtienneLt: Yes but I think in the future python3.7 will be deprecated because some important dependencies does not support python 3.7 any more (for example pandas)

PR #632, opened 2023-07-19
Modify powsybl-diagram apis

So-Fras:

**Other information**:


geofjamg:
@So-Fras could you please update this PR ?

geofjamg:
We usually used not to break python code compatibility, we should probably think about adding deprecated backward old API (the LayoutParameters) ?

So-Fras:
I understand. But I think it would be confusing to keep both, especially since the old LayoutParameters contained everything but layout parameters!

So-Fras:
In the end, we have not broken compatibility with previous versions. If users use the old LayoutParameters class, a warning is issued to encourage them to switch to the new SldParameters class.

geofjamg (approved):
@So-Fras it could be done in another PR but we should also improve the user guide by describing better SLD write parameters

So-Fras:
Agreed. I was also thinking of offering full customization possibilities for Python users and fully document parameters.

EtienneLt:
Are we okay ? Do I merge ?

geofjamg:
> Are we okay ? Do I merge ?

yes

PR #659, opened 2023-10-11
Remove useless logs

geofjamg:

No

Bug fix

When debug log level is activated for powsybl logger, we get lot's of useless logs because:
 - log level is set to root logger instead of com.powsybl logger
 - for the specific case of CGMES import we get triple store logs that are too verbose

When debug log level is activated for powsybl logger,  we get less logs.

**Other information**:


java/src/main/resources/logback.xml
```diff
@@ -11,4 +11,7 @@
     <logger name="org.nocrala.tools.texttablefmt" level="ERROR" additivity="false">
         <appender-ref ref="custom" />
     </logger>
+    <logger name="com.powsybl.triplestore" level="INFO" additivity="false">
```
annetill, line 14: This is my configuration.
````
<!-- To see the query text in the log, set level to debug -->
<logger name="com.powsybl.triplestore.TripleStore" level="error"/>
<!-- To see how much time each query costs, set level to debug -->
<logger name="com.powsybl.cgmes.triplestore.CgmesModelTripleStore" level="error"/>
<!-- To watch the properties of every object converted, level to debug -->
<logger name="com.powsybl.cgmes.conversion.Conversion" level="error"/>
````

geofjamg: This is really too much restrictive, you never get any of the warning with this config. We rather should improve and synthesize CGMES converter logs. The config that I did is already less verbose that current one without taking the risk to hide to the user some very useful informations.

annetill (approved):

PR #662, opened 2023-10-13
Add target v sensitivity on current and reactive power

EtienneLt:

No

Feature

Only a subset of available sensitivity are interfaced in pypowsybl.

We added a generic factor add method that allows to configure all of the available sensitivity types (functions and variables). For instance we can now compute sensitivity of a reactive power flow or a current to a voltage. 

**Other information**:


cpp/src/bindings.cpp
```diff
@@ -683,6 +687,17 @@ PYBIND11_MODULE(_pypowsybl, m) {
             .value("NONE", contingency_context_type::NONE)
             .value("SPECIFIC", contingency_context_type::SPECIFIC);
 
+    py::enum_<sensitivity_function_type>(m, "SensitivityFunctionType")
```
annetill, line 684: I don't understand python very much but you miss BUS_VOLTAGE

geofjamg (approved):

PR #669, opened 2023-10-27
GraalVM JDK 21

geofjamg:

No

JDK version update

In order to use '--pgo-instrument  -H:+ProfilingEnableProfileDumpHooks' to instrument for PGO a shared library we need to upgrade to JDK 21. This is not supported with JDK 17.
But there is an issue in JDK 21 that does not allow to use native image with the class path https://github.com/oracle/graal/issues/7712. So we need to wait for a fix or a workaround.
Another issue that have to be solved for upgrading to JDK21: https://github.com/oracle/graal/issues/7759

24/08/04 Still same error with JDK22
```
Error: Expected Object but got Word for call argument in com.powsybl.python.network.NetworkCFunctions.setWorkingVariant(NetworkCFunctions.java:376). One possible cause for this error is when word values are passed into lambdas as parameters or from variables in an enclosing scope, which is not supported, but can be solved by instead using explicit classes (including anonymous classes).
```
 

**Other information**:


java/pom.xml
```diff
@@ -61,9 +61,9 @@
     <version>1.1.0-SNAPSHOT</version>
 
     <properties>
-        <java.version>17</java.version>
+        <java.version>21</java.version>
```
EtienneLt, line 70: no problem with maven version ?

geofjamg: We will revert this to 17 in this pr to keep java 17 source compatibility, even using jdk 21

java/pypowsybl/src/main/java/com/powsybl/python/commons/CommonCFunctions.java
```diff
@@ -96,20 +128,33 @@ private static void freeSeries(SeriesPointer seriesPointer) {
...
+            }
+        });
     }
 
     @CEntryPoint(name = "getWorkingVariantId")
     public static CCharPointer getWorkingVariantId(IsolateThread thread, ObjectHandle networkHandle, ExceptionHandlerPointer exceptionHandlerPtr) {
-        return doCatch(exceptionHandlerPtr, () -> {
-            Network network = ObjectHandles.getGlobal().get(networkHandle);
-            return CTypeUtil.toCharPtr(network.getVariantManager().getWorkingVariantId());
+        return doCatch(exceptionHandlerPtr, new PointerProvider<>() {
```
obrix, line 141: I think you can simplify the syntax here and everywhere this pattern is used by providing a lambda as second parameter "() -> ..." instead of reimplementing a PointProvider locally.

obrix (changes requested):
A minor syntax comment but I am ok with the rest.

java/pypowsybl/src/main/java/com/powsybl/python/commons/CommonCFunctions.java
HugoKulesza: It is mandatory to now use these classes because lambdas no longer work with GraalVM for java 21 if they capture GraalVM objects (in your example an ObjectHandle object)

obrix: Okay I see.

obrix (approved):

PR #671, opened 2023-10-31
Update voltage initializer to open reac version 0.3

### LINKED ISSUE #665: Integration of new version of OpenReac

For specific voltage `VoltageLimitOverride` (Java), we map currently to python like this:
````
params.add_specific_voltage_limits({"vl_id": (0.5, 1.2)}) // vl id, minVoltage, maxVoltage
(see unit tests)
````
Where `minVoltage` and `maxVoltage` are the relative difference between IIDM voltage limits (low for min and high for max). 
new Limit of voltage level vl_id = vl.getLow(High)VoltageLimit() + min(max)Voltage

Now, the voltage limits override has changed in Java to
````
VoltageLimitOverride(String voltageLevelId, VoltageLimitType type, boolean isRelative, double limit)
````
with voltage limit type low or high. 

I think that in pypowsybl we can imagine a dataframe:
````
params.add_specific_voltage_limits({"vl_id": (0.5, true, 1.2, true)}) // vl id, minVoltage, relative, maxVoltage, relative
````
or maybe:
````
params.add_specific_voltage_limits({"vl_id": (0.5, 1.2, true, true)}) // vl id, minVoltage, maxVoltage, relative for min, relative for max
params.add_specific_voltage_limits({"vl_id": (380, 420, false, false)}) // for an absolute change.
params.add_specific_voltage_limits({"vl_id": (380, 0, false, true)}) // for one side change.
````

### Describe the motivation

_No response_

### Extra Information

_No response_

### PR DESCRIPTION
**Pull request by pjeanmarie**
…ith new java API for addSpecificVoltageLimits

**Other information**:

### CODE REVIEW

**Comment by pjeanmarie** [comment]
Question: I have updated create_voltage_limit_override (using two fonctions as for add_specific_voltage_limits) but is it used? Maybe create_voltage_limit_override need to be removed instead of updated

@geofjamg  We will supress create_voltage_limit_override which was created in https://github.com/powsybl/pypowsybl/pull/627, is it a problem?

**Comment by pjeanmarie** [comment]
Notes about chosen solution:
It allows to:
- give only low or only high limit to a voltage level as it is now possible with the new java api
- use the new "is relative" parameter
- instead of the java API addSpecificVoltageLimits which have enum (LOW_VOLTAGE_LIMIT, HIGH_VOLTAGE_LIMIT) as a parameter, the python API is splitted in two functions add_specific_low_voltage_limits and add_specific_high_voltage_limits

**Comment by pjeanmarie** [comment]
(Integration) testing?

File: tests/test_voltage_initializer.py [hunk 1]
```diff
@@ -19,7 +19,12 @@ def test_parameters():
     params.add_variable_two_windings_transformers(["twt1", "twt2"])
 
     params.add_algorithm_param({"foo": "bar", "bar": "bar2"})
-    params.add_specific_voltage_limits({"vl_id": (0.5, 1.2)})
```
annetill, line 22: Yes I understand your solution but this is a breaking change. The rule is to never introduce breaking changes for python users. @geofjamg what should we do in that case?

EtienneLt: maybe we can keep also the previous python method and add a deprecated tag

annetill: See with @geofjamg, can you also keep the previous method to avoid breaking change?

pjeanmarie:
WIP -> Missing one test which will check that the values send by the new API are taken into account in the java network

tests/test_voltage_initializer.py
pjeanmarie: Yes no problem, and I can add "deprecated" on the previous one. I guess the reason for deprecation being to avoid having multiple similar API with similar names for same kind of behavior.

pjeanmarie:
Do not know how to really test this feature, all VoltageInitializerParameters.run() seems to crash with error "Module ampl is missing in the configuration file" for all given networks. It was unseen because the test VoltageInitializerParameters.run is skipped in the main, so no test in main. However, I can go from "At least one voltage level has an undefined or incorrect voltage limit." error to the error above by adding missing voltage level limits with the new API, so it "works".

obrix:
@pjeanmarie For deleting create_voltage_limit_override I think it is okay, it returns a JavaHandle for a VoltageLimitOverride but there is no use for it because there is not API taking it. This may have been a useless function that we missed in the initial review.
@EtienneLt Are you okay with that ?

EtienneLt:
> @pjeanmarie For deleting create_voltage_limit_override I think it is okay, it returns a JavaHandle for a VoltageLimitOverride but there is no use for it because there is not API taking it. This may have been a useless function that we missed in the initial review. @EtienneLt Are you okay with that ?

ok for me

obrix (approved):
Looks good to me !

Issue #401, opened 2022-05-23
Network loading API: support for binary streams


sylvlecl:
* **Do you want to request a *feature* or report a *bug*?**

Feature.

* **What is the current behavior?**

We have 2 loading methods:
- `load`: takes a file path as argument
- `load_from_string`: takes the content of a file, as a string, as argument. Does not support byte entries.

So, in particular, in-memory "blobs" cannot be provided as arguments.
Streaming content is not supported either.

* **What is the expected behavior?**

As proposed in #144, we should support byte streams (file object) as arguments to `load`.

**Type checking and runtime behaviour**
After some digging, it seems there is not a very standard way of type cheking for file objects, neither at runtime nor at typecheck time.
We have `typing.BinaryIO` and various classes in `io` module. But `io.BinaryIO` exposes much more methods than necessary:
we could go for a lighter protocol.
As an example, pandas lib seems to perform typechecking by using a union of many types (see comment below), and at runtime will only check for the presence of a `read` method.

--> a good, mixed approach could be to define a simple protocol with at least `read` method, and check for their presence at runtime.

**String input handling**

We should deprecate the `load_from_string` method:
users can provide an in memory buffer instead, with `io.BytesIO` for example.

Once question remains: do we allow only to input binary IO (such as `io.BytesIO`), or also text IO (such as `io.StringIO`) ?
The latter could be handy for text formats.
If we want to distinguish between the 2, there is no standard way ... For example, pandas ends up looking for 'b' character in "mode", and also checking the actual class of the object (against a predefined set of class including `io.TextIOWrapper` etc).

* **What is the motivation / use case for changing the behavior?** 

A use case is being able to load a network from in memory zip content.

sylvlecl:
For reference, pandas uses those unions for typing file inputs:
```python
# filenames and file-like-objects
Buffer = Union[IO[AnyStr], RawIOBase, BufferedIOBase, TextIOBase, TextIOWrapper, mmap]
FileOrBuffer = Union[str, Buffer[AnyStr]]
FilePathOrBuffer = Union["PathLike[str]", FileOrBuffer[AnyStr]]
```

Haigutus:
- I would propose to support only io.BytestIO
 - The same should be done for export
 - On the packaging of files I would propose that the function would take in list of binary objects and find all XML-s independent of packaging

Network.load_bytest(list(object1, object2, objectN)

where object could be a XML file, Zip file with single XML inside, Zip file with multiple XML inside

[...]

PR #672, opened 2023-11-12
Save network to a binary buffer

geofjamg:

#401 

Feature

We can write a network to a binary buffer. The binary buffer is always a zip file whatever the number of created files.
Also some refactoring has been done to rename 'dump' to 'save' which is more usual and consistent with 'load'.

**Other information**:


java/src/main/java/com/powsybl/python/network/ZipMemDataSource.java
```diff
@@ -0,0 +1,84 @@
+/**
+ * Copyright (c) 2023, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
EtienneLt, line 5: SPDX-License-Identifier: MPL-2.0 is missing

geofjamg: Done

EtienneLt (approved):

PR #675, opened 2023-11-17
refactor network area diagram parameters

EtienneLt:

**Other information**:


pypowsybl/network/impl/network.py
```diff
@@ -250,32 +251,39 @@ def get_single_line_diagram(self, container_id: str, parameters: SldParameters =
     def write_network_area_diagram_svg(self, svg_file: PathOrStr, voltage_level_ids: Union[str, List[str]] = None,
                                        depth: int = 0, high_nominal_voltage_bound: float = -1,
                                        low_nominal_voltage_bound: float = -1,
-                                       edge_name_displayed: bool = False) -> None:
+                                       nad_parameters: NadParameters = None) -> None:
```
So-Fras, line 275: I think that we should not break compatibility and keep the old signature as well (with a deprecated mention)

tests/test_network.py
```diff
@@ -859,6 +860,7 @@ def test_sld_nad():
                                          depth=10)
         n.write_network_area_diagram_svg(test_svg, low_nominal_voltage_bound=10, depth=10)
         n.write_network_area_diagram_svg(test_svg, high_nominal_voltage_bound=50, depth=10)
+        n.write_network_area_diagram_svg(test_svg, nad_parameters=NadParameters(edge_name_displayed=False))
```
So-Fras: False is the default value for this parameter. Is it possible to test True so that we are sure that the user's choice is taken into account?

So-Fras (changes requested):
Thank you for refactoring the NadParameters. It will make it symmetrical to the SldParameters!

pypowsybl/network/impl/network.py
EtienneLt: I have a problem : the new method can not have the same name if we keep compatibility

pypowsybl/network/impl/network.py
```diff
@@ -252,30 +253,58 @@ def write_network_area_diagram_svg(self, svg_file: PathOrStr, voltage_level_ids:
                                        low_nominal_voltage_bound: float = -1,
                                        edge_name_displayed: bool = False) -> None:
         """
+        .. deprecated:: 1.1.0
+        Use :class:`write_network_area_diagram_svg` with  `NadParameters` instead.
```
So-Fras: It seems to me that it  should be `Use :class:  write_network_area_diagram ...` instead (without the svg)

So-Fras (changes requested):
I think there is a little typo remaining and then it's good for me!

So-Fras (approved):

PR #681, opened 2023-11-21
Add missing dangling line visitor function

So-Fras:

No

Bug fix

Tie lines are not visited and crossed when filtering voltage levels according to their nominal voltages.

Tie lines are properly visited.

**Other information**:
This is a temporary fix. This piece of code should be transfered into powsybl-diagram (a PR has already be opened for that : https://github.com/powsybl/powsybl-diagram/pull/566)

### DIFF
--- a/java/src/main/java/com/powsybl/python/network/NetworkAreaDiagramUtil.java
+++ b/java/src/main/java/com/powsybl/python/network/NetworkAreaDiagramUtil.java
@@ -183,5 +183,11 @@ public final class NetworkAreaDiagramUtil {
             }
 
         }
+
+        public void visitDanglingLine(DanglingLine danglingLine) {
+            if (danglingLine.isPaired()) {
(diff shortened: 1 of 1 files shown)


java/src/main/java/com/powsybl/python/network/NetworkAreaDiagramUtil.java
```diff
@@ -183,5 +183,11 @@ private void visitTerminal(Terminal terminal) {
             }
 
         }
+
+        public void visitDanglingLine(DanglingLine danglingLine) {
```
EtienneLt, line 187: it is not used anywhere ?

EtienneLt (approved):

java/src/main/java/com/powsybl/python/network/NetworkAreaDiagramUtil.java
So-Fras: It works the same way as visitHvdcConverterStation for example.

PR #683, opened 2023-12-06
add regulating element for generator, vsc and svc creation

EtienneLt:

feature

when a generator, svc or vsc is created his regulating point can not set immediately


java/src/main/java/com/powsybl/dataframe/network/adders/Util.java
```diff
@@ -0,0 +1,35 @@
...
+import com.powsybl.iidm.network.Injection;
+import com.powsybl.iidm.network.Network;
+import com.powsybl.iidm.network.Terminal;
+
+import java.util.function.Consumer;
+
+/**
+ * @author Etienne Lesot <etienne.lesot at rte-france.com>
+ */
+public final class Util {
```
geofjamg: There is already another Util class. You should give it a more precise name like AdderUtil

java/src/main/java/com/powsybl/dataframe/network/adders/Util.java
```diff
@@ -0,0 +1,35 @@
...
+
+    private Util() {
+
+    }
+
+    public static void setRegulatingTerminal(Consumer<Terminal> adder, Network network, String elementId) {
```
geofjamg: This is very similar to `NetworkDataframes.setRegulatedElement` which is used for setting regulating terminal to an existing equipment. We should try to share code as much as possible.

geofjamg (changes requested):

geofjamg (approved):

PR #684, opened 2023-12-06
Add tooltip enabled parameter to single line diagram

EtienneLt:

feature

add new sld parameter to display tooltip

### DIFF
--- a/cpp/src/bindings.cpp
+++ b/cpp/src/bindings.cpp
@@ -469,6 +469,7 @@ PYBIND11_MODULE(_pypowsybl, m) {
         .def_readwrite("center_name", &pypowsybl::SldParameters::center_name)
         .def_readwrite("diagonal_label", &pypowsybl::SldParameters::diagonal_label)
         .def_readwrite("nodes_infos", &pypowsybl::SldParameters::nodes_infos)
+        .def_readwrite("tooltip_enabled", &pypowsybl::SldParameters::tooltip_enabled)
         .def_readwrite("topological_coloring", &pypowsybl::SldParameters::topological_coloring)
         .def_readwrite("component_library", &pypowsybl::SldParameters::component_library);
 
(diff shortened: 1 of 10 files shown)


cpp/src/bindings.cpp
```diff
@@ -470,7 +470,8 @@ PYBIND11_MODULE(_pypowsybl, m) {
         .def_readwrite("diagonal_label", &pypowsybl::SldParameters::diagonal_label)
         .def_readwrite("nodes_infos", &pypowsybl::SldParameters::nodes_infos)
         .def_readwrite("topological_coloring", &pypowsybl::SldParameters::topological_coloring)
-        .def_readwrite("component_library", &pypowsybl::SldParameters::component_library);
+        .def_readwrite("component_library", &pypowsybl::SldParameters::component_library)
+        .def_readwrite("tool_tip_enabled", &pypowsybl::SldParameters::tool_tip_enabled);
```
So-Fras: I would put the `tool_tip_enabled` line just below the `nodes_infos` one because tool_tip_enabled is an `SvgParameters` field (as well as use_name, center_name, diagonal_label and nodes_infos).
The same remark applies to the 9 modified files of this PR.

So-Fras (changes requested):

cpp/src/bindings.cpp
EtienneLt: ok I'll change that

So-Fras (approved):

geofjamg (changes requested):
tooltip is a word, it looks bad to have tool_tip everywhere

java/src/main/java/com/powsybl/python/network/NetworkCFunctions.java
```diff
@@ -918,13 +919,15 @@ public static void freeNadParameters(IsolateThread thread, NadParametersPointer
     public static SldParameters convertSldParameters(SldParametersPointer sldParametersPtr) {
         String componentLibraryName = CTypeUtil.toString(sldParametersPtr.getComponentLibrary());
         SldParameters sldParameters = SingleLineDiagramUtil.createSldParameters()
-                .setStyleProviderFactory(sldParametersPtr.isTopologicalColoring() ? new DefaultStyleProviderFactory() : new NominalVoltageStyleProviderFactory())
+                .setStyleProviderFactory(sldParametersPtr.isTopologicalColoring() ? new DefaultStyleProviderFactory() :
```
geofjamg: the : should be in that case just under the ?

EtienneLt: ok

geofjamg (approved):

PR #702, opened 2024-02-08
Replace ucte-x-node-code with pairing_key

colineplqt:

No

Update

Dangling lines have a field ucte-x-node-code

This field is renamed pairing_key and ucte-x-node-code is deprecated

- [x] Yes
- [ ] No

When creating a dangling line, the column or keyword argument `ucte-x-node-code` should be replaced by `pairing_key`.
When accessing the field from the dangling dataframe, the name of the column is also updated to `pairing_key`.


pypowsybl/network/impl/network.py
```diff
@@ -3619,6 +3620,16 @@ def create_dangling_lines(self, df: DataFrame = None, **kwargs: ArrayLike) -> No
                 network.create_dangling_lines(id='BAT-1', voltage_level_id='VL1', bus_id='B1',
                                               p0=10, q0=3, r=0, x=5, g=0, b=1e-6)
         """
+        if df is not None:
```
colineplqt, line 3625: I need a review on that, I don't know if other ways to deprecate should be used.

EtienneLt: maybe you can try with a test

EtienneLt: I'll do it if you can't run test

colineplqt: I tested in a jupyter notebook and it worked but I did not write any tests as I can't manage to make them run

EtienneLt: ok I test

EtienneLt: @geofjamg @annetill  are we doing a breaking change or we maintain compatibility (can be much code for not so much)

EtienneLt (approved):

PR #717, opened 2024-02-23
add optional on getRegulatedSide

EtienneLt:

no

Bug fix

if the regulated side is null when validating a loadflow there is a null pointer exception

### DIFF
--- a/java/src/main/java/com/powsybl/dataframe/loadflow/validation/Validations.java
+++ b/java/src/main/java/com/powsybl/dataframe/loadflow/validation/Validations.java
@@ -162,7 +162,7 @@ public final class Validations {
                 .ints("low_tap_position", TwtValidationData::getLowTapPosition)
                 .ints("high_tap_position", TwtValidationData::getHighTapPosition)
                 .doubles("target_v", TwtValidationData::getTargetV)
-                .strings("regulated_side", d -> d.getRegulatedSide().name())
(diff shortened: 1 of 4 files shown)


java/src/main/java/com/powsybl/dataframe/loadflow/validation/Validations.java
```diff
@@ -162,7 +162,13 @@ public static DataframeMapper<List<TwtValidationData>> twtsValidationMapper() {
                 .ints("low_tap_position", TwtValidationData::getLowTapPosition)
                 .ints("high_tap_position", TwtValidationData::getHighTapPosition)
                 .doubles("target_v", TwtValidationData::getTargetV)
-                .strings("regulated_side", d -> d.getRegulatedSide().name())
+                .strings("regulated_side", d -> {
+                    if (d.getRegulatedSide().isPresent()) {
```
geofjamg: Could by done using .map(...)

geofjamg (changes requested):
A unit test showing the NPE would be welcome.

java/src/main/java/com/powsybl/dataframe/loadflow/validation/Validations.java
EtienneLt: I dont know how to test in python I cant set a null regulated side on a ratio tap changer

EtienneLt: done

geofjamg (approved):

PR #728, opened 2024-03-22
Map phase and ratio tap changer and shunt compensator actions

obrix:
Map phase and ratio tap changer and shunt compensator action. Also shunt compensator still raise an exception because they are only supported from olf 1.8.0.

no

feature

add two new actions on tap changers and refactor side

- [x] Yes
- [ ] No

BranchSide and Side enums are merged become now only Side 
Side.NONE becomes Side.UNDEFINED


pypowsybl/_pypowsybl.pyi
```diff
@@ -678,6 +678,9 @@ def add_load_active_power_action(security_analysis_context: JavaHandle, action_i
 def add_load_reactive_power_action(security_analysis_context: JavaHandle, action_id: str, load_id: str, is_relative: bool, reactive_power: float) -> None: ...
 def add_generator_active_power_action(security_analysis_context: JavaHandle, action_id: str, generator_id: str, is_relative: bool, active_power: float) -> None: ...
 def add_switch_action(security_analysis_context: JavaHandle, action_id: str, switch_id: str, open: bool) -> None: ...
+def add_phase_tap_changer_position_action(security_analysis_context: JavaHandle, action_id: str, transformer_id: str, is_relative: bool, tap_position: int) -> None: ...
+def add_ratio_tap_changer_position_action(security_analysis_context: JavaHandle, action_id: str, transformer_id: str, is_relative: bool, tap_position: int) -> None: ...
```
EtienneLt: I think the side is missing

EtienneLt: maybe there is no need ?

obrix: Side is optional for two winding transformers but needed for three winding transformers. I think it should be supported. I will add it.

EtienneLt: perfect thanks

pypowsybl/_pypowsybl.pyi
```diff
@@ -551,22 +552,6 @@ class FlowDecompositionParameters:
     sensitivity_variable_batch_size: int
     def __init__(self) -> None: ...
 
-
-class BranchSide:
```
obrix, line 555: @EtienneLt  I tried to to rationalize a bit how side enum is handled and use the same enum everywhere.

cpp/src/pypowsybl-api.h
```diff
@@ -364,9 +364,11 @@ typedef enum {
 } DynamicMappingType;
 
 typedef enum {
-    ONE = 0,
+    UNDEFINED = -1,
```
obrix, line 367: @EtienneLt  I had a small issue here because NONE is an enum value already used and graalvm seems to not like that somehow. So NONE in python -> UNDEFINED on cpp/java .

EtienneLt: ok works for me

pypowsybl/_pypowsybl.pyi
EtienneLt: perfect

EtienneLt (approved):

PR #739, opened 2024-04-25
Three-winding transformers creation

colineplqt:

No

Feature

It is now possible to create a three-winding transformer. 
Tap changer can also be created on three-winding transformers.
Operational limits are already available.

- [ ] Yes
- [x] No

### DIFF
--- a/java/src/main/java/com/powsybl/dataframe/network/adders/PhaseTapChangerDataframeAdder.java
+++ b/java/src/main/java/com/powsybl/dataframe/network/adders/PhaseTapChangerDataframeAdder.java
@@ -150,19 +150,17 @@ public class PhaseTapChangerDataframeAdder implements NetworkElementAdder {
         }
 
         private static void setRegulatedSide(ThreeWindingsTransformer transformer, PhaseTapChangerAdder adder, String regulatedSideStr) {
-            if (regulatedSideStr.isEmpty()) {
-                return;
+            if (!regulatedSideStr.isEmpty()) {
+                ThreeSides regulatedSide = ThreeSides.valueOf(regulatedSideStr);
+                adder.setRegulationTerminal(transformer.getTerminal(regulatedSide));
             }
-            ThreeSides regulatedSide = ThreeSides.valueOf(regulatedSideStr);
-            adder.setRegulationTerminal(transformer.getTerminal(regulatedSide));
         }
 
         private static void setRegulatedSide(TwoWindingsTransformer transformer, PhaseTapChangerAdder adder, String regulatedSideStr) {
-            if (regulatedSideStr.isEmpty()) {
-                return;
+            if (!regulatedSideStr.isEmpty()) {
+                TwoSides regulatedSide = TwoSides.valueOf(regulatedSideStr);
+                adder.setRegulationTerminal(transformer.getTerminal(regulatedSide));
             }
-            TwoSides regulatedSide = TwoSides.valueOf(regulatedSideStr);
-            adder.setRegulationTerminal(transformer.getTerminal(regulatedSide));
         }
     }
 
--- a/java/src/main/java/com/powsybl/dataframe/network/adders/PhaseTapChangerDataframeAdder.java
+++ b/java/src/main/java/com/powsybl/dataframe/network/adders/PhaseTapChangerDataframeAdder.java
@@ -32,7 +32,8 @@ public class PhaseTapChangerDataframeAdder implements NetworkElementAdder {
             SeriesMetadata.booleans("regulating"),
             SeriesMetadata.strings("regulated_side"),
             SeriesMetadata.ints("low_tap"),
-            SeriesMetadata.ints("tap")
+            SeriesMetadata.ints("tap"),
+            SeriesMetadata.strings("side")
     );
 
     private static final List<SeriesMetadata> STEPS_METADATA = List.of(
@@ -71,6 +72,7 @@ public class PhaseTapChangerDataframeAdder implements NetworkElementAdder {
         private final DoubleSeries targetValues;
         private final IntSeries regulating;
         private final StringSeries regulatedSide;
+        private final StringSeries sides;
 
         private final Map<String, TIntArrayList> stepsIndexes;
         private final DoubleSeries g;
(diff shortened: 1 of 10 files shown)


pypowsybl/network/impl/network.py
```diff
@@ -3893,6 +3893,82 @@ def create_2_windings_transformers(self, df: DataFrame = None, **kwargs: ArrayLi
...
+
+            Valid attributes are:
+
+            - **id**: the identifier of the new transformer
+            - **rated_u0**: the rated voltage at the star bus
+            - **voltage_level1_id**: the voltage level where the new transformer will be connected on side 1.
+              The voltage level must already exist.
+            - **bus1_id**: the bus where the new transformer will be connected on side 1,
+              if the voltage level has a bus-breaker topology kind.
+            - **connectable_bus1_id**: the bus where the new transformer will be connectable on side 1,
```
colineplqt: Can anybody explain to me what this field is for? I think it is not very clear, maybe we could improve a bit the documentation?

pypowsybl/network/impl/network.py
```diff
@@ -3893,6 +3893,82 @@ def create_2_windings_transformers(self, df: DataFrame = None, **kwargs: ArrayLi
...
+            - **g2**: the shunt conductance of the side 2, in S
+            - **g3**: the shunt conductance of the side 3, in S
+
+
+        Examples:
+            Using keyword arguments:
+
+            .. code-block:: python
+
+                network.create_3_windings_transformers(id='T-1', rated_u0 = , voltage_level1_id='VL1', bus1_id='B1',
```
colineplqt: Here also I would like to put values that are not completely off range. What could be the rated S?

EtienneLt:
Could you add the documentation in docs/ ?

colineplqt:
> Could you add the documentation in docs/ ?

Yes I will, actually I have to work on a few more things too, creation of tap changers, limits.

java/src/main/java/com/powsybl/dataframe/network/adders/PhaseTapChangerDataframeAdder.java
```diff
@@ -132,14 +148,22 @@ void create(Network network, int row) {
...
+        private static void setRegulatedSide(ThreeWindingsTransformer transformer, PhaseTapChangerAdder adder, String regulatedSideStr) {
+            if (regulatedSideStr.isEmpty()) {
+                return;
+            }
+            ThreeSides regulatedSide = ThreeSides.valueOf(regulatedSideStr);
+            adder.setRegulationTerminal(transformer.getTerminal(regulatedSide));
+        }
+
+        private static void setRegulatedSide(TwoWindingsTransformer transformer, PhaseTapChangerAdder adder, String regulatedSideStr) {
+            if (regulatedSideStr.isEmpty()) {
```
EtienneLt: could you inverse the condition to delete the "return" and simplify ?

java/src/main/java/com/powsybl/dataframe/network/adders/RatioTapChangerDataframeAdder.java
```diff
@@ -116,14 +132,22 @@ void create(Network network, int row) {
...
             adder.add();
         }
-    }
 
-    private static void setRegulatedSide(TwoWindingsTransformer transformer, RatioTapChangerAdder adder, String regulatedSideStr) {
-        if (regulatedSideStr.isEmpty()) {
-            return;
+        private static void setRegulatedSide(TwoWindingsTransformer transformer, RatioTapChangerAdder adder, String regulatedSideStr) {
+            if (regulatedSideStr.isEmpty()) {
+                return;
```
EtienneLt: same here

java/src/main/java/com/powsybl/dataframe/network/adders/RatioTapChangerDataframeAdder.java
```diff
@@ -116,14 +132,22 @@ void create(Network network, int row) {
...
+            }
+            TwoSides regulatedSide = TwoSides.valueOf(regulatedSideStr);
+            adder.setRegulationTerminal(transformer.getTerminal(regulatedSide));
+        }
+
+        private static void setRegulatedSide(ThreeWindingsTransformer transformer, RatioTapChangerAdder adder, String regulatedSideStr) {
+            if (regulatedSideStr.isEmpty()) {
+                return;
```
EtienneLt: here too

pypowsybl/network/impl/network.py
HugoKulesza: I think that it is used if we want to create any network element that can be connected to a bus, by specifying which bus it can be connected to but not actually connecting it to this bus 
"bus1_id" defined -> the element is created connected to the bus 1
"connectable_bus1_id" defined -> the element is created not connected, when it is connected its bus is connectable_bus1_id

java/src/main/java/com/powsybl/dataframe/network/adders/ThreeWindingsTransformerDataframeAdder.java
HugoKulesza, line 1: Can you add a test for this adder in NetworkElementAddersTest ?

java/src/main/java/com/powsybl/dataframe/network/adders/ThreeWindingsTransformerSeries.java
```diff
@@ -0,0 +1,140 @@
...
+        String id = ids.get(row);
+        VoltageLevel vl1 = getVoltageLevelOrThrow(network, voltageLevels1.get(row));
+        VoltageLevel vl2 = getVoltageLevelOrThrow(network, voltageLevels2.get(row));
+        VoltageLevel vl3 = getVoltageLevelOrThrow(network, voltageLevels3.get(row));
+
+        Substation s1 = vl1.getSubstation().orElseThrow(() -> new PowsyblException(COULD_NOT_CREATE_TRANSFORMER + id + NO_SUBSTATION));
+        Substation s2 = vl2.getSubstation().orElseThrow(() -> new PowsyblException(COULD_NOT_CREATE_TRANSFORMER + id + NO_SUBSTATION));
+        Substation s3 = vl3.getSubstation().orElseThrow(() -> new PowsyblException(COULD_NOT_CREATE_TRANSFORMER + id + NO_SUBSTATION));
+        if (!(s1 == s2 && s1 == s3)) {
+            throw new PowsyblException(COULD_NOT_CREATE_TRANSFORMER + id + ": both voltage ids must be on the same substation");
```
HugoKulesza: I think "all voltage ids" is better than "both voltage ids" in the error since there are more than two here

HugoKulesza (changes requested):

geofjamg (approved):

pypowsybl/network/impl/network.py
colineplqt: Ok, thank you, I am going to try to explain that a bit better here

HugoKulesza (approved):

PR #748, opened 2024-05-21
Get bus/breaker view buses dataframe

geofjamg:

No

Feature

We cannot easily get a full list of buses from the bus/breaker view. We need to use the bus/breaker topology object voltage level by voltage level to have it. 

We now have a `get_bus_breaker_view_buses()` directly from the network.

- [ ] Yes
- [ ] No

**Other information**:


java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
```diff
@@ -51,7 +51,8 @@ public static NetworkDataframeMapper getDataframeMapper(DataframeElementType typ
     private static Map<DataframeElementType, NetworkDataframeMapper> createMappers() {
         Map<DataframeElementType, NetworkDataframeMapper> mappers = new EnumMap<>(DataframeElementType.class);
         mappers.put(DataframeElementType.SUB_NETWORK, subNetworks());
-        mappers.put(DataframeElementType.BUS, buses());
+        mappers.put(DataframeElementType.BUS, buses(false));
```
EtienneLt, line 54: cant we use the dataframeContext ? or it is to complicated for a simple thing ?

geofjamg: The first version I coded was using the dataframe context, this is why in previous PR I also made it accessible when creating the element list. I was at the end simpler to just create a new dataframe type and more consistent with the way it is exposed on python side with a dedicated method.

EtienneLt (approved):

PR #762, opened 2024-06-14
[WIP] refactor c part

EtienneLt:

- [ ] Yes
- [ ] No

**Other information**:


cpp/powsybl-cpp/powsybl-api.h
```diff
@@ -314,6 +314,7 @@ typedef struct {
 typedef struct zone_struct {
     char* id;
     char** injections_ids;
+    int injections_ids_count;
```
HugoKulesza, line 317: what is this field used for ?
I guess it is necessary for a size somewhere but it is not used in that way

HugoKulesza: Ok I found it in the java method to create and delete the object

PR #766, opened 2024-06-17
Fix generator creation name

geofjamg:

No

Bug fix

We cannot pass a name to a generator creation like for the other elements creation

We can pass a name

- [ ] Yes
- [ ] No

**Other information**:


java/src/main/java/com/powsybl/dataframe/network/adders/GeneratorDataframeAdder.java
```diff
@@ -35,6 +35,7 @@ public class GeneratorDataframeAdder extends AbstractSimpleAdder {
 
     private static final List<SeriesMetadata> METADATA = List.of(
             SeriesMetadata.stringIndex("id"),
+            SeriesMetadata.strings("name"),
```
EtienneLt, line 39: It needs to be added in the adder

geofjamg: ?

EtienneLt: in createAdder I think a  applyIfPresent(name, row, adder::setName); is missing, or maybe I am wrong

geofjamg: it is already in the base class

EtienneLt (approved):

PR #791, opened 2024-06-28
adding documentation for reduce function

EtienneLt:

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/docs/user_guide/network.rst
+++ b/docs/user_guide/network.rst
@@ -877,3 +877,58 @@ parent network and become again a standalone network.
     :options: +NORMALIZE_WHITESPACE
 
     >>> nl_sub.detach()
+
+Reducing a network
+------------------
+
+Pypowsybl provides methods to reduce a network to a smaller one. It can be done with different parameters.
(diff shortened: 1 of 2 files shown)


geofjamg (approved):
We should also write something in the user guide

docs/user_guide/network.rst
```diff
@@ -877,3 +877,58 @@ parent network and become again a standalone network.
...
+    S3VL1                 S3      400.0               440.0              390.0
+    S4VL1                 S4      400.0               440.0              390.0
+
+    >>> net.reduce(vl_depths=[['S1VL1', 1]])
+    >>> net.get_voltage_levels()
+          name substation_id  nominal_v  high_voltage_limit  low_voltage_limit
+    id
+    S1VL1                 S1      225.0               240.0              220.0
+    S1VL2                 S1      400.0               440.0              390.0
+
```
colineplqt, line 932: I think one sentence about the concept of depth would be useful here. I don't have a good idea but something like that:
"S1VL1 is connected to S1VL2 by the transformer TWT, so it is kept after the network reduction. It is the only voltage level connected to S1VL1 by one branch."

PR #794, opened 2024-07-01
Fix CI manylinux

jeandemanged:

no

bugfix / workaround

we yum install `wget` and then use `wget` to install maven

CI fails if `mirrorlist.centos.org` unreacheable

We just use `curl` to install maven

- [x] No

**Other information**:

@EtienneLt @geofjamg not sure his PR is relevant, feel free to close it if not....

maybe `mirrorlist.centos.org` is definitively closed due to CentOS 7 reaching end-of-life on June 30th, 2024...

### DIFF
--- a/.github/workflows/dev-ci.yml
+++ b/.github/workflows/dev-ci.yml
@@ -22,12 +22,9 @@ jobs:
           }
 
     steps:
-      - name: Install Linux packages
-        run: yum install -y wget
-
       - name: Install Maven
         run: |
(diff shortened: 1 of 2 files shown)


geofjamg:
Thanks, indeed because of centos end of life, they shutdowned to centos update server.
What you propose is ok for me.
Another way to solve it would have been to switch the package update server to https://vault.centos.org/ which is kind of a read only archive

geofjamg (approved):

PR #797, opened 2024-07-03
Fix multi substation sld svg

CBiasuzzi:

- [ ] Docs have been added / updated (for bug fixes / features) --> no need.

Bug fix

` write_matrix_multi_substation_single_line_diagram_svg `and `get_matrix_multi_substation_single_line_diagram`:  errors are raised depending on the matrix_ids parameter (the 2D list of substation ids);  
Tests were added to demonstrate the issue in a couple of cases. 
On the java side the matrix's size, decoded from the python parameter, does not seem to be  computed correctly.  

The matrix's size on the Java side is computed correctly.

- [ ] Yes
- [x] No


So-Fras (approved):
Thanks for the fix!

PR #825, opened 2024-09-06
Improve network get_branches and get_injections

geofjamg:

No

Feature

We can get branches and injections by id and also get flow attributes

- [ ] Yes
- [ ] No

**Other information**:


java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
```diff
@@ -87,55 +87,55 @@ private static Map<DataframeElementType, NetworkDataframeMapper> createMappers()
         return Collections.unmodifiableMap(mappers);
     }
 
-    static <U extends Injection<U>> ToDoubleBiFunction<U, NetworkDataframeContext> getPerUnitP() {
+    static <U extends Injection<?>> ToDoubleBiFunction<U, NetworkDataframeContext> getPerUnitP() {
```
HugoKulesza, line 90: Why is it necessary to change U to ? everywhere in this class ?

geofjamg: Because otherwise signature is not compatibles with what is expected in doubles() function when building the dataframe

EtienneLt (approved):

PR #840, opened 2024-09-22
Add Network area model support

jeandemanged:

No

Feature

No area support

Add area support (starting from PowSyBl 2024.2.0 / powsybl-core 6.4.x), see:
- https://github.com/powsybl/powsybl-core/pull/2955

Implementation followup:
- [x] areas
  - [x] get -> network.get_areas()
  - [x] update -> network.update_areas(...)
  - [x] create -> network.create_areas(...)
  - [x] remove -> network.remove_elements(...)
- [x] areas voltage levels
  - [x] network.get_areas_voltage_levels(...)
  - [x] network.create_areas_voltage_levels(...)
- [x] areas boundaries
  - [x] network.get_areas_boundaries(...)
  - [x] network.create_areas_boundaries(...)

- [x] No


jeandemanged:
works but not so happy with having to always redefine all voltagelevels or boundaries on creation, will try to think about something better

pypowsybl/network/impl/network.py
```diff
@@ -2397,6 +2397,148 @@ def get_terminals(self, all_attributes: bool = False, attributes: List[str] = No
...
+
+            .. code-block:: python
+
+                net = pp.network.create_eurostag_tutorial_example1_with_tie_lines_and_areas()
+                net.get_areas_boundaries()
+
+            will output something like:
+
+            ============= ================ ===== =========== ===========
+            \             element    ac           p           q
```
jeandemanged: misaligned

java/src/main/java/com/powsybl/python/commons/PyPowsyblApiHeader.java
```diff
@@ -623,7 +623,14 @@ public enum ElementType {
...
         TERMINAL,
-        SUB_NETWORK;
+        SUB_NETWORK,
+        AREA,
+        AREA_VOLTAGE_LEVELS,
+        AREA_BOUNDARIES,
+        AREA_ADD_VOLTAGE_LEVELS,
+        AREA_REMOVE_VOLTAGE_LEVELS,
+        AREA_ADD_BOUNDARIES,
+        AREA_REMOVE_BOUNDARIES;
```
jeandemanged: this is quite tweaking the `ElementType` concept ... please let me know if this is OK of if must be done differently...

geofjamg: Nok for me, this is really not the purpose of the dataframes. Dataframes contains some data, not operations, and this is the way we use data (though a creation, modif or read method) that we express the operation.

jeandemanged: yes it was a failed attempt at solving https://github.com/powsybl/pypowsybl/pull/840#issuecomment-2373259730, but thinking about it, we don't redefine areas that often, so I reverted to something simpler

pypowsybl/network/impl/network.py
```diff
@@ -2403,6 +2403,147 @@ def get_terminals(self, all_attributes: bool = False, attributes: List[str] = No
         """
         return self.get_elements(ElementType.TERMINAL, all_attributes, attributes)
 
+    def get_areas(self, all_attributes: bool = False, attributes: List[str] = None,
+                  **kwargs: ArrayLike) -> DataFrame:
+        r"""
+        Get a dataframe of areas.
```
HugoKulesza, line 2409: I think it would be a good idea to add something like : "To retrieve the voltage levels of the areas or their boundaries see get_areas_voltage_levels and get_areas_boundaries" to have an easier full picture when browsing the documentation

HugoKulesza (changes requested):

pypowsybl/network/impl/network.py
jeandemanged: good point, done

HugoKulesza (approved):

PR #859, opened 2024-10-10
Add mapping to all operational limits, allow to modify selected set of limits for each element

### LINKED ISSUE #857: Allow to retrieve all operational limits groups and not only the active one

In powsybl-core, the management of operational limits has recently changed.
Now, network elements may have multiple operational limit groups identified by an id, with one being selected as the active one.

Currently, the `get_operational_limits` method only retrieve the active limit set for each element. 

We should be able to retrieve the limits from the sets that are not active, and probably also to change the active set in the python interface.
Some thoughts on the solution :
- either modify `get_operational_limits` to add inactive sets sets, and two columns precising for each limit the set id and if it is active.
- or add another method to retrieve all limits
- either way, add an `update_active_operational_limits` to change the active sets of operational limits for network elements

### Describe the motivation

_No response_

### Extra Information

_No response_

HugoKulesza:

Fixes #857 

- [ ] Yes
- [x] No


java/src/main/java/com/powsybl/dataframe/network/NetworkDataframes.java
```diff
@@ -519,6 +524,10 @@ static NetworkDataframeMapper twoWindingTransformers() {
                 .booleans("connected1", twt -> twt.getTerminal1().isConnected(), connectBranchSide1())
                 .booleans("connected2", twt -> twt.getTerminal2().isConnected(), connectBranchSide2())
                 .booleans("fictitious", Identifiable::isFictitious, Identifiable::setFictitious, false)
+                .strings("selected_limits_group_1", twt -> twt.getSelectedOperationalLimitsGroupId1().orElse("DEFAULT"),
```
geofjamg: To do for next release: in core expose this string in the API and reuse it here.
```java
class OperationalLimitsGroupsImpl implements FlowsLimitsHolder {

    private static final String DEFAULT_SELECTED_OPERATIONAL_LIMITS_GROUP_ID = "DEFAULT";
```

geofjamg: In the meantime you should create a constant in that class.

java/src/main/java/com/powsybl/dataframe/network/adders/SelectedOperationalLimitsDataframeAdder.java
```diff
@@ -0,0 +1,19 @@
+package com.powsybl.dataframe.network.adders;
+
+import com.powsybl.dataframe.SeriesMetadata;
+import com.powsybl.dataframe.update.UpdatingDataframe;
+import com.powsybl.iidm.network.Network;
+
+import java.util.List;
+
+public class SelectedOperationalLimitsDataframeAdder implements NetworkElementAdder {
```
geofjamg: Is this class used ?

java/src/main/java/com/powsybl/python/network/TemporaryLimitData.java
```diff
@@ -25,6 +25,8 @@ public class TemporaryLimitData {
     private final double value;
     private final int acceptableDuration;
     private final boolean isFictitious;
+    private final String groupId;
+    private final boolean isSelected;
```
geofjamg: You should call it "selected", isSelected is for the getter

pypowsybl/network/impl/network.py
```diff
@@ -3438,17 +3452,23 @@ def get_operational_limits(self, all_attributes: bool = False, attributes: List[
           - **value**:      The value of the limit
           - **acceptable_duration**: The duration, in seconds, for which the element can securely be
             operated under the limit value. By convention, the value -1 represents an infinite duration.
-          - **is_fictitious**: true if this limit is fictitious
+          - **is_fictitious** (optional): `True` if this limit is fictitious
```
geofjamg: -> fictitious

geofjamg (approved):

PR #873, opened 2024-10-21
Support creation and removal of internal connections

### LINKED ISSUE #849: Node/Breaker topology: Add python bindings for creation and removal of internal connections

No Python bindings are available for:
- adding internal connections
- removing internal connections

Python bindings should be available for:
- Internal Connection adder: https://javadoc.io/static/com.powsybl/powsybl-core/6.4.1/com/powsybl/iidm/network/VoltageLevel.NodeBreakerView.InternalConnectionAdder.html
- Internal Connection removal: [com.powsybl.iidm.network.VoltageLevel.NodeBreakerView#removeInternalConnections](https://javadoc.io/static/com.powsybl/powsybl-core/6.4.1/com/powsybl/iidm/network/VoltageLevel.NodeBreakerView.html#removeInternalConnections(int,int))

### Describe the motivation

Python bindings completeness

### Extra Information

_No response_

jeandemanged:

yes, fixes #849 

- [x] No


pypowsybl/network/impl/network.py
HugoKulesza, line 1: Have you considered adding these methods in the NodeBreakerTopology class ? 
I'm not sure if this is a good idea, but since it is already used to get all the info regarding these internal connections ?

jeandemanged: considered yes but not adopted because:
- all creation methods are today on network, this would make an exception
- to be able to create or delete user would be forced first to get a NodeBreakerTopology object so it is not so convenient - but maybe you are expecting a static method ?
- after creating or deleting the NodeBreakerTopology object would not reflect the creation or deletion - but again, maybe you are expecting a static method ?

I think it is simpler / easier / more consistent to have these methods at Network level

HugoKulesza: I think the last point is indeed the worst one to deal with (even with static method I guess)

geofjamg: >> I think it is simpler / easier / more consistent to have these methods at Network level

Yes, what you have done is the right way to do it.

geofjamg (approved):

PR #879, opened 2024-10-29
Fix optional import of pandapower

HugoKulesza:

No

Bug fix

If the user does not import pandapower, "import pypowsybl" result in an error

An import of pypowsybl without pandapower installed will notify the user that the convert_from_pandapower method will not be available and then completes correctly

- [ ] Yes
- [X] No

### DIFF
--- a/pypowsybl/network/impl/pandapower_converter.py
+++ b/pypowsybl/network/impl/pandapower_converter.py
@@ -18,8 +18,6 @@ from pandas import DataFrame
 try:
     from pandapower import pandapowerNet
 except ImportError:
-    print("No installation of pandapower found, network.convert_from_pandapower method will not be usable")
-    print("Use 'pip install pypowsybl[pandapower]' to install the optional dependency if needed.")
     pandapowerNet = any
 
 from .network import Network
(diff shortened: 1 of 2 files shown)


pypowsybl/network/impl/pandapower_converter.py
```diff
@@ -13,9 +13,14 @@
 import numpy as np
 import pandas as pd
 import pypowsybl._pypowsybl as _pp
-from pandapower import pandapowerNet
 from pandas import Series
 from pandas import DataFrame
+try:
+    from pandapower import pandapowerNet
```
geofjamg, line 19: This is not useful to warm the user until he wants to use it. There is already such a warning in the convert function

setup.cfg
```diff
@@ -35,7 +35,7 @@ install_requires =
     networkx
 
 [options.extras_require]
-extra =
+pandapower =
```
geofjamg, line 38: Are you sure of that?

HugoKulesza: We can name the groups of extra dependencies the way we want, so I think it can be more explicit to call pypowsybl[pandapower] instead of pypowsybl[extra] when trying to install with pandapower

geofjamg: ok

geofjamg (approved):

PR #881, opened 2024-11-01
Fix bus/breaker topology in case of no connected terminal present

jeandemanged:

No

Bug fix

As reported on Slack by a PyPowSyBl user (many thanks):

> Hello everyone, I might have found a bug when using network.get_bus_breaker_topology.
> In my network there are 3 buses in one voltage level. The third is only connected via breakers to the other two. My expectation would be, that all 3 buses in network.get_bus_breaker_topology.buses have the same bus_id. But the ones only connected by breakers without additional elements are not linked to any bus whatsoever. Is that intended behaviour or a bug?
> Here is a minimal example:

```python
import pypowsybl as pp
import pandas as pd
network = pp.network.create_empty()
network.create_substations(id=['S1', 'S2'])
voltage_levels = pd.DataFrame.from_records(index='id', data=[
    {'substation_id': 'S1', 'id': 'VL1', 'topology_kind': 'BUS_BREAKER', 'nominal_v': 400},
    {'substation_id': 'S2', 'id': 'VL2', 'topology_kind': 'BUS_BREAKER', 'nominal_v': 400},
])
network.create_voltage_levels(voltage_levels)


network.create_buses(id=['B1_1', 'B1_2', 'B1_3', 'B2_1', 'B2_2'], voltage_level_id=['VL1', 'VL1', 'VL1', 'VL2', 'VL2'])
network.create_lines(id='LINE 1', voltage_level1_id='VL1', bus1_id='B1_2',
                     voltage_level2_id='VL2', bus2_id='B2_1',
                     b1=0, b2=0, g1=0, g2=0, r=0.5, x=10)
network.create_switches(pd.DataFrame.from_records(index='id', data=[
    {'voltage_level_id': 'VL1', 'id': 'DISC-BBS1-LOAD', 'kind': 'BREAKER', 'bus1_id': 'B1_1', 'bus2_id': 'B1_2', 'open':False},
    {'voltage_level_id': 'VL1', 'id': 'DISC-BBS2-LOAD', 'kind': 'BREAKER', 'bus1_id': 'B1_1', 'bus2_id': 'B1_3', 'open':False}
]))
network.get_bus_breaker_topology('VL1').buses

# Result
#  	name	bus_id
# id		
# B1_1		
# B1_2		VL1_0
# B1_3
```

Root cause of the issue is that the existing code assumes that every bus in bus/breaker view has at least one connected non-switch equipment.

We don't assume anymore that every bus in bus/breaker view has at least one connected non-switch equipment.

```python
# Result
#  	name	bus_id
# id		
# B1_1		VL1_0
# B1_2		VL1_0
# B1_3		VL1_0
```

- [x] No


java/src/main/java/com/powsybl/python/network/NetworkUtil.java
```diff
@@ -246,4 +247,37 @@ public static String getRegulatedElementId(Supplier<Terminal> regulatingTerminal
...
+    /**
+     * @param b bus in Bus/Breaker view
+     * @return bus in bus view containing b if there is one, or null if none.
+     */
+    public static Bus getBusViewBus(Bus b) {
+        // First we try the fast and easy way using connected terminals. Works for the vast majority of buses.
+        Bus busInBusView = b.getConnectedTerminalStream().map(t -> t.getBusView().getBus())
+                .filter(Objects::nonNull)
+                .findFirst()
+                .orElse(null);
```
flo-dup: As the link is clean and direct in bus breaker, could you call this code only if the voltage level is node breaker?

java/src/main/java/com/powsybl/python/network/NetworkUtil.java
```diff
@@ -246,4 +247,37 @@ public static String getRegulatedElementId(Supplier<Terminal> regulatingTerminal
...
+            return busInBusView;
+        }
+        // Didn't find using connected terminals. There is the possibility that the bus has zero connected terminal
+        // on its own but is still part of a Merged Bus via a closed retained switch. We examine this case below.
+        VoltageLevel voltageLevel = b.getVoltageLevel();
+        if (voltageLevel.getTopologyKind() == TopologyKind.BUS_BREAKER) {
+            // Bus/Breaker. There is an easy method directly available.
+            return voltageLevel.getBusView().getMergedBus(b.getId());
+        } else {
+            // Node/Breaker. We should probably build something more efficient on powsybl-core side to avoid having
```
flo-dup: Indeed! even though this is a corner case which should rarely occur.
Besides each call to `getBusStreamFromBusViewBusId` loops over all the nodes of the voltage level graph.

java/src/main/java/com/powsybl/python/network/NetworkUtil.java
```diff
@@ -246,4 +247,37 @@ public static String getRegulatedElementId(Supplier<Terminal> regulatingTerminal
...
+            // to loop over all buses in the voltage level.
+            for (Bus bus : voltageLevel.getBusView().getBuses()) {
+                boolean found = voltageLevel.getBusBreakerView().getBusStreamFromBusViewBusId(bus.getId())
+                        .anyMatch(b2 -> b.getId().equals(b2.getId()));
+                if (found) {
+                    return bus;
+                }
+            }
+            return null;
```
flo-dup: maybe with a stream instead?
```java
return voltageLevel.getBusView().getBusStream()
        .filter(bus -> ...)
        .findFirst()
        .orElse(null);
```

flo-dup (commented):
Thanks for having sorted this out!

flo-dup (approved):

java/src/main/java/com/powsybl/python/network/NetworkUtil.java
```diff
@@ -246,4 +247,35 @@ public static String getRegulatedElementId(Supplier<Terminal> regulatingTerminal
...
+            Bus busInBusView = b.getConnectedTerminalStream().map(t -> t.getBusView().getBus())
+                    .filter(Objects::nonNull)
+                    .findFirst()
+                    .orElse(null);
+            if (busInBusView != null) {
+                return busInBusView;
+            }
+            // Didn't find using connected terminals. There is the possibility that the bus has zero connected terminal
+            // on its own but is still part of a Merged Bus via a closed retained switch. We examine this case below.
+            // We should probably build something more efficient on powsybl-core side to avoid having
```
jonenst, line 272: A good use case for
https://github.com/powsybl/powsybl-core/issues/3107
>  you can't get from busbreakerbuses to their busviewbuses

PR #883, opened 2024-11-06
Voltage level topology kind update

geofjamg:

No

Feature

We can convert a node/breaker topology model to a bus/breaker one.

- [ ] Yes
- [x] No

**Other information**:


java/src/main/java/com/powsybl/python/commons/Util.java
```diff
@@ -68,6 +68,7 @@ public static void setException(PyPowsyblApiHeader.ExceptionHandlerPointer excep
         // we need to create a non null message as on C++ side a null message is considered as non exception to rethrow
         // typically a NullPointerException has a null message and an empty string message need to be set in order to
         // correctly handle the exception on C++ side
+        t.printStackTrace();
```
HugoKulesza: This causes a security issue in sonar, is it meant to be used only for debug or to keep it in the PR ?

geofjamg: Now ready for review!

HugoKulesza (approved):

PR #887, opened 2024-11-07
Tap controllers from pandapower

### LINKED ISSUE #513: Misleading tap_pos (ratio_tap_position1,2,3 and phase_tap_position1,2,3) values for trafo without tap controller
BUG

Currently in the trafos dataframe one can see for the trafos with no tap controller defined, the tap_pos , referring to the columns ratio_tap_position1,2 and phase_tap_position1,2 as default values (-99999), this leads to inconsistency in the interpretation of the dataframe.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Simply check the df generated for trafos without any Ratio or phase tap controllers for any trafo ends

Instead of the default value one should post process the dataframe and fiölter out these default values, further replacing them with Nones.
Something like this 
```

df_traf3w=powsybl_net.get_3_windings_transformers()

df_traf3w.query("ratio_tap_position1 == -99999").replace(-99999,None)
```
Obviously then we still need an update statement to update the dataframe I am not aware of a faster more performant alternative.

* **What is the motivation / use case for changing the behavior?**

Data consistency

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

AnkurArohi:
Signed-off-by: Ankur Arohi <ankurarohi1@gmail.com>

- [ ] Yes
- [x] No

**Other information**:


AnkurArohi:
Tests are already provided in the previous PR from @geofjamg, this PR aims to change the value of tap position based on special attributes set in pandapower network.

pypowsybl/network/impl/pandapower_converter.py
```diff
@@ -204,6 +206,51 @@ def create_shunts(n: Network, n_pdp: pandapowerNet) -> None:
...
 
+def extract_tap_info(trafo_and_bus):
+    """
+    This function ensures that the pandapower model was correct wrt phase_shifter boolean
+    Furthermore, the tap_step_percent is calculated from tap_step_degree which is
+    then augmented to the tap_step_percent already present, this means powsybl can just use
+    tap_step_percent for the calculations and then calculates the n_tap value
+    """
+    # Setup logging configuration
+    logging.basicConfig(level=logging.ERROR, format='%(levelname)s: %(message)s')
```
geofjamg: To remove, this is not the right place for configuring logs

pypowsybl/network/impl/pandapower_converter.py
```diff
@@ -204,6 +206,51 @@ def create_shunts(n: Network, n_pdp: pandapowerNet) -> None:
...
+    # Identify phase shift transformers (tap_phase_shifter == True)
+    phase_shift_mask = trafo_and_bus['tap_phase_shifter']
+
+    # Log error if both tap_step_degree and tap_step_percent are parameterized for phase shift transformers
+    both_parameterized_mask = phase_shift_mask & degree_mask & ~np.isnan(trafo_and_bus['tap_step_percent'])
+
+    if np.any(both_parameterized_mask):
+        for idx in trafo_and_bus.index[both_parameterized_mask]:
+            logging.error(
+                f"Transformer at index {idx} has both 'tap_step_degree' and 'tap_step_percent' parameterized.")
```
geofjamg: IIDM model allows creation of a ratio tap changer AND a phase tap changer, we don't have to choose between the 2.

geofjamg (changes requested):

pypowsybl/network/impl/pandapower_converter.py
AnkurArohi: yes, this log only suggests the user that in pandapower it is not allowed to set tap_step_degree as tap_step_percent. The motivation here is to help the user to debug the cases ogf incorrect network modelling.

AnkurArohi:
@geofjamg I have worked on your comments

geofjamg (commented):
I'm not sure to understand whar you are trying to solve in this PR:
 - Are you trying to take into account the effect of the phase shifter to the voltage ratio ? in that case could you give me a link to the pandapower doc describing this.
 - Why are you not also creating a phase tap changer with the given phase shift. It seems to me the most important part of a ĥase shifter to take into account.
 - Could you add a test case with a phase shifter to how the issue before and solved one after ?

PR #909, opened 2024-12-04
Get network elements properties as data frame

### LINKED ISSUE #894: Handling large number of properties

With the introduction of https://github.com/powsybl/powsybl-core/pull/3139 in CGMES importer, getting e.g. line data frame on large CGMES based networks using `all_attributes=True` is really slow (tens of seconds).
Each line now has properties `CGMES.OperationalLimitSet_<mRID>` that is unique to the line, hence the data frame has many many columns since a column is added for each property. Same for 2W and 3W transformers.

Probably https://github.com/powsybl/powsybl-core/pull/3139 made an abusive use of the iIDM Identifiable property feature, and CGMES importer/exporter should be improved to make use of an extension instead of properties in this case. Or alternatively, continue using property, but with stable keys / not including equipment specific keys/mRID.

That being said, there are things which could be improved on PyPowSyBl side:
- First, improve to add properties columns only on request in the `Network.get_<equipmentType>` methods, e.g. via an additional argument `property_attributes=True/False` to be used in conjunction with `all_attributes=True/False`, i.e. `all_attributes` would only care about "optional" columns, and the new argument would decide on the addition of the "property" columns.
- Second, add a method `Network.get_elements_properties` allowing to get properties as rows instead of columns, much like the existing [`Network.get_aliases`](https://powsybl.readthedocs.io/projects/pypowsybl/en/latest/reference/api/pypowsybl.network.Network.get_aliases.html#pypowsybl.network.Network.get_aliases)

### Describe the motivation

Performance, eventually usability

### Extra Information

Since PyPowSyBl 1.8.x

jeandemanged:

Related to #894, in part only, please do not close the issue when merging this PR.

Feature

- We can get network elements properties only if:
  - we know exactly the properties we are looking for (we know their keys), 
  - or, we systematically use `all_attributes=True` in the `get_<element_type>` methods, in which case properties are added as columns.
- Also, we cannot get properties of Network objects. (root network and subnetworks)

- new method `Network.get_elements_properties`, returning a data frame of properties "row-oriented" - to be compared with the `get_<element_type>` methods returning properties in columns.
We can also get properties of the network objects with this method

``` python
>>> import pypowsybl as pp
>>> network = pp.network.create_eurostag_tutorial_example1_network()
>>> network.get_elements_properties()
Empty DataFrame
Columns: [key, value]
Index: []
>>> network.add_elements_properties(id='sim1', key1='value1', key2='value2')
>>> network.add_elements_properties(id='GEN', key3='value3')
>>> network.get_elements_properties()
           type   key   value
id
sim1    NETWORK  key1  value1
sim1    NETWORK  key2  value2
GEN   GENERATOR  key3  value3
```

- [x] No

**Other information**:


geofjamg (commented):
@jeandemanged maybe we should add an "element_type" column to the datadrame to make it easy to use/associate with other dataframes ?

jeandemanged:
> @jeandemanged maybe we should add an "element_type" column to the datadrame to make it easy to use/associate with other dataframes ?

Hi @geofjamg,
Indeed
I named the column `type` for consistency with e.g. `get_injections()`, `get_identificables()`, `get_branches()` - although I am not really comfortable having a python reserved keyword as a column name...
Also added a `type` column to `get_aliases()`.

geofjamg (approved):

PR #914, opened 2024-12-12
Dynamic simulation output variables supplier

Lisrte:

No

Feature

Dynamic simulation `CurveMapping` handle curve for dynamic models.

`CurveMapping` is replaced with `OutputVariableMapping` handling curves and final state values for equipment with or without dynamic models.
Final state values results are provided by `SimulationResult.final_state_values`.

- [x] Yes
- [ ] No

Replace `CurveMapping::add_curve` and `CurveMapping::add_curve` with `OutputVariableMapping::add_dynamic_model_curves`.

**Other information**:

Linked to the powsybl-dynawo PR [413](https://github.com/powsybl/powsybl-dynawo/pull/413)
Without the PR, mapping with more than two curves will end with `pandas.errors.InvalidIndexError`.


pypowsybl/dynamic/impl/simulation_result.py
```diff
@@ -37,4 +38,8 @@ def _get_all_curves(self) -> pd.DataFrame:
         curve_name_lst = _pp.get_all_dynamic_curves_ids(self._handle)
         df_curves = [self._get_curve(curve_name)
                      for curve_name in curve_name_lst]
-        return pd.concat(df_curves, axis=1) if df_curves else pd.DataFrame()
+        return pd.concat(df_curves, axis=1).ffill() if df_curves else pd.DataFrame()
```
HugoKulesza, line 41: By curiosity, why is .ffill needed here ?

pypowsybl/dynamic/impl/output_variable_mapping.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) 2023, RTE (http://www.rte-france.com)
```
HugoKulesza: Just a wrong date here

java/src/main/java/com/powsybl/dataframe/dynamic/OutputVariablesSeries.java
```diff
@@ -0,0 +1,41 @@
+/**
+ * Copyright (c) 2020-2022, RTE (http://www.rte-france.com)
```
HugoKulesza: Same here

integration_tests/test_dynawo.py
```diff
@@ -9,13 +9,16 @@
...
+import logging
 
 def test_simulation():
     """
     Running that test requires to have installed dynawo,
     and configured its path in your config.yml.
     """
+
+    logging.basicConfig()
+    logging.getLogger('powsybl').setLevel(1)
```
HugoKulesza: This is probably not useful to keep for future run of integration tests

pypowsybl/dynamic/impl/simulation_result.py
Lisrte: Curves are represented as a map of `DoubleTimeSeries` in Powsybl. When creating a `DoubleTimeSeries` for an individual curve, entries with the same value as the previous one are skipped. Thus when we concat  curves series we have to fill forward in order to get these values back.

HugoKulesza (approved):

PR #928, opened 2025-01-08
adds a fixed_positions param to  get_network_area_diagram

CBiasuzzi:

feature

  Adds an optional parameter fixed_positions to the get_network_area_diagram, to be able to generate a NAD with elements set to fixed positions. 
  Fixed_positions is a JSON encoded metadata string ( e.g., result of a previously computed NAD).
   Note that positions for diagram elements that are not specified in the JSON will be computed using the current layout algorithm.

- [ ] Yes
- [ ] No

**Other information**:


pypowsybl/network/impl/network.py
```diff
@@ -392,7 +392,7 @@ def write_network_area_diagram(self, svg_file: PathOrStr, voltage_level_ids: Uni
 
     def get_network_area_diagram(self, voltage_level_ids: Union[str, List[str]] = None, depth: int = 0,
                                  high_nominal_voltage_bound: float = -1, low_nominal_voltage_bound: float = -1,
-                                 nad_parameters: NadParameters = None) -> Svg:
+                                 nad_parameters: NadParameters = None, fixed_positions: str = None) -> Svg:
```
flo-dup: It would be a lot more handy to have a dataframe of positions here if we want to process the positions in python. And then a utility method to get such a dataframe from the metadata json

CBiasuzzi: we must not forget that, currently, in the json metadata are  encoded the text nodes' positions , other than the nodes' positions (and could it be that, in the future, the json would contain some more graph elements?),  so 
a single dataframe parameter would not be enough in the get_network_area_diagram python api.  
Besides, the get_network_area_diagram alredy returns a couple: the svg and the json metadata.
I wonder if it could not be better to keep it simpler, leave here the json str metadata parameter and (in the future when it is needed) write a python utility/adapter method that creates a suitable json metadata from one or more positions dataframes (and vice versa a json -> multiple dataframes converter). What do you think?

flo-dup: It's clearly simpler what you suggest, but I think that having a json as one parameter among several parameters of a method is strange. Even though it's very handy in current development state.

Besides, currently the json metadata is the only thing that holds the information. So it's the easiest way to give it. But we've planned to have a specific json format which would contain all the positions (all nodes including text nodes), and only this information. If we do it now with the metadata, there will soon be a breaking change.

About the multiple dataframes, I thought that the needed information could be in a single dataframe like this (or with less columns with `(x,y)` pairs) :
screenshot

pypowsybl/network/impl/network.py
```diff
@@ -390,9 +390,19 @@ def write_network_area_diagram(self, svg_file: PathOrStr, voltage_level_ids: Uni
...
                                            voltage_level_ids, depth, high_nominal_voltage_bound, low_nominal_voltage_bound, nad_p)
 
+    def _create_nad_positions_c_dataframe(self, df: DataFrame) -> _pp.Dataframe:
+        nad_positions_metadata=[_pp.SeriesMetadata('id',0,True,False,False),
+                  _pp.SeriesMetadata('x',1,False,False,False),
+                  _pp.SeriesMetadata('y',1,False,False,False),
+                  _pp.SeriesMetadata('shiftX',1,False,False,False),
+                  _pp.SeriesMetadata('shiftY',1,False,False,False),
+                  _pp.SeriesMetadata('connectionShiftX',1,False,False,False),
+                  _pp.SeriesMetadata('connectionShiftY',1,False,False,False)]
```
flo-dup: the parameters about the legend should be more explicit, something like legendShiftX, legendConnectionShiftX?

flo-dup (commented):
Could you also update the corresponding documentation?

flo-dup (approved):

geofjamg (changes requested):
fixed_positions should be also added to write_network_area_diagram_svg

geofjamg (approved):

PR #936, opened 2025-01-20
Update open rao to 6.2.1

phiedw:

No

Update open rao to 6.2.1

using version 6.0

using version 6.2.1

- [ ] Yes
- [x] No

**Other information**:


java/pypowsybl/src/main/java/com/powsybl/python/rao/RaoCFunctions.java
```diff
@@ -127,8 +125,11 @@ public static PyPowsyblApiHeader.ArrayPointer<CCharPointer> serializeRaoResultsT
             RaoResult raoResult = ObjectHandles.getGlobal().get(raoResultHandle);
             Crac crac = ObjectHandles.getGlobal().get(cracHandle);
             try (ByteArrayOutputStream output = new ByteArrayOutputStream()) {
-                new RaoResultJsonExporter().exportData(
-                    raoResult, crac, Set.of(Unit.MEGAWATT), output);
+                ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
```
obrix: The variable outputstream seems to be unused.

java/pypowsybl/src/main/java/com/powsybl/python/rao/RaoCFunctions.java
```diff
@@ -127,8 +125,11 @@ public static PyPowsyblApiHeader.ArrayPointer<CCharPointer> serializeRaoResultsT
...
+                Properties properties = new Properties();
+                properties.setProperty("rao-result.export.json.flows-in-amperes", "true");
```
obrix, line 129: Do we really want to hardcode those two properties ?

geofjamg: Cleaned

java/pypowsybl/src/main/java/com/powsybl/python/rao/RaoCFunctions.java
geofjamg: Removed

java/pypowsybl/src/main/java/com/powsybl/python/rao/RaoCFunctions.java
geofjamg: Discussed with @phiedw. We let it as it is for this version and maybe later on add a parameter.

geofjamg (approved):

obrix (approved):

PR #938, opened 2025-01-29
Switch snapshot ci to new app and rework notification design

rolnico:

No

Feature

CI uses some action to format slack notification
image

Better format
image

- [ ] Yes
- [x] No

- [X] The *Breaking Change* or *Deprecated* label has been added
- [X] The migration steps are described in the following section

Existing integration branches for the current snapshot version of powsybl-core should be renamed by adding the prefix `ci/core-`. (For instance, `6.7.0-SNAPSHOT` => `ci/core-6.7.0-SNAPSHOT`).

**Other information**:

~~:warning: **Important note:** Before this PR, the integration branches used for each repo were called with the snapshot version of powsybl-core (for instance `6.6.0-SNAPSHOT`). Now their name should be prefixed with `integration/powsyblcore-`. In our previous example, the branches must be called `integration/powsyblcore-6.6.0-SNAPSHOT`.
As a consequence, only maintainers can create these branches or force-push into them (see [the branching strategy documentation](https://github.com/powsybl/.github/blob/main/BRANCHING_RELEASE_STRATEGY.md#integration-branches)).~~


clementleclercRTE (dismissed):

olperr1:
Once merged, powsybl/powsybl-dev-tools#120 should be adapted to reflect the integration branch naming pattern change.

.github/workflows/snapshot-ci.yml
```diff
@@ -60,151 +66,234 @@ jobs:
...
+      - name: Checkout powsybl-dynawo sources
+        uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
         with:
           repository: powsybl/powsybl-dynawo
-          ref: ${{ env.SNAPSHOT_BRANCH }}
+          ref: ${{ env.INTEGRATION_BRANCH }}
           path: powsybl-dynawo
-      - name: Change core version
+
+      - name: Change core version in pom.xml in pom.xml
```
olperr1:
```suggestion
      - name: Change core version in pom.xml
```

.github/workflows/snapshot-ci.yml
```diff
@@ -60,151 +66,234 @@ jobs:
           fi
           echo "SCRIPTS_PATH=$SCRIPTS_PATH" >> $GITHUB_ENV  
 
-      #BUILD CORE
+      # Build powsybl-core on main branch
       - name: Checkout core sources
-        uses: actions/checkout@v4
+        uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
```
olperr1: The workflow previously worked with the default v4 version (so v4.2.2).
Did something prevents to use this version?

.github/workflows/snapshot-ci.yml
```diff
@@ -217,73 +306,250 @@ jobs:
...
         working-directory: ./pypowsybl/tests
         run: python3 basic_import_test.py
+
       - name: Run tests
         working-directory: ./pypowsybl/tests
         run: pytest
+
       - name: Upload wheel
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # 4.3.3
```
olperr1: Same with this action:
The workflow previously worked with the default v4 version (so v4.6.0).
Did something prevents to use this version?

olperr1 (changes requested):

.github/workflows/snapshot-ci.yml
rolnico: Nope, I just put the same version we use on some other repos but we surely can update it

.github/workflows/snapshot-ci.yml
rolnico: same answer

.github/workflows/scripts/check_integration_branch.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+repo=$1
+core_version=$2
+
+# Add "-SNAPSHOT" to powsybl-core version if not already there
+core_snapshot_version=$(echo "$core_version" | grep -q SNAPSHOT && echo "$core_version" || echo "$core_version-SNAPSHOT")
+
+# Find if an integration branch exists
+INTEGRATION_BRANCH=$(git ls-remote --heads "$repo" | grep -E "refs/heads/integration/powsyblcore-$core_snapshot_version" | sed 's/.*refs\/heads\///')
```
olperr1: After consulting the main repositories' maintainers, it was decided to keep non-protected branches for the CI snapshot process.
Can you change this script to use `refs/heads/ci/core-$core_snapshot_version` instead  ("integration" -> "ci" + "powsyblcore" -> "core") and to remove references to "integration"?

olperr1 (changes requested):

olperr1 (approved):

PR #942, opened 2025-02-07
adds a branch_labels df param to the get NAD and write NAD methods

CBiasuzzi:

no

feature

No label customization available

Adds a branch_labels parameter  to the get_network_area_diagram and  write_network_area_diagram.
Allows you to set branch labels (on the branch and on the branch edges) from the branch_labels dataframe parameter.

- [ ] Yes
- [x] No


java/pypowsybl/src/main/java/com/powsybl/python/network/NetworkCFunctions.java
```diff
@@ -1189,16 +1197,88 @@ private static void applyFixedPositions(DataframePointer fixedPositions, NadPara
...
+                }
+                return Optional.of(new EdgeInfo("ActivePower", arrowDirection, null, label));
+            }
+
+            @Override
+            public String getLabel(Edge edge) {
+                CustomBranchLabels bl = branchLabels.get(edge.getEquipmentId());
+                return (bl != null) ? bl.middle : null;
+            }
+        };
```
flo-dup, line 1256: Could you open a PR on powsybl-diagram to add such a LabelProvider, with a constructor based on the given map? So that after next powsybl-diagram release we remove this code here.

docs/user_guide/network_visualization.rst
```diff
@@ -314,7 +314,29 @@ We can generate a network area diagram using fixed positions, defined in a dataf
...
+                                                  ('LINE_S3S4', 'L2_1', 'L2_2', 'L2', 'OUT', 'IN'),
+                                                  ('TWT', 'TWT1_1', 'TWT1_2', 'TWT1', 'OUT', 'OUT')
+                                              ])
+    >>> pars=pp.network.NadParameters(edge_name_displayed=True)
+    >>> network.get_network_area_diagram(voltage_level_ids='S1VL1', depth=2, nad_parameters=pars, branch_labels=labels_df)
+
+    In the dataframe:
+    - id is the branch id
+    - side1 and side2 define the labels along the two branch's edges
+    - middle defines the branch's label
```
flo-dup, line 338: arrow1/arrow2 description is missing

docs/user_guide/network_visualization.rst
```diff
@@ -314,7 +314,29 @@ We can generate a network area diagram using fixed positions, defined in a dataf
...
+
+.. code-block:: python
+
+    >>> import pandas as pd
+    >>> network = pp.network.create_four_substations_node_breaker_network()
+    >>> labels_df = pd.DataFrame.from_records(index='id', columns=['id', 'side1', 'side2', 'middle', 'arrow1', 'arrow2'],
+                                              data=[
+                                                  ('LINE_S2S3', 'L1_1', 'L1_2', 'L1', 'IN', 'IN'),
+                                                  ('LINE_S3S4', 'L2_1', 'L2_2', 'L2', 'OUT', 'IN'),
+                                                  ('TWT', 'TWT1_1', 'TWT1_2', 'TWT1', 'OUT', 'OUT')
```
flo-dup: maybe add a null value for an arrow to show this possibility

docs/user_guide/network_visualization.rst
```diff
@@ -314,7 +314,29 @@ We can generate a network area diagram using fixed positions, defined in a dataf
...
-    
+
+
+We can set the labels of the NAD branches and the arrows direction, using a dataframe:
+
+.. code-block:: python
+
+    >>> import pandas as pd
+    >>> network = pp.network.create_four_substations_node_breaker_network()
+    >>> labels_df = pd.DataFrame.from_records(index='id', columns=['id', 'side1', 'side2', 'middle', 'arrow1', 'arrow2'],
```
flo-dup: order 'side1', 'middle', 'side2' would be more logical?

pypowsybl/network/impl/network.py
```diff
@@ -403,9 +406,19 @@ def _create_nad_positions_c_dataframe(self, df: DataFrame) -> _pp.Dataframe:
                   _pp.SeriesMetadata('legend_connection_shift_y',1,False,False,False)]
         return _create_c_dataframe(df, nad_positions_metadata)
 
+    def _create_nad_branch_labels_c_dataframe(self, df: DataFrame) -> _pp.Dataframe:
+        nad_positions_metadata=[_pp.SeriesMetadata('id',0,True,False,False),
+                  _pp.SeriesMetadata('side1',0,False,False,False),
+                  _pp.SeriesMetadata('side2',0,False,False,False),
+                  _pp.SeriesMetadata('middle',0,False,False,False),
```
flo-dup: same order question

pypowsybl/network/impl/network.py
```diff
@@ -416,6 +429,7 @@ def get_network_area_diagram(self, voltage_level_ids: Union[str, List[str]] = No
             low_nominal_voltage_bound: low bound to filter voltage level according to nominal voltage
             nad_parameters: parameters for network area diagram
             fixed_positions: optional dataframe used to set fixed coordinates for diagram elements. Positions for elements not specified in the dataframe will be computed using the current layout.
+            branch_labels: optional dataframe used to set the labels on a branch
```
flo-dup: next step should be the other parameters of the `LabelProvider`: a dataframe with buses description (bus id / description), a dataframe with voltage level description (vl id / description / details). And then comes the dataframe for the styles. Should we encapsulate all these dataframes in a bigger config object, or would it be ok to have these four dataframes as input at the end?

flo-dup (changes requested):
Nice work!

flo-dup:
After some discussing and thinking, could you encapsulate the branch_labels dataframe in an object? I think we could call it something like diagram customizer.

Indeed, at the end there will be specific customization for specific usecases, each one customizing all the dataframes mentioned in above comment. To that end, we will provide for each usecase a method to build this object. This object would make it easier to use those predefinite customization than if it remains in separated arguments.

flo-dup (approved):

geofjamg (dismissed):
Very nice features.
A few issues/remarks:
  - I don't like NadCustomizer naming. We should use something more business like NadProfile or NadUserProfile for example.
  - One thing that is quite disturbing is that with default NadParameter, even if specifying a middle label we cannot see anything and we need to activate manually a badly named parameter to make it drawn. When using the profile mode (super user one!) we whould not rely on parameters link to the default profile. What has been filled  in the dataframe has to be shown in the diagram.
  - Maybe we should think about a refactoring to move some of the NadParameters that are related to default profile and that would not make sens anymore in a user profile.
  - Another disturbing thing is that, we cannot stack the user profile to default profile. That would make sense for a user to only update some of the element of the network and rely on default profile for the other one. I propose in an other PR to create a function like `create_branch_labels_dataframe_from_default_profile`or something like this to be able to start and update a user profile for default one.

CBiasuzzi:
Thanks for your feedback

- NadCustomizer has been renamed to NadProfile.
- Now EdgeNameDisplayed (the parameter that prevents the middle label  from appearing, when set to False) is forced to True - when the labels dataframe is set.

PR #994, opened 2025-04-24
[WIP] Add mapping of scalable network modification to python 

### LINKED ISSUE #816: Add the mapping to python of scalable network modification

_No response_

The general idea is to be able to use scalable network modifications available in powsybl in pypowsybl.

The means to use scalable in powsybl-core that we want to use are defined in static methods in the `Scalable` class :  the `proportional`, `stack`, `upDown`...
We should be able for each to call a python method, with a ScalingParameters object, that will create the corresponding java Scalable object and apply its scale method.
For example, `scale_injections_proportionaly(list_injections, scaling_parameters, asked_scaling_value, distribution)` should call `proportional(List<Injection> injections, ...)` to create a scalable on java side, then call on it the `scale` method with the scaling_parameters and asked_scaling_value.
The python method should return the double value of the actual power adjustment

### Describe the motivation

_No response_

### Extra Information

_No response_

### LINKED ISSUE #993: Scalable in pypowsybl : first POC

_No response_

We should try at first to add only one type of scalable modification (for example generators), without considering the ScalingParameters java object.

To do so :
- create a `Scalable` python object that would be parametrized by a type, an id and limits (min/max)
- Creating one such object should create its java counterpart to be used as JavaHandle by python
- This class should have a `scale` method that takes a network and a value and apply the java scalable to the network 

### Describe the motivation

_No response_

### Extra Information

_No response_

nao1345678:

Fixes #816 
Fixes #993


HugoKulesza:
Evolution reimplemented in #1159 (with ScalingParameters mapping obtained from this PR)

PR #1035, opened 2025-07-04
Limit update

### LINKED ISSUE #1021: Operational limits value update

We can create a new operational limit but not update an existing one like for instance update the limit value.

We should be able to update an existing operational limit.

### Describe the motivation

_No response_

### Extra Information

_No response_

HugoKulesza:

Fixes #1021 

- [x] Yes
- [ ] No

The dataframe created by the `get_operational_limits` method has now a multi-index that contains the columns `element_id`, `side`, `type`, `acceptable_duration`, `group_name`, instead of only the `element_id`.
It can break dataframe manipulation because all existing `.loc["<element>"]` on the dataframe will now return a dataframe with the same lines, but without the `element_id` column.


pypowsybl/network/impl/network.py
```diff
@@ -4055,6 +4030,39 @@ def get_operational_limits(self, all_attributes: bool = False, attributes: Optio
...
+
+        See Also:
+            :meth:`get_operational_limits`
+
+        Examples:
+            Some examples using keyword arguments:
+
+            .. code-block:: python
+
+                network.update_generators(id='G-1', connected=True, target_p=500)
```
obrix: Maybe to replace with an operational limit example

obrix (changes requested):
A minor comment but it looks good to me.

obrix (approved):

PR #1063, opened 2025-07-30
OpenRao : fix search tree parameters and load flow parameters handling

Godelaine:

No

Costly min margin parameters extension was added to openrao (java) but not included in pypowsybl, breaking Costly section in tutorial notebook.

In Costly section in openRao tutorial notebook (see pypowsybl-notebooks), null ptr exception is thrown in ObjectiveFunctionCreator::getVirtualCostEvaluators when trying to fetch getMinMarginsParameters

costly min margin parameters defined. All search tree parameters are now taken into account.

- [ ] Yes
- [ ] No

**Other information**:


obrix:
@Godelaine Thanks it is clearer now that I see the fix, it seems I indeed did not create the extension properly when I moved to the updated parameters API with parts of the parameters moved into extensions.

For the mentionned issue with load flow provider specific parameters, I think properly creating the search tree extension is part of the solution, but the load flow parameters conversion process itself should also be updated so that SensitivityAndLoadFlowParameters are properly handled.

If you agree I will add the fix here, so I can also add a proper "roundTrip" equivalent test so this PR fix everything parameters related.

As for the python parameters model, it reflects the previous RaoParameters structure as I initially based my development on a previous rao version and the rao version was updated just before release. Now that a lot of parameters have been moved into extensions the python model can probably be simplified a bit. But this can be done later in another pull request.

obrix:
@Godelaine Well open load flow seems to not be able to reload its extracted provider properties out of the box (see test error with reportedFeatures property). Open load flow may need a patch, I'll look into it.

obrix:
Waiting for the OLF version containing the necessary fix to finish the PR.

HugoKulesza (approved):

PR #1069, opened 2025-08-28
Factorize CI, to only build 1 time by OS the native libraries and reuse them for each python

HugoKulesza:

No

Technical change

For the full CI, we currently rebuild the native libraries for every pair (OS, Python version) (so 20 builds in total).
They are actually only dependant on the OS, so could be built only once for each one (only 4 builds left).
Since the native libraries build takes most of the total build time and resources, it will also allow us to greatly reduce both of these.

- [ ] Yes
- [x] No


.github/workflows/dev-ci.yml
```diff
@@ -111,12 +111,12 @@ jobs:
             macosx_deployment_target: "10.16", # to ensure pip finds wheel when Big Sur is configured to return 10.16 as version instead of 11.0
```
obrix, line 111: No change on this file  ? Maybe to restore ?

obrix (dismissed):
Minor comment, but it looks good to me.

.github/workflows/dev-ci.yml
HugoKulesza: I actually had to remove the mac arm builds here, we will stop making them since we have now to pay for them (on large mac runners)

obrix (approved):

PR #1081, opened 2025-09-23
[RAO] Split remedial action results in different dataframes

### LINKED ISSUE #1080: Separate Remedial Actions Results in different dataframes

Currently, all the remedial actions results are gathered in a single dataframe that contains the following columns:
- remedial_action_id (string)
- optimized_instant (string)
- contingency (string)
- activated (boolean)
- optimized_tap (int)
- optimized_set_point (float)

This can be accessed using the `get_ra_results` method.

To be more in line with how remedial actions are handled in OpenRAO, it make ssense to split the remedial actions results in several dataframes depending on the type of remedial actions being dealt with.

The `get_ra_results` method can be kepts to list all the remedial actions activated during RAO, without further information. The columns of the returned dataframe would be:
- remedial_action_id (string)
- optimized_instant (string)
- contingency (string)

> - There is no need for the `activated` column since only the activated remedial actions should reported.
> - Renaming this method to `get_remedial_action_results` would be cleaner

For **network actions** (i.e. binary actions), a specific method called `get_network_action_results` should be implemented to return the information about the state at which they were activated. The columns of the returned dataframe would be:
- remedial_action_id (string)
- optimized_instant (string)
- contingency (string)

For **PST range actions**, a specific method called `get_pst_range_action_results` should be implemented to return the information about the state at which they were activated and the optimal tap. The columns of the returned dataframe would be:
- remedial_action_id (string)
- optimized_instant (string)
- contingency (string)
- optimized_tap (int)

For other **range actions**, a specific method called `get_range_action_results` should be implemented to return the information about the state at which they were activated and the optimal set-point. The columns of the returned dataframe would be:
- remedial_action_id (string)
- optimized_instant (string)
- contingency (string)
- optimized_set_point (float)

### Describe the motivation

_No response_

### Extra Information

_No response_

bqth29:

Fixes #1080 

- [ ] Yes
- [ ] No

**Other information**:


java/pypowsybl/src/main/java/com/powsybl/python/rao/RaoDataframes.java
```diff
@@ -67,7 +83,13 @@ public record AngleCnecResult(String cnecId, Instant instant, String contingency
 
     public record VoltageCnecResult(String cnecId, Instant instant, String contingency, TwoSides side, double minVoltage, double maxVoltage, double margin) { }
 
-    public record ActivatedRemedialActionResult(String remedialActionId, Instant instant, String contingency, double optimizedTap, double optimizedSetPoint) { }
+    public record ActivatedRemedialActionResult(String remedialActionId, Instant instant, String contingency) { }
```
obrix, line 86: Maybe use a single "ActivatedActionResult" record, instead of two identical records

obrix (changes requested):
Looks good to me, just a minor comment.

obrix (approved):

PR #1082, opened 2025-09-23
Perform round trip on RAO result

bqth29:

When running a RAO, if only the preventive state was optimized, a [OneStateOnlyRaoResultImpl](https://github.com/powsybl/powsybl-open-rao/blob/main/ra-optimisation/search-tree-rao/src/main/java/com/powsybl/openrao/searchtreerao/result/impl/OneStateOnlyRaoResultImpl.java) object is returned by OpenRAO. When calling `get_ra_results`, if curative RAs are defined in the CRAC, an exception is thrown because results based on curative state a queried whereas the RAO Result only supports preventive results.

Running a round trip on the RAO result forces the RAO to export a [RaoResultImpl](https://github.com/powsybl/powsybl-open-rao/blob/main/data/rao-result/rao-result-impl/src/main/java/com/powsybl/openrao/data/raoresult/impl/RaoResultImpl.java) object all the time which fixes the issue.

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/java/pypowsybl/src/main/java/com/powsybl/python/rao/RaoContext.java
+++ b/java/pypowsybl/src/main/java/com/powsybl/python/rao/RaoContext.java
@@ -11,6 +11,7 @@ import com.powsybl.commons.PowsyblException;
 import com.powsybl.glsk.api.GlskDocument;
 import com.powsybl.iidm.network.Network;
 import com.powsybl.loadflow.LoadFlowParameters;
+import com.powsybl.openrao.commons.OpenRaoException;
 import com.powsybl.openrao.data.crac.api.Crac;
 import com.powsybl.openrao.data.raoresult.api.RaoResult;
 import com.powsybl.openrao.monitoring.Monitoring;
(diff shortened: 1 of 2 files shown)


obrix:
Hello @bqth29, thanks for this PR. Is this a temporary workaround ? It seems to me being forced to serialize and deserialize again the rao result to avoid an exception means there is probably an issue in open rao itself ? Is there a fix being prepared on open rao side ?

bqth29:
> Hello @bqth29, thanks for this PR. Is this a temporary workaround ? It seems to me being forced to serialize and deserialize again the rao result to avoid an exception means there is probably an issue in open rao itself ? Is there a fix being prepared on open rao side ?

Hi @obrix, indeed your point is relevant. The origin of this problem is when the RAO optimization only goes through one optimization state (the preventive state) but CRAs are defined. The RAO Result implementation that is returned will throw exceptions as soon as a result is queried on a state which is not the optimization state. Yet, the user could wish to know if the CRA was applied since they may not be aware of the different steps carried out by the optimizer.

We discussed with the OpenRAO team and we agree that it is quite harsh to proceed this way. We will update our implementation to return relevant default values instead. We have opened an issue for this purpose and will address it after the upcoming October release: https://github.com/powsybl/powsybl-open-rao/issues/1476.

Thanks for your review!

PR #1085, opened 2025-09-25
Force RAO implementation to "SearchTreeRao"

bqth29:

In OpenRAO's next release, a new RAO implementation called "FastRao" will be available. Currently, we load the default implementation which is "SearchTreeRao". However, with the two implementations, we get the following error:

```
Several RAO implementations found ([SearchTreeRao, FastRao]), you must add configuration to select the implementation
```

FastRAO will possibly be included in future PyPowSyBl versions. In the meantime, we force the RAO implementation to "SearchTreeRao" to avoid issues.

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/java/pypowsybl/src/main/java/com/powsybl/python/rao/RaoContext.java
+++ b/java/pypowsybl/src/main/java/com/powsybl/python/rao/RaoContext.java
@@ -50,7 +50,7 @@ public class RaoContext {
         if (glsks != null) {
             inputBuilder.withGlskProvider(glsks.getZonalGlsks(network));
         }
-        return Rao.run(inputBuilder.build(), parameters);
+        return Rao.find("SearchTreeRao").run(inputBuilder.build(), parameters);
     }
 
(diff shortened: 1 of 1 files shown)


Godelaine (approved):

obrix (changes requested):
Hello @bqth29, even if FastRao is not already fully supported I think we can add the provider option in the runRao method and set the "SearchTreeRao" as the default one.
This can be done pretty quickly and in a very similar way to what is done with the runLoadFlow method and its provider option.

bqth29:
> Hello @bqth29, even if FastRao is not already fully supported I think we can add the provider option in the runRao method and set the "SearchTreeRao" as the default one. This can be done pretty quickly and in a very similar way to what is done with the runLoadFlow method and its provider option.

Hello @obrix, indeed this is a better way to proceed. I implemented your solution in #1086 if you want to have a look at it.

PR #1102, opened 2025-10-14
Battery voltage regulation extension support

geofjamg:

No

Feature

We support the batteries `VoltageRegulation` extension

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/java/pypowsybl/src/main/java/com/powsybl/dataframe/network/extensions/VoltageRegulationDataframeProvider.java
+++ b/java/pypowsybl/src/main/java/com/powsybl/dataframe/network/extensions/VoltageRegulationDataframeProvider.java
@@ -1,0 +1,75 @@
+/**
+ * Copyright (c) 2025, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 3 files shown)


java/pypowsybl/src/main/java/com/powsybl/dataframe/network/extensions/VoltageRegulationDataframeProvider.java
```diff
@@ -0,0 +1,75 @@
...
+                .filter(Objects::nonNull);
+    }
+
+    private VoltageRegulation getOrThrow(Network network, String id) {
+        Battery battery = network.getBattery(id);
+        if (battery == null) {
+            throw new PowsyblException("Battery '" + id + "' not found");
+        }
+        return battery.getExtension(VoltageRegulation.class);
+    }
```
HugoKulesza, line 56: I have a small problem here (that, after checking, also exists in other extensions). If someone, on a battery that exists in the network, tries to update its extension that does not exist, this person will get a NullPointerException.
For some other extensions (like GeneratorShortCircuit) we also check that the extension is not null in order to throw a more understandable error.

HugoKulesza (changes requested):

HugoKulesza (approved):

PR #1166, opened 2026-03-05
Connect/disconnect with mode

### LINKED ISSUE #1163: Parametrize "connect/disconnect" to select which switch to move

The current connect/disconnect methods only allow the real breakers to be used (to be coherent with the "operational" way of work).

We should have parameters for these methods in order to select the mode of connection (from the SwitchPredicate on java side)

### Describe the motivation

_No response_

### Extra Information

_No response_

HugoKulesza:

Fixes #1163 

- [ ] Yes
- [x] No


pypowsybl/network/impl/network.py
```diff
@@ -185,11 +185,33 @@ def open_switch(self, id: str) -> bool:
     def close_switch(self, id: str) -> bool:
         return _pp.update_switch_position(self._handle, id, False)
 
-    def connect(self, id: str) -> bool:
-        return _pp.update_connectable_status(self._handle, id, True)
+    def connect(self, id: str, operate_disconnectors: bool = False, operate_fictitious: bool = False) -> bool:
```
colineplqt, line 188: Having operate_disconnectors and operate_fictitious as False is breaking change no? Before every switch would have been closed to connect an equipment no?

HugoKulesza: The predicate used by default in powsybl-core is `IS_NONFICTIONAL_BREAKER` for connection so I don't think so (this change is actually made just so that people can force the change on fictitious or disconnectors which would not have been possible before)

colineplqt: Ok, nice!

colineplqt (approved):

PR #1191, opened 2026-04-09
Add NAD parameters

rolnico:

No

Feature

Several NAD parameters are missing

The following NAD parameters have been added:
- scale_factor
- timeout_seconds
- edge_info_included
- voltage_level_legends_included

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/pypowsybl/network/impl/nad_parameters.py
+++ b/pypowsybl/network/impl/nad_parameters.py
@@ -25,7 +25,8 @@ class NadParameters:
                  scaling_factor: int = 150000, radius_factor: float = 150.0,
                  edge_info_displayed: Optional[EdgeInfoType] = None, voltage_level_details: bool = True,
                  injections_added: bool = False,
-                 edge_info_parameters: Optional[EdgeInfoParameters] = None):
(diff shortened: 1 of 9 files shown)


pypowsybl/network/impl/nad_parameters.py
```diff
@@ -43,6 +45,10 @@ def __init__(self, edge_name_displayed: Optional[bool] = None, id_displayed: boo
         self._check_edge_info_parameters(edge_name_displayed=edge_name_displayed,
                                          edge_info_displayed=edge_info_displayed,
                                          edge_info_parameters=edge_info_parameters)
+        self._scale_factor = scale_factor
```
HugoKulesza, line 48: Is it really necessary to add this when we already have a "scaling_factor" parameter in the class (or are these just too closely named despite being used for different things) ?

rolnico: I tried, but they are not the same, and not even the same type (double vs int)...

HugoKulesza:
Tried and tested on local build : all three new parameters do work.

HugoKulesza (approved):

PR #1208, opened 2026-04-28
Add DcSwitch API in PyPowSyBl.

### LINKED ISSUE #1176: Support enhanced DC circuit breaker

There is a DcSwitch class, with on/off states but no resistance. A DC circuit breaker with resistance is requested in PowSyBl Core https://github.com/powsybl/powsybl-core/issues/3827

It is possible to create and manipulate in PowSyBl the DC breaker requested in https://github.com/powsybl/powsybl-core/issues/3827

### Describe the motivation

Resistance of breakers are essential to analyze losses as well as security in AC/DC load flow analyses.

### Extra Information

Proposed solution in PowSyBl Core: add a resistance parameter in class DcSwitch.
This will be reflected in the Python API of DcSwitch.

SGI-Landry:
It already takes the resistance property into account.

Fixes #1176

Add support for DcSwitch in PyPowSyBl
- main operations for dc_switches: open_dc_switch, close_dc_switch, get_dc_switches, update_dc_switches and create_dc_switches.
- this is reflected in the Java adder and the C binding.
- It supposes the existence of resistance in the DcSwitch model in PowSyBl-core.

- [ ] Yes
- [x] No


SGI-Landry:
Hello @bperr can you please review this activity? Beware that it won't compile unless you build it against the `main` version of PowSyBl core.

SGI-Landry:
Hello @HugoKulesza, can you please start reviewing this activity? As mentioned over Slack, this can be tested against the `main` branch of PowSyBl-Core, but it cannot be merged before the June release train.

SGI-Landry:
Hello @HugoKulesza did you have the chance to review this activity ?

HugoKulesza:
Hi @SGI-Landry, I did a first pass on it and I didn't see any major problem with the PR. For the complete review we'll wait for dependencies so that we have the build and sonar as support. Do not worry, since I know this PR must be in the next release (and is tagged accordingly) I'll take the time needed before releasing to ensure it passes.

SGI-Landry:
> Hi @SGI-Landry, I did a first pass on it and I didn't see any major problem with the PR. For the complete review we'll wait for dependencies so that we have the build and sonar as support. Do not worry, since I know this PR must be in the next release (and is tagged accordingly) I'll take the time needed before releasing to ensure it passes.

Thanks for this first feedback. Now I can wait for the release train and not worry about anticipating large required refactoring.

java/pypowsybl/src/main/java/com/powsybl/python/network/NetworkCFunctions.java
```diff
@@ -1190,6 +1189,19 @@ public boolean getAsBoolean() {
...
+        return doCatch(exceptionHandlerPtr, new BooleanSupplier() {
+            @Override
+            public boolean getAsBoolean() {
+                Network network = ObjectHandles.getGlobal().get(networkHandle);
+                String idStr = CTypeUtil.toString(id);
+                return NetworkUtil.updateDcSwitchPosition(network, idStr, open);
+            }
+        });
+    }
+
```
HugoKulesza, line 1204: We could avoid adding a whole new method here, and in every cpp files, by just adding a "dc" boolean parameter to the existing `updateSwitchPosition` and using it to call the right NetworkUtil method underneath.

pypowsybl/network/impl/network.py
```diff
@@ -225,6 +225,36 @@ def open_switch(self, id: str) -> bool:
...
+        Args:
+            id: the identifier of the DC switch to open.
+
+        Returns:
+            ``True`` if the switch state changed (was closed before), ``False`` if it was already open.
+
+        See Also:
+            :meth:`close_dc_switch`, :meth:`get_dc_switches`
+        """
+        return _pp.update_dc_switch_position(self._handle, id, True)
```
HugoKulesza, line 241: By not creating a new method and only adding a parameter, you would have here :

```suggestion
        return _pp.update_switch_position(self._handle, id, True, True)
```
(with the second boolean referencing to "dc")

SGI-Landry:
@HugoKulesza thanks for your review. I'm not very keen on changing the philosophy around DC equipment at this stage. The approach was initiated by @jeandemanged in July 2025. It results in a bit more code, but it avoids confusion between AC and DC equipment, which are modeled very differently anyway. As an added benefit, it allows us to distinguish between AC disconnection and DC disconnection for ACDC converters.

HugoKulesza (approved):
After discussing with @SGI-Landry, the discussions about the technical choices of implementation in the C/C++ parts can be treated later. The API in python is correct and will not have to be changed anyway and it's the more important validation.

(+ the subject of discussion also encompasses former methods that have a direct mapping via C/C++ such as the similar `update_switch_position`)

PR #1209, opened 2026-04-28
bus to bus switch import for pandapower import

pawellytaev:

- [ ] Docs have been added / updated (for bug fixes / features) - _EDIT: there is actually no doc on the pandapower import yet_

No

import bus to bus switches from pandapower

switches are ignored leading to buses being disconnected

switches are imported, so all bus to bus switches can be imported

- [ ] Yes
- [x] No

**Other information**:

For now, only bus to bus switches are considered. Line switches and transformer switches from pandapower can be done in a next step (by adding an additional auxiliary bus in powsybl and then re-connecting the line and the switch accordingly)


pypowsybl/network/impl/pandapower_converter.py
```diff
@@ -44,10 +44,24 @@ def convert_from_pandapower(n_pdp: pandapowerNet) -> Network:
     create_shunts(n, n_pdp)
     create_lines(n, n_pdp)
     create_transformers(n, n_pdp)
+    create_switches(n, n_pdp)
     create_extensions(n, slack_weight_by_gen_id)
 
     return n
 
+def create_switches(n: Network, n_pdp: pandapowerNet) -> None:
```
HugoKulesza: I would prefer to call it `create_switches_between_buses` (and remove the comment about line_switches) so that later it is clearer that not all switches are currently imported.

HugoKulesza (changes requested):
Just a little naming change, but otherwise it's all good, than you for the PR !

pypowsybl/network/impl/pandapower_converter.py
pawellytaev: done, thanks for the review :-)

HugoKulesza (approved):

PR #1223, opened 2026-05-28
feat: Add a binding for the java switchflow functionality

nicow-elia:

No, just a powsybl [slack post](https://powsybl.slack.com/archives/C03DUM4UJCW/p1779168059766819).

A feature which wraps the java SwitchFlow functionality in pypowsybl, so flows over a switch can be retrieved.

There is no such binding.

A binding exists.

- [ ] Yes
- [x] No


HugoKulesza (commented):
Hello, thank you for your PR ! 
Is there a reason to not add the P and Q values as new columns in the dataframe retrieved by network.get_switches(), so as to not create a whole new dataframe only for this ?

nicow-elia:
My thought why to add this separately is because adding it to get_switches would add overhead to the function call. There is a relatively heavy graph walk behind the switch flow method, while get_switches just returns "static" data. But if you prefer, I can also add it to get_switches with an additional argument boolean, maybe something like ```net.get_switches(return_switchflow=True)```?

HugoKulesza:
Ok I understand the performance issue. 
Since all methods call get_element underneath, I prefer what you did (a new dataframe) to a change in this method with a new parameter.
So it's all good for the design i'll do another pass with more detail

HugoKulesza (approved):

PR #1225, opened 2026-06-04
feat: Bind voltage angle violations functionality

nicow-elia:

It binds the voltage angle limit functionality that is already existing in the java world through python. Now, the user can not only add operationalLimits but also voltageAngleLimits. It adds a binding for listing, creation and deletion and validates that the security analysis respects these.

Voltage angle limits are not available from pypowsybl.

Voltage angle limits are available from pypowsybl.

- [ ] Yes
- [x] No

**Other information**:
I considered adding the voltage angle limits through the operational limits interface, but it seemed semantically too different. Preexisting limits only have one value and have a duration, which doesn't match the voltage angle limit specs of two values and no duration.
Also I wondered how to name the limits. Voltage Angle Limits are technically also operational limits, but from a different perspective of operations. E.g. when you operate a station in two-node configuration you do not want the voltage angle to drift too far so you can never close the coupler again. Nothing physically happens if it is violated, but it is still an operational concern that dictates this limit. Maybe we could rename "operational limits" to "thermal limits" or something like that? Should likely not be the scope of this PR though to execute that rename.


HugoKulesza:
I agree that we should think of something for the "operational_limits" methods, because it would be strange that voltage_angle limits are not included in it if the method is called this way. On powsybl-core side, the distinction between CURRENT/ACTIVE_POWER/APPARENT_POWER and VOLTAGE_ANGLE is that the first three are `LoadingLimits` on top of being `OperationalLimits` (and that's the class that defines the structure with groups and durations).
So I guess we can deprecate the "operational_limits" methods in favor of "loading_limits" ones, so that it is clearly split between them and the voltage_angle ones.

docs/user_guide/security.rst
```diff
@@ -37,6 +37,28 @@ the violations created by the contingency, they are collected by contingency in
                       NHV1_NHV2_2                   CURRENT  permanent   500.0           2147483647              1.0  1477.824335  TWO
                       VLHV1                     LOW_VOLTAGE              400.0           2147483647              1.0   392.158685
```
HugoKulesza:
```suggestion
                      VLHV1                     LOW_VOLTAGE              400.0           2147483647              1.0   392.158685     
```
I'm not sure why the formatting changed but that is the only difference I could find in the doctest error of the build

HugoKulesza (commented):
Everything's good for me, there's only the doctest that does not go through.

The renaming of operational limits methods to loading limits will be done in a next PR

HugoKulesza (approved):

PR #1227, opened 2026-06-09
fix Full CI for OPF Windows

HugoKulesza:

Bug fix : Full CI error due to missing url in Windows Ipopt install 

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -194,13 +194,13 @@ jobs:
           otool -L "$(brew --prefix ipopt)/lib/libipopt.dylib"
           echo "DYLD_LIBRARY_PATH=${IPOPT_LIB}:${DYLD_LIBRARY_PATH}" >> $GITHUB_ENV
 
-      - name: Install Miniconda and cyipopt (Windows)
+      - name: Install Ipopt (Windows, via miniconda)
         if: matrix.config.name == 'windows'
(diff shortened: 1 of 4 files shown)


.github/workflows/full-ci.yml
```diff
@@ -198,7 +198,7 @@ jobs:
         if: matrix.config.name == 'windows'
         shell: pwsh
         run: |
-          curl -fsSL ${{ matrix.config.miniconda-url }} -o miniconda.exe
+          curl -fsSL https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe -o miniconda.exe
```
olperr1: It would be more cautious to set the Miniconda version to use.
Also, the in the miniconda repo, there are different versions depending on the python version. I think you use the right version depending on the `matrix.python` (`cp310`, ...,  `cp313`). Unfortunately there's no `py314` version in the repo...

olperr1 (changes requested):

.github/workflows/full-ci.yml
HugoKulesza: miniconda is only used as an "easy" install for all the native libraries used by ipopt, and not directly to use it as the python interpreter in the CI, so setting a specific python version is not useful.

olperr1 (approved):

PR #1228, opened 2026-06-10
feat: Added a binding for trafo phase clocks

nicow-elia:

It introduces a binding for the two and three winding transformer clock extension that is present in powsybl core. 

Currently, transformer clocks are not bound, so python users have no way of accessing the information.

Now they can read and write the transformer clock values stored in the .iidm

- [ ] Yes
- [x] No

**Other information**:
I have two open questions:

1. How do I document this extension? I found no existing patterns in other extension except for the getExtensionInformation string that describes to the user what this extension does?
2. OpenLoadFlow currently does not respect the trafo clock values, meaning they currently have no impact on voltage angles in the loadflow. I would argue the read/write feature simply for data management already has a value, but if you want to wait with merging this PR until trafo clock values are effective I can fully understand this. Alternatively I would propose that, wherever the extension shall be documented in pypowsybl, I would mention that this is not wired through openloadflow yet.


HugoKulesza (commented):
Regarding your two questions :
- for the documentation, indeed all the extensions are only documented through the `get_extensions_information` method. One way to ease the understanding of extensions would be to add a small paragraph on "The network model" User Guide readthedocs page, that indicates how to interact with extensions in general (`create_extensions`, `get_extensions`, `update_extensions`  and `get_extensions_information` to list available ones). If you want to do that, it would be great !

- Since the extension is already importable on powsybl-core side, I think that being able to see it/modify it is a good thing in itself even if it is not used in OpenLoadFlow. It seems used somewhere in powsybl-core at least (even if I'm not sure to even understand what it represents 🙃). I'm not sure that everything that is usable in pypowsybl is well used in simulators.

nicow-elia:
... so my challenge is now to write trafo clock docs that are so good that you know afterwards what they do - challenge accepted! :) (they just have a hardwired angle shift and a weird notion how to represent them)

HugoKulesza:
> ... so my challenge is now to write trafo clock docs that are so good that you know afterwards what they do - challenge accepted! :) (they just have a hardwired angle shift and a weird notion how to represent them)

I just think that this documentation should be written in powsybl-core, I don't think I have seen it there

nicow-elia:
There are a few sentences - can I link them sensibly without the link going out of sync? https://powsybl.readthedocs.io/projects/powsybl-core/en/v6.4.0/grid_model/extensions.html#two-windings-transformer-phase-angle-clock

HugoKulesza:
I actually think that there should be a link in pypowsybl doc that points to the general page of extensions (https://powsybl.readthedocs.io/projects/powsybl-core/en/latest/grid_model/extensions.html) and the general methods `get_extensions_information` etc...) and not documenting separately every extension (since it is just a mapping, and it is more probable that the documentation will be up to date on powsybl-core where it is defined).

HugoKulesza (approved):

PR #1252, opened 2026-07-06
Direct copy of buffer in createDoubleMatrix instead of intermediate b…

gautierbureau:
…uffer copy.

I avoids a useless copy of data

It adds a useless copy

The copy is removed

- [ ] Yes
- [ ] No

**Other information**:


HugoKulesza (commented):
I understand that the copy in the former createDoubleMatrix method was not  necessary, but I'm not sure to understand why the use of a ByteBuffer is needed in the end ? Why was the former "for" loop not suitable to form directly the valuePtr object ?

PR #1264, opened 2026-07-20
Ground symmetrical monopole OPF test network

thomas-durand-SGI:

No.

Bug fix.

On Windows CI, `test_vsc_symmetrical_monopole` can non-deterministically return `ALMOST_LOCALLY_SOLVED` instead of `OPTIMAL_SOLUTION`.

The tested DC component has no connection to ground. Therefore, no equation fixes the DC-node voltages relative to zero: all DC voltages can shift by the same value without changing voltage differences, currents, or powers.

This undetermined voltage level is a possible cause of the numerical instability observed in `test_vsc_symmetrical_monopole`.

The test temporarily adds an ideal ground to `dcNodeFrNeg` before running the OPF. This provides an explicit DC-voltage reference and removes the undetermined voltage shift. This is a targeted workaround for the test, not a general implementation for ungrounded DC components.

Locally, before the change, the test returned `OPTIMAL_SOLUTION` after 13 iterations, entered IPOPT restoration at iterations 4 and 5, and ended with a dual infeasibility of approximately `0.13`. With the ground added, the test returns `OPTIMAL_SOLUTION` after 11 iterations, with a final constraint violation of approximately `6.7e-15` and a final dual infeasibility of approximately `6.0e-14`. 

Therefore, the final numerical result is significantly cleaner than before the change.

The intermittent CI failure was not reproduced locally. This change is intended to verify whether adding an explicit DC-voltage reference stabilizes the Windows CI result.

- [ ] Yes
- [x] No

### DIFF
--- a/tests/test_opf.py
+++ b/tests/test_opf.py
@@ -231,11 +231,25 @@ def test_micro_grid_nl():
 
 
 def test_vsc_symmetrical_monopole():
-    opf_parameters = OptimalPowerFlowParameters(mode=OptimalPowerFlowMode.ACDC)
-    assert pp.opf.run_ac(pp.network.create_dc_detailed_vsc_symmetrical_monopole_network(), opf_parameters)
-    
+    network = pp.network.create_dc_detailed_vsc_symmetrical_monopole_network()
+    # Temporary workaround: ground one pole because floating detailed-DC components are not yet supported reliably by the ACDC OPF.
+    dc_network = network.get_sub_network("VscSymmetricalMonopole")
+    dc_network.create_dc_grounds(
+        id="dcGroundFrNeg",
+        dc_node_id="dcNodeFrNeg",
+        r=0.0,
+    )
+    opf_parameters = OptimalPowerFlowParameters(
+        mode=OptimalPowerFlowMode.ACDC
+    )
+    assert pp.opf.run_ac(network, opf_parameters)
 
 '''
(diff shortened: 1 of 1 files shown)


thomas-durand-SGI:
Hi @HugoKulesza,

This PR attempts to fix the non-deterministic results of `test_vsc_symmetrical_monopole`. It adds an explicit DC-voltage reference by temporarily grounding one DC node, allowing us to check whether this stabilizes the Windows CI result.

If you also want to test the symmetrical monopole without adding a ground, the `fix-vsc-symmetrical-monopole` branch is available. It introduces an `enforce_dc_pole_symmetry` boolean, disabled by default and enabled only for this test.

When enabled, the OPF imposes:
```
V1 + V2 = 0
```
Combined with the existing constraint:
```
V1 - V2 = target_v_dc
```
this gives symmetrical pole voltages:
```
V1 = +target_v_dc / 2
V2 = -target_v_dc / 2
```

This alternative produced similar numerical improvements locally, reducing the final dual infeasibility from approximately `0.13` to `4.6e-11`.

HugoKulesza (approved):
Thanks for the fix tried here ! It indeed seems like a probable cause, and we'll see then how to permanently fix this after

thomas-durand-SGI:
Sounds good. I will monitor this feed in case the fix of this PR does not fully resolve the CI windows issue.

HugoKulesza:
Since we had merged the ACDC OPF, all Full CI builds had an error on at least one windows version. For the last 3 after this merged PR there was no issue (linked to these tests at least :upside_down_face:) so I think it's good !

PR #1273, opened 2026-08-04
feat: Direct Outage group lookup for equipment

nicow-elia:

No

It adds a feature with which the user can see for any equipment which outage group it belongs to before computing a security analysis. A parity test ensures equality with the contingency results, but this method is easier/faster than invoking a full security analysis as it only relies on `PropagatedContingency.createList` and the connectivity loss computation in `toLfContingency`.

For finding outage groups, the user has to invoke a security analysis

Added two methods:
* network.get_outage_group returns the outage group for a single piece of equipment
* network.get_outage_groups is a batched call to return outage groups for multiple pieces of equipment (e.g. all N-1 cases)

- [ ] Yes
- [x] No

**Other information**:
I have two things I am not sure about:

- I added a direct dependency to openloadflow in NetworkUtil.java - is this problematic? I have a feeling that the loadflow provider pattern is a bit broken through this change, but then I don't know how to avoid it.
- I found an unrelated memory leak in freeStringMap in CommonCFunctions.java and fixed it (the git diff is a bit ugly there, but before the pointers were not correctly freed). Is it foreseen to add something like a valgrind CI to check for memory leaks automatically? Or how can we ensure that we do not produce memory leaks in this handler code which is not nice to read already?


HugoKulesza:
Hi ! I started to look at this PR, and already have some returns regarding your last questions.

I also think we should not add this direct dependency on the OLF methods in pypowsybl. Actually, these methods in OLF are not really hard-linked to the simulator and could be moved to powsybl-core, and other java users have recently asked for such a move to be made. 
Do you have strong deadlines for this feature to be integrated in pypowsybl ?
I'm going to discuss it with powsybl-core maintainers to see how/when we could do that and decide then with this info and your return the best way to go forward (do this PR and adapt later with the powsybl-core move or wait for it before adding the feature to pypowsybl).

Regarding the memory leaks, I did not think about adding something to check in the CI (and do not really know how to do so) but it could be indeed useful.

nicow-elia:
Hey, thanks for your answer!

I do not have a deadline anymore, so a refactor is fine, we have just rebuild the graph walk internally in ToOp because we already didn't expect this PR to be merged by the time we need it. 

However, I would propose another option where to move these methods to: for the integration of the state estimator, a new component "OpenSteadyState" would be needed long term as the state estimator will need the same matrix building/equation toolkits as OLF. Contingency handling seems like a sensible task for that module as well. But if core seems a better fit for this specific outage group logic, then so be it - but then you should be careful how exactly the responsibility split is there.

And while writing, I have yet another idea. In python packages, it is common to present add-on packages. We could do the same with pypowsybl, that we'd have pypowsybl-core that is only hardwired to powsybl-core and has the loadflow provider discovery enabled, and a pypowsybl-olf that exhibits olf specifics - because i have the feeling that there is more demand for OLF specifics. The aforementioned matrix/equation toolkits for example would be a great thing to have for scientists to play around with and they currently sit in OLF.

When you are moving it though, I would be interested in another feature in the outage group logic. Currently, the logic only returns the elements *inside* an outage group, but it would also be extremely interesting to know the breakers *at the boundary* of an outage group. E.g. if I want to visually present how an outage unfolds in the grid to an operator, I would open the switches at the boundary.

HugoKulesza:
Hi ! in the end, another user has asked me for this feature in python and would use it as soon as it is available. After also discussing with other people I'm ok with integrating it with this kind of olf import, we can change this later if/when the refactor happens.
So I will now really review in detail the PR :slightly_smiling_face:

pypowsybl/network/impl/network.py
```diff
@@ -4919,6 +4919,87 @@ def get_bus_breaker_topology(self, voltage_level_id: str) -> BusBreakerTopology:
...
+                net.get_outage_group('LINE_S3S4')
+
+            will return:
+
+            .. code-block:: python
+
+                ['LD6', 'LINE_S3S4', 'SVC']
+
+        """
+        return _pp.get_outage_group(self._handle, equipment_id)
```
HugoKulesza, line 4959: I have a doubt that duplicating these methods to have one singular and one plural is necessary.
We tend to have one methods accepting str or List[str] as parameter with the underlying cpp methods only taking a list as argument (e.g. see `remove_elements`).

I admit that the difference in return types can be a good reason to have both methods, but we could then have both in python relying on a single one through the stack (in cpp/java) : 

```suggestion
        return _pp.get_outage_groups(self._handle, [equipment_id])[equipment_id]
```

java/pypowsybl/src/main/java/com/powsybl/python/network/NetworkUtil.java
```diff
@@ -199,6 +215,143 @@ static List<SwitchFlowContext> getSwitchFlowResults(Network network, List<String
...
+     * @return map keyed by initiating equipment identifier with sorted disconnected equipment identifiers as values,
+     * following the connectivity-result semantics used for outage-group computation
+     */
+    static Map<String, List<String>> getOutageGroups(Network network, List<String> equipmentIds) {
+        if (equipmentIds.isEmpty()) {
+            return Collections.emptyMap();
+        }
+
+        return computeOutageGroupsByEquipmentId(network, new HashSet<>(equipmentIds));
+    }
```
HugoKulesza, line 244: To complete what I wrote in network.py, I would not distinguish here between the two methods but only on python side to reduce the duplicated path.

java/pypowsybl/src/main/java/com/powsybl/python/commons/CommonCFunctions.java
```diff
@@ -167,10 +167,32 @@ public static void freeStringMap(IsolateThread thread, StringMap map, ExceptionH
         doCatch(exceptionHandlerPtr, new Runnable() {
             @Override
             public void run() {
+                Util.freeCharPtrPtr(map.getKeys(), map.getLength());
+                Util.freeCharPtrPtr(map.getValues(), map.getLength());
+                UnmanagedMemory.free(map);
```
HugoKulesza, line 172: Since the leak fix is not directly connected to this PR it would be good to split the correction in another small PR (it will also make the change of this PR clearer).

cpp/pypowsybl-java/powsybl-api.h
```diff
@@ -23,6 +23,12 @@ typedef struct string_map_struct {
     char** values;
 } string_map;
 
+typedef struct string_map_array_struct {
+    int length;
+    char** keys;
+    array* values;
+} string_map_array;
```
HugoKulesza, line 30: I'm not sure about this but isn't string_array_map a better name for this object (since it is a map containing string array values and not an array of string maps) ?

pypowsybl/network/impl/network.py
```diff
@@ -4919,6 +4919,87 @@ def get_bus_breaker_topology(self, voltage_level_id: str) -> BusBreakerTopology:
...
+
+            .. code-block:: python
+
+                {
+                    'S1VL1_BBS': ['LD1', 'TWT'],
+                    'LINE_S3S4': ['LD6', 'LINE_S3S4', 'SVC'],
+                }
+        """
+        if isinstance(element_ids, str):
+            element_ids = [element_ids]
```
HugoKulesza, line 5000: If we keep two separate python methods, I would only allow element_ids to be of List type to avoid already covering the use case of calling the `get_outage_group` method with only a str (it will also remove the sonar error on coverage by removing this if block 🙃)

java/pypowsybl/src/test/java/com/powsybl/python/network/NetworkUtilTest.java
```diff
@@ -52,6 +67,52 @@ void testBusFromBusBreakerViewBus() {
...
+                LinkedHashMap::new));
+
+        List<String> elementIds = contingencies.stream()
+            .map(Contingency::getId)
+            .toList();
+        Map<String, List<String>> outageGroups = NetworkUtil.getOutageGroups(network, elementIds);
+
+        assertThat(outageGroups.values()).allSatisfy(outageGroup -> assertThat(outageGroup).isSorted());
+        assertThat(outageGroups).containsOnlyKeys(elementIds);
+        elementIds.forEach(elementId -> assertThat(outageGroups.get(elementId)).isEqualTo(expectedOutageGroups.getOrDefault(elementId, List.of())));
```
HugoKulesza, line 105:
```suggestion
        elementIds.forEach(elementId -> assertThat(outageGroups).containsEntry(elementId, expectedOutageGroups.getOrDefault(elementId, List.of())));
```
(from Sonar issue suggestion)

java/pypowsybl/src/main/java/com/powsybl/python/network/NetworkUtil.java
```diff
@@ -199,6 +215,143 @@ static List<SwitchFlowContext> getSwitchFlowResults(Network network, List<String
...
+     *
+     * @param network network being evaluated
+     * @param loadFlowParameters base load-flow parameters used to derive OpenLoadFlow settings
+     * @param breakers whether breaker topology must be modeled in the load-flow network
+     * @return OpenLoadFlow network parameters configured for outage-group computation
+     */
+    private static LfNetworkParameters createOutageGroupLfNetworkParameters(Network network, LoadFlowParameters loadFlowParameters,
+                                                                            boolean breakers) {
+        return OpenLoadFlowParameters.createAcParameters(network, loadFlowParameters,
+            OpenLoadFlowParameters.get(loadFlowParameters), new SparseMatrixFactory(), new EvenShiloachGraphDecrementalConnectivityFactory<LfBus, LfBranch>(), breakers, false)
```
HugoKulesza, line 326:
```suggestion
            OpenLoadFlowParameters.get(loadFlowParameters), new SparseMatrixFactory(), new EvenShiloachGraphDecrementalConnectivityFactory<>(), breakers, false)
```
also from sonar suggestion (if I understood correctly, because the specific types are already set in the `createAcParameters` definition)

HugoKulesza (changes requested):
I have some design questions/suggestions on the singular/plural method "duplication".

Otherwise, small Sonar warnings to fix and naming to discuss, and the PR should be good to go !

nicow-elia:
Thanks for the feedback, all points seem sensible. Now I have to ask for how urgent it is because I realistically won't have time before end of september for following up, I am very busy with conferences at the moment. If you want it before that you can take over.

PR #1286, opened 2026-09-08
Droop curve API

SGI-Landry:

- [X] A PR or issue has been opened in all impacted repositories (if any)

The PR incidentally implements https://github.com/powsybl/pypowsybl/issues/1280. This is more broadly an implementation of the API for droop curves for AC-DC converters.

This PR introduces an API to create and get droop curves for `VoltageSourceConverter`s. It reflects the breaking change introduced in PowSyBl Core in https://github.com/powsybl/powsybl-core/pull/4067

Droop curves for voltage source converters are implemented / under implementation in PowSyBl Core and Open Load Flow. However, they cannot be accessed in PyPowSyBl.

Implementation of `create_droop_curve_segments` and `get_droop_curve_segments` in PyPowSyBl.

- [ ] Yes
- [X] No

**Other information**:

I made the following non-trivial choices
1. Support VoltageSourceConverter only and not generally AcDcConverter, because LCCs are not supported yet in PyPowSyBl. The extension should be trivial when needed.
2. No _update_ method for the droop curve, as in Core.
3. Support of per-unit get, even though this is not trivial for converters.
4. Segments are represented with min_v and max_v for each of them, with the obligation of consistency. Another possibility would be to have voltage points, but this would require $n+1$ points for $n$ droop curve segments, and data access would likely be more complicated.


SGI-Landry:
Hello @bperr can you please review this activity?

java/pypowsybl/src/main/java/com/powsybl/dataframe/network/PerUnitUtil.java
```diff
@@ -320,6 +320,16 @@ public static double perUnitV(NetworkDataframeContext context, double v, DcTermi
         return perUnitV(context, v, terminal.getDcNode().getNominalV());
     }
 
+    public static double perUnitK(NetworkDataframeContext context, double k, DcTerminal terminal) {
```
bperr, line 323: Maybe add some docsting to precise what "K" is. This quite obvious for V,R, X but K is not that common

bperr (approved):

java/pypowsybl/src/main/java/com/powsybl/dataframe/network/PerUnitUtil.java
SGI-Landry: Fixed, thanks.

SGI-Landry:
Hello @HugoKulesza 
This PR should be ready for review.

